PR #6, opened 2017-10-24
Fixed observer from registering selector on app delegate instead of K…

MikeDeasyBB:
…laviyo.


chrisconlon-klaviyo:
Apologies for not catching this until now. I just merged [a separate PR](https://github.com/klaviyo/klaviyo-swift-sdk/pull/7) that also fixes this, so I will close this one out now. Thank you for your contribution.

chrisconlon-klaviyo:
I've actually refactored again to use the selector defined on `Klaviyo` as you had here, great catch.

PR #11, opened 2019-01-17
Update to swift 4.2 #21244

ndurell:
* first pass at SDK cleanup
* add support for swift 4.2
* fix all warnings
* fix a few places where we were unsafely unwrapping optionals
* update project files to remove other warnings


.travis.yml
```diff
@@ -11,6 +11,6 @@ script:
 - xcodebuild -workspace Example/KlaviyoSwift.xcworkspace -scheme KlaviyoSwift-Example -sdk iphonesimulator
 - xcodebuild test -workspace Example/KlaviyoSwift.xcworkspace -scheme KlaviyoSwift-Example -destination "platform=iOS Simulator,OS=10.0,name=iPhone 7" | xcpretty -f `xcpretty-travis-formatter`
 - pod lib lint --quick
-osx_image: xcode8
+osx_image: xcode10.1
```
ndurell: since we are using swift 4.2 I believe we need a newer version of xcode.

Example/KlaviyoSwift.xcodeproj/project.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ndurell: Can't remember what these are for but I think it's ok to check them.

PR #14, opened 2019-02-02
Automate sdk updates with fastlane

ndurell:
* pulled script from here https://mar.codes/2018-11-14/Automate-open-source-libraries-releases-with-fastlane

tp link: https://klaviyo.tpondemand.com/entity/21271-automate-ios-sdk-deployment
NOTE: have not tested it yet.


fastlane/Appfile
```diff
@@ -0,0 +1,6 @@
+# app_identifier("[[APP_IDENTIFIER]]") # The bundle identifier of your app
```
chrisconlon-klaviyo: Do we need to fill any of this out for fastlane to work?

ndurell: Ahh good catch.
It looks like this is meant to be used across fastlane tasks. https://docs.fastlane.tools/advanced/Appfile/
I suspect for this project it probably makes sense to remove this line since this project is for a library rather an app.

ndurell: But also I'd like to test this out together once we have a release together then fix as necessary. My hunch though is none of this is needed.

PR #33, opened 2022-09-29
PUSH-127 - Fix archiveRootObject deprecation

ndurell:
archiveRootObject is deprecated now. This removes use of that method. Also this introduces some testing infrastructure so the changes can be unit tested.

# Changes:

* Fix deprecation for archiving queue and unarchiving
* Add new methods for archive/unarchive and file utilities
* Add environment file to enable dependency injection
* Add tests for archive and file utilties
* Add test environment

# Testing
New unit tests added

## Manual Test Plan
Happy Path:
1. Run test app and force queue to be created.
2. Restart app and ensure queue is emptied

Upgrade Path
1. Run existing test app with current sdk and ensure queue is created.
2. Upgrade the sdk and run the app again ensure the queue is emptied.

# Release Plan
Update spm version and changelog
Update cocoapods version to 1.5.2 and publish it.


kennyklaviyo (approved):

Sources/KlaviyoSwift/FileUtils.swift
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+filePathForData: returns a string representing the filepath where archived event queues are stored
+
+- Parameter apiKey: api key for distinguishing between sets of data
+- Parameter data: name representing the event queue to locate (will be either people or events)
+- Returns: filePath string representing the file location
+*/
+func filePathForData(apiKey: String, data: String) -> URL {
+    let fileName = "klaviyo-\(apiKey)-\(data).plist"
```
kennyklaviyo, line 35: hmm thoughts on using "api key" here?

ndurell: So I'm ok with it...what were you thinking?

PR #37, opened 2022-10-21
[TP 147318] - fix crash that occurs in the test app

ndurell:
Adds some guards in track and identify to prevent test app crash.

# Testing
I tested this manually with the test by sending a push after stopping the app.
When starting the app from a dead state it would crash on launch due a missing API key.

I also added some unit test to increase coverage.

### DIFF
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -229,7 +229,7 @@ public class Klaviyo : NSObject {
         assertPropertyTypes(properties: propertiesDict)
         
         guard let apiKey = apiKey else {
-            environment.logger.error("Track event called before API was set.")
+            environment.logger.error("Track event called before API key was set.")
             //TODO: store pending event for when api key is set.
             return
         }
@@ -294,8 +294,8 @@ public class Klaviyo : NSObject {
         }
         
         guard let apiKey = apiKey else {
-            environment.logger.error("Track person called before setting the API Key.")
-            //TODO: store pending event for when api key is set.
+            environment.logger.error("Track person called before API key was set.")
+            //TODO: store pending data for when api key is set.
             return
         }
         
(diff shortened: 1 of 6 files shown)


Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -228,15 +228,15 @@ public class Klaviyo : NSObject {
         let customerPropertiesDict = updatePropertiesDictionary(propDictionary: customerProperties)
         assertPropertyTypes(properties: propertiesDict)
         
+        guard let apiKey = apiKey else {
+            environment.logger.error("Track event called before API key was set.")
```
kennyklaviyo, line 232: should we replace existing usages of `print` with this logger too?

ndurell: Eventually yah...trying to keep this focused though. I can file a ticket for it.

kennyklaviyo (approved):

klaviyo-swift-sdk.xcodeproj/project.pbxproj
```diff
@@ -22,6 +22,7 @@
 
 /* Begin PBXBuildFile section */
 		7C375BC028E77149008605B3 /* KlaviyoTestUtils.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7C375BBF28E77149008605B3 /* KlaviyoTestUtils.swift */; };
+		7C54A46F2902F01E00611E4E /* LoggerClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7C54A46E2902F01E00611E4E /* LoggerClient.swift */; };
```
jordan-griffin, line 25: Question: Should we start gitignoring this file now that we are using SPM? Locals should be able to regenerate it with `swift package generate-xcodeproj` right?

jordan-griffin: I'm not entirely sure how safe that is though...

jordan-griffin (approved):

PR #42, opened 2022-11-14
TP150060 - klaviyo api client

ndurell:
The purpose of this PR is provide a new http client that can be used to send requests to Klaviyo via the v3 API. This also introduces snapshot testing to the SDK. This is being used to verify results of calls to the client produce expected http request as well whether or not payload as encoded properly (as json).

# Testing
The API client should have full test coverage.
TBD - test in the test app


Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -13,6 +13,115 @@ let version = "2.0.0"
...
+                    self.city = city
+                    self.country = country
+                    self.latitude = latitude
+                    self.longitude = longitude
+                    self.region = region
+                    self.zip = zip
+                    self.timezone = timezone
+                }
+            }
+            let email: String?
```
ndurell, line 87: Make public maybe...

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -13,6 +13,115 @@ let version = "2.0.0"
...
+                 properties: [String : Any]?=nil) {
+                self.email = email
+                self.phoneNumber = phoneNumber
+                self.externalId = externalId
+                self.firstName = firstName
+                self.lastName = lastName
+                self.organization = organization
+                self.title = title
+                self.image = image
+                self.location = location
```
ndurell, line 115: Provide default location with just timezone....

Sources/KlaviyoSwift/KlaviyoAPI.swift
```diff
@@ -0,0 +1,332 @@
...
+        case rateLimitError
+        case missingOrInvalidResponse(URLResponse?)
+        case networkError(Error)
+        case internalError(String)
+        case internalRequestError(Error)
+        case unknownError(Error)
+        case dataEncodingError(KlaviyoRequest)
+        case invalidData
+    }
+    var post: (KlaviyoRequest, @escaping (Result<Data, KlaviyoAPIError>) -> Void) -> Void = { request, result in
```
ndurell, line 181: sendRequest

Tests/KlaviyoSwiftTests/__Snapshots__/EncodableTests/testEventPayload.1.json
```diff
@@ -0,0 +1,25 @@
+{
```
ndurell, line 1: Add token encoding test.

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
```diff
@@ -9,18 +9,33 @@ import Foundation
...
 struct KlaviyoEnvironment {
     var archiverClient: ArchiverClient
     var fileClient: FileClient
     var data: (URL) throws -> Data
     var logger: LoggerClient
     var networkSession: NetworkSession
+    var apiURL: String
+    var encodeJSON: (Encodable) throws -> Data
+    var uuid: () -> UUID
+    var date: () -> Date
```
kennyklaviyo, line 29: not to make this more complex but seems it would be worth refactoring into a new struct or something like `RequestHelper` and have one of those in the Environment struct, it seems strange to have a date lambda on the environment

ndurell: Well there might be a lot of cases where you want the current date beyond building requests...

ndurell: The idea is to control it in the test environment.

Sources/KlaviyoSwift/KlaviyoAPI.swift
```diff
@@ -0,0 +1,332 @@
...
+
+
+struct KlaviyoAPI {
+    struct KlaviyoRequest {
+      enum KlaviyoEndpoint {
+          struct CreateProfilePayload: Encodable {
+              /**
+               Internal structure which details not needed by the public API.
+               */
+              struct Profile: Encodable {
```
kennyklaviyo, line 18: can this be achieved without this nesting structure? does `internal` exist as a keyword? i feel like this file really long... if we add a new endpoint that would have its own struct Payload and continues to add to this file

ndurell: I'm happy to move it to it's own file! :)

ndurell: Everything that isn't public should be internal.

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -13,6 +13,115 @@ let version = "2.0.0"
 
 public class Klaviyo: NSObject  {
     
+    public struct Event {
+        public struct Attributes {
+            public struct Metric {
```
kennyklaviyo, line 18: can we split this into its own file?

ndurell: Uhh this is gonna be public API - I was thinking it might be easier to read api docs with this nesting.

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -13,6 +13,115 @@ let version = "2.0.0"
...
+                        profile: [String : Any],
+                        value: Double? = nil,
+                        time: Date? = nil,
+                        uniqueId: String? = nil) {
+                self.profile = profile
+                self.metric = metric
+                self.properties = properties
+                self.value = value
+                self.time = time
+                self.time = time ?? environment.date()
```
kennyklaviyo, line 44: is line 43 redundant?

ndurell: Yup good catch!

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -13,6 +13,115 @@ let version = "2.0.0"
 
 public class Klaviyo: NSObject  {
     
+    public struct Event {
+        public struct Attributes {
+            public struct Metric {
+                public let name: String
+                public let service: String?
+                public init(name: String,
+                            service: String?="ios-analytics") {
```
kennyklaviyo, line 22: should we default on line 24 instead? does doing `Metric(name: "some name", service: nil)` bypass the default here?

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -13,6 +13,115 @@ let version = "2.0.0"
...
+            let phoneNumber: String?
+            let externalId: String?
+            let firstName: String?
+            let lastName: String?
+            let organization: String?
+            let title: String?
+            let image: String?
+            let location: Location?
+            let properties: [String: Any]?
+            public init(email: String?=nil,
```
kennyklaviyo, line 97: is there no data class object that avoids all this boilerplate of init?

ndurell: You get init for free but it's not public.

ndurell: And then if you want optional stuff you need to specify it.

Sources/KlaviyoSwift/Vendor/AnyEncodable.swift
```diff
@@ -0,0 +1,303 @@
+/**
```
kennyklaviyo, line 1: will we have to keep this updated in any way?

ndurell: Probably not - it's mostly base data types. I can make a regular dependency but it had other stuff I didn't need.

ndurell: We need to encode [String: Any] stuff.

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
kennyklaviyo: i just find that its a little strange to have this if its mainly for the test environment, doesnt feel like a well integrated solution

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: Yah I guess we don't need them to change this.

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: Punting on this made this whole in a separate file - should solve both problems.

Sources/KlaviyoSwift/KlaviyoAPI.swift
ndurell: Moved to a new file internal models. Still nested under the old one.

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
ndurell: I added a sub environment.

kennyklaviyo (approved):

PR #44, opened 2022-11-29
Update the handlePush method for $opened_push event

evan-masseau:
Pertains to [TP148025](https://klaviyo.tpondemand.com/entity/148025-update-opened-push-event) so that push token can be added to the Opened Push event's properties.

Updated Opened Push to (intentionally) send whole push payload back as properties. Serendipitously we already were doing that because the previous code was checking for "_k" in the wrong place anyways.

Added push token to the Opened Push track event payload directly (is it going to be removed from customer properties?) 

And added a suggestion for how to manage access to the push token as a private property with getter/setter function that handle persistence to/from UserDefaults


ndurell (commented):
This would work however we will most likely end up storing this a different way.

evan-masseau:
OK, at the end of the day adding push token to that payload is all I really needed. You did say the push token is going to be removed from customer properties, right?

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -167,16 +168,18 @@ public class Klaviyo : NSObject {
...
      
      - Parameter userInfo: NSDictionary containing the push notification text & metadata
      */
     public func handlePush(userInfo: NSDictionary) {
-        if let metadata = userInfo["_k"] as? NSDictionary {
-            trackEvent(eventName: KLPersonOpenedPush, properties: metadata)
-        } else {
-            trackEvent(eventName: KLPersonOpenedPush, properties: userInfo as NSDictionary)
+        // Test it originated from klaviyo by checking presence of _k tracking parameters
+        if let body = userInfo["body"] as? NSDictionary, let _ = body["_k"] as? NSDictionary {
```
ajaysubra, line 178: ❔ Curious why do we need to type cast  `body["_k"]` to a `NSDictionary` in order to see if body has the key `_k`?

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -167,16 +168,18 @@ public class Klaviyo : NSObject {
...
+            //But we still want to send the full payload, plus the push token
+            let properties = userInfo.mutableCopy() as! NSMutableDictionary
```
ajaysubra, line 180: ❔ Are there any risk associated with force type casting `userInfo.mutableCopy()` to a `NSMutableDictionary` like a crash? Can we conditionally  type cast it?

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -331,17 +334,31 @@ public class Klaviyo : NSObject {
...
+
+    /**
+     getPushDeviceToken: Get push device token if registered
+     Provides managed access to the in-memory apnDeviceToken property,
+     to fetch from UserDefaults if it hasn't been yet
+
+     - Returns: String?
+     */
+    public func getPushDeviceToken() -> String? {
+        if apnDeviceToken == nil {
```
ajaysubra, line 354: 💭  Does it make sense to keep `UserDefaults` the source of truth for getting the device token instead of the in memory variable just so that we don't have to share the variable between functions here?

ajaysubra: If we did this this function could be a simple read only computed property. Just a suggestion.

ajaysubra (commented):
Added some minor questions but nothing that should block you.

ndurell:
> OK, at the end of the day adding push token to that payload is all I really needed. You did say the push token is going to be removed from customer properties, right?

Yah I don't think it makes sense to include in every request how I will ensure it gets included in the open event as a part of the payload.

Sources/KlaviyoSwift/Klaviyo.swift
evan-masseau: I suppose we don't, other than to make sure _k is a dict as we'd expect. When we circle back to this after noah's bigger changes, I can revisit

Sources/KlaviyoSwift/Klaviyo.swift
evan-masseau: I don't think it should ever fail, but fair enough, it could be another condition in the if let above.

Sources/KlaviyoSwift/Klaviyo.swift
evan-masseau: Just a little quicker to have it in memory too is all

evan-masseau:
Updated to remove conflict after Noah's big PRs were merged... but most of what I had changed is just totally moved around now so the only purpose of the PR at this point is just to make sure push_token gets added to the $opened_push track event

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -139,14 +139,20 @@ public class Klaviyo: NSObject  {
...
      
      - Parameter userInfo: NSDictionary containing the push notification text & metadata
      */
     public func handlePush(userInfo: NSDictionary) {
-        if let _ = userInfo["_k"] as? NSDictionary {
-            trackEvent(eventName: KLPersonOpenedPush, properties: userInfo)
+        // Test it originated from klaviyo by checking presence of _k tracking parameters
+        if let body = userInfo["body"] as? NSDictionary, let _ = body["_k"] as? NSDictionary {
+            //But we still want to send the full payload, plus the push token
+            _ = environment.analytics.store.state.sink{ completion in
```
evan-masseau, line 151: I need to read up on this composed architecture pattern some more, maybe this isn't the proper way to access current state?

evan-masseau:
Unneeded.

PR #45, opened 2022-12-01
[TP150065] -  Network Session Now Using async/await

ndurell:
# Description
This is a bit more than a refactor but also this code isn't used by the SDK.
The idea here is to get the network session updated so it can be used with the future SDK store (forthcoming PR).
This also introduces the url protocol usage in the network session so that we can override url results in tests.
Most other changes are related to moving code around to more appropriate spots.

# Changes

- make network session use async/await
- add test over network session
- import url protocol to facilitate testing
- change environment to prevent initialization cycle
- app context info class to get information user agent header
- refactor klaviyo api test to support async await testing.
- move klaviyo api related out of model file.
- remove a not used encoding method

# Testing
This code is mostly just unit tested but I have tested it in the test on the final [PR](https://github.com/klaviyo/klaviyo-swift-sdk/pull/43).
Overall this should be a pretty safe PR and should solve an issue Ajay discovered which made some of our code backwards incompatible with and older version of swift.

# Links
More info on URLProtocol [testing](https://developer.apple.com/videos/play/wwdc2018/417/)


Tests/KlaviyoSwiftTests/NetworkSessionTests/testCreateEmphemeralSesionHeaders.1.txt
```diff
@@ -0,0 +1,20 @@
+▿ Optional<Dictionary<AnyHashable, Any>>
```
jordan-griffin, line 1: Should this file and the one below be in a `__snapshots__` folder? They look like more snapshots anyway...

ndurell: No this got added by accident I think good catch!

jordan-griffin (approved):
Kinda hard to follow the logic so might be good to get a second look over if you are unsure about anything in here but otherwise I thought the changes looked solid. Didnt see any other issues!

PR #46, opened 2022-12-01
[TP160065] - Klaviyo State and Encoding fixes.

ndurell:
# Description
This introduces KlaviyoState which is where we will store sdk state for the time being.
Requests will eventually be moved out of memory into separate files once we have batching.

# Testing
Testing was in place for some encoding.
I've added a ton of test coverage for the migration path.
I will also be increasing this as we get closer to the final PR.

# Changes
* unvendorize anyencodable now that we need decoding
* Add codable klaviyo state - future home of where sdk state is stored
* fix a snapshot - country was missing :(
* update cocoapods pod spec


Sources/KlaviyoSwift/InternalAPIModels.swift
```diff
@@ -190,7 +209,24 @@ extension Klaviyo.Profile.Attributes.Location: Encodable {
...
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(attributes, forKey: .attributes)
+    }
+}
+
+extension Klaviyo.Event.Attributes.Metric: Codable {
+    init(from decoder: Decoder) throws {
+        let values = try decoder.container(keyedBy: CodingKeys.self)
+//        self.service = try values.decode(String.self, forKey: .service)
```
jordan-griffin, line 226: accident?

ndurell: No actually it's a constant for now...I should probably just remove that or change how it's set.

jordan-griffin (approved):

Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -0,0 +1,134 @@
...
+        try environment.fileClient.removeItem(file.path)
+    } catch {
+        environment.logger.error("Unable to remove state file.")
+    }
+}
+
+func loadKlaviyoStateFromDisk(apiKey: String) -> KlaviyoState {
+    let fileName = klaviyoStateFile(apiKey: apiKey)
+    guard environment.fileClient.fileExists(fileName.path) else {
+        return migrateLegacyDataToKlaviyoState(with: apiKey, to: fileName)
```
evan-masseau, line 50: Doesn't this mean migration will be run on even a clean install that never had the old SDK? Perhaps that is just fine.

ndurell: Yah I probably could do a little more to detect if it's needed but it should be very quick and only run once the first time.

evan-masseau: As long as its safe to run it on a clean install i wouldn't worry, just hard to tell from only reading the code.

Package.resolved
```diff
@@ -1,5 +1,14 @@
 {
   "pins" : [
+    {
```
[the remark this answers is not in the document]
ndurell: I'm going to have vendorize something else soon though because the whole is pretty massive and I'm not sure they support cocoapods anymore. Fortunatley this one is tiny.

Sources/KlaviyoSwift/KlaviyoState.swift
ndurell: Ok...fair point...I'll see if I can make it a bit more readable in a future PR.

PR #48, opened 2022-12-15
[TP-150071] - numerous updates for migration and app life cyle

ndurell:
* app life cycle events moved into store.
* finished migration code for maintaining idfv
* many test updates
* a few new state management tests
* cleaned up TODOs


Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -210,28 +260,38 @@ struct KlaviyoReducer: ReducerProtocol {
...
-                await send(.enqueueRequest(request))
+            
+            guard let identifiers = legacyEvent.identifiers else {
+                return .none
+            }
+            state.email = identifiers.email ?? state.email
+            state.anonymousId = identifiers.anonymousId ?? state.anonymousId
+            state.phoneNumber = identifiers.phoneNumber ?? state.phoneNumber
+            state.externalId = identifiers.externalId ?? state.externalId
+            guard let request = try? legacyEvent.buildEventRequest(with: apiKey, from: state) else {
```
ndurell, line 274: @evan-masseau state is passed in here to be used in building the request.

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -363,4 +366,92 @@ class StateManagementTests: XCTestCase {
...
+            guard let request = try legacyEvent.buildEventRequest(with: apiKey, from: $0) else {
+                XCTFail()
+                return
+            }
+            $0.queue = [request]
+        }
+        
+    }
+    
+    func testLegacyEventOpenedPush() async throws {
```
ndurell, line 427: @evan-masseau this should test the push token logic.

Sources/KlaviyoSwift/InternalAPIModels.swift
```diff
@@ -239,24 +265,23 @@ struct LegacyEvent: Equatable {
...
-        }
-        if let externalId: String = customerProperties.removeValue(forKey: "$id") as? String ?? state.externalId {
-            customerProperties["externalId"] = externalId
+        // v3 events api still uses these properties - we are just ensuring we are using the latest
+        // identifiers here.
+        customerProperties["$email"] = state.email
+        customerProperties["$phone_number"] = state.phoneNumber
+        customerProperties["$id"] =  state.externalId
+        customerProperties["$anonymous"] =  state.anonymousId
+        if eventName == "$opened_push" {
```
ndurell, line 281: @evan-masseau push token logic here.

Sources/KlaviyoSwift/AppLifeCycleEvents.swift
```diff
@@ -0,0 +1,58 @@
+//
```
ndurell, line 1: Writing a test for this.

Sources/KlaviyoSwift/InternalAPIModels.swift
```diff
@@ -227,10 +227,36 @@ extension Klaviyo.Profile.Attributes.Location: Codable {
...
 
+struct LegacyIdentifiers {
+    let email: String?
+    let phoneNumber: String?
+    let externalId: String?
+    let anonymousId: String?
+    
+    static func extractFrom(from customerProperties: NSDictionary) -> LegacyIdentifiers? {
+        guard let customerProperties = customerProperties as? [String: Any] else {
+            return nil
```
ndurell, line 238: Need a test on this. Unlikely to occur though...

Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -210,28 +225,58 @@ struct KlaviyoReducer: ReducerProtocol {
...
-            return .run { [state] send in
-                guard let request = try? legacyProfile.buildProfileRequest(with: apiKey, from: state) else {
-                    return
+            guard let identifiers = legacyProfile.identifiers else {
+                return .none
+            }
+            state.email = identifiers.email ?? state.email
+            state.anonymousId = identifiers.anonymousId ?? state.anonymousId
+            state.phoneNumber = identifiers.phoneNumber ?? state.phoneNumber
+            state.externalId = identifiers.externalId ?? state.externalId
```
jordan-griffin, line 254: nit: but should we functionize this state setup somewhere? That's probably all we need but just thinking about if the amount of state we are sending to each request grows, we'll be maintaining that in two places currently

ndurell: I'm hoping this'll go away when we have a new api.

ndurell: But yah...I'm actually looking at testing this better now. Will do.

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
```diff
@@ -23,20 +25,45 @@ let decoder = {
     return decoder
 }()
 
+private let reachabilityService = Reachability(hostname: "a.klaviyo.com")
```
evan-masseau, line 28: Should we have only 1 reference for this base url?

evan-masseau: I mean, only one place where the base url is defined, and reference that here

ndurell: Oh technically this should be refreshed when the url changes...but that would only affect sdks running against locals. Are you more referring to having a constant for this?

Sources/KlaviyoSwift/StateManagement.swift
ndurell: Moved a mutating method on the state.

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
evan-masseau: Yeah i think you had introduced a way to have this as a constant, but overridable in DEBUG mode, so I was thinking about that.

ndurell: Yah I was actually gonna deprecate that in favor of the URLProtocol method but didn't get to that yet.

Sources/KlaviyoSwift/APIRequestErrorHandling.swift
```diff
@@ -0,0 +1,69 @@
...
+let MAX_RETRIES = 50
+let MAX_BACKOFF = 60 * 3 // 3 minutes
+
+private func getDelaySeconds(for count: Int) -> Int {
+    let delay = Int(pow(2.0, Double(count)))
+    let jitter = environment.randomInt()
+    return min(delay + jitter, MAX_BACKOFF)
+}
+
+func handleRequestErorr(request: KlaviyoAPI.KlaviyoRequest, error: KlaviyoAPI.KlaviyoAPIError, retryInfo: RetryInfo) -> KlaviyoAction {
```
ajaysubra, line 19: Should the function name be `handleRequestError`?

Sources/KlaviyoSwift/AppLifeCycleEvents.swift
```diff
@@ -0,0 +1,67 @@
...
+        let backgrounded = environment
+            .notificationCenterPublisher(UIApplication.didEnterBackgroundNotification)
+            .handleEvents(receiveOutput: { _ in
+                environment.stopReachability()
+            })
+            .map { _ in KlaviyoAction.stop }
+        // The below is a bit convoluted since network status can be nil.
+        let reachability = environment
+            .notificationCenterPublisher(ReachabilityChangedNotification)
+            .map { notification in
```
ajaysubra, line 43: [compactMap](https://developer.apple.com/documentation/swift/sequence/compactmap(_:)) instead of `.map.filter`?

Sources/KlaviyoSwift/InternalAPIModels.swift
```diff
@@ -227,10 +227,34 @@ extension Klaviyo.Profile.Attributes.Location: Codable {
 
 // MARK: Legacy request data
 
+struct LegacyIdentifiers {
+    let email: String?
+    let phoneNumber: String?
+    let externalId: String?
```
ajaysubra, line 233: Should at least one of the property in the struct be non-nil/required?

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
```diff
@@ -23,20 +25,45 @@ let decoder = {
...
     var archiverClient: ArchiverClient
     var fileClient: FileClient
     var data: (URL) throws -> Data
     var logger: LoggerClient
     var analytics: AnalyticsEnvironment
     var getUserDefaultString: (String) -> String?
-    static let production = KlaviyoEnvironment(
+    var appLifeCycle: AppLifeCycleEvents
+    var notificationCenterPublisher: (NSNotification.Name) -> AnyPublisher<Notification, Never>
+    var legacyIDFV: () -> String
```
ajaysubra, line 39: idfv?

Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -16,103 +16,114 @@ import Foundation
 // Request flush interval in seconds.
 let CELLULAR_FLUSH_INTERVAL = 30.0
 let WIFI_FLUSH_INTERVAL = 10.0
+let MAX_QUEUE_SIZE = 200
+
+enum RetryInfo: Equatable {
+    case retry(Int) // Int is current count for first request
+    case retryWithBackoff(Int, Int, Int) // requestCount, totalRetryCount, currentBackoff
```
ajaysubra, line 23: would it make sense to label the enums associated values? 
Ex: `case retryWithBackoff(requestCount : Int, totalRetryCount: Int, currentBackoff: Int)`

ajaysubra (commented):
Added some minor comments that are mostly questions. None of these should stop you from merging. Thanks!

Sources/KlaviyoSwift/AppLifeCycleEvents.swift
ndurell: Thanks for reminding me of this!

Sources/KlaviyoSwift/InternalAPIModels.swift
ndurell: Nope! We support anonymous identification via a uuid. The old sdk used idfv - identifierForVendor. This PR is about moving away from that one.

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
ndurell: Yes. We moved from this to a stored uuid. Maybe I'll change this to legacyIdentifier since it's not quite an IDFV. It has a "iOS:" prefixed to it.

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
evan-masseau: Cool, this is definitely a nit, its still easy enough to change the url while developing locally

ndurell: I think actually I can turn this into a function that creates the service.

evan-masseau (approved):

PR #52, opened 2022-12-23
[TP148558] - Add github actions

ndurell:
Builds and run tests for our SDK on xcode 13 and 14. 

This PR also adds back support for xcode 13 which was accidentally removed in a previous release.


Sources/KlaviyoSwift/KlaviyoAPI.swift
```diff
@@ -96,11 +97,11 @@ extension KlaviyoAPI.KlaviyoRequest {
     func encodeBody() throws -> Data {
         switch self.endpoint {
         case .createProfile(let payload):
-            return try environment.analytics.encodeJSON(payload)
```
ndurell, line 99: Theme here is we need explicit types for older versions of swift. AnyEncodable should encode the same way as before. The other option here would be to create specific encode method for each type which might have been tedious.

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
```diff
@@ -13,13 +13,13 @@ import UIKit
 var environment = KlaviyoEnvironment.production
 
 let PRODUCTION_HOST = "https://a.klaviyo.com"
-let encoder = {
+let encoder = { () -> JSONEncoder in
```
ndurell, line 16: You'll see this a lot in the PR explicit return types.

Sources/KlaviyoSwift/NetworkSession.swift
```diff
@@ -30,18 +30,36 @@ func createEmphemeralSession(protocolClasses: [AnyClass] = URLProtocolOverrides.
...
+        var data: (URLRequest) async throws -> (Data, URLResponse)
+        
+        static let production = { () -> NetworkSession in
+            let session = createEmphemeralSession()
+            
+            return NetworkSession(data: { request async throws -> (Data, URLResponse) in
+       
+                session.configuration.protocolClasses = URLProtocolOverrides.protocolClasses
+                if #available(iOS 15, *) {
+                    return try await session.data(for: request)
```
ndurell, line 44: Sadly this API is not available until iOS 15 so I fallback to my own version of it.

Sources/KlaviyoSwift/Vendor/ComposableArchitecture/Store.swift
```diff
@@ -398,161 +409,3 @@ final class Store<State, Action> {
 /// let store: StoreOf<Feature>
 /// ```
 typealias StoreOf<R: ReducerProtocol> = Store<R.State, R.Action>
-
-#if swift(>=5.7)
```
ndurell, line 402: I removed scoping out of this since we are using it right now and it was difficult to get to compile.

Tests/KlaviyoSwiftTests/StateChangePublisherTests.swift
```diff
@@ -80,20 +92,24 @@ final class StateChangePublisherTests: XCTestCase {
...
         }
         
         let reducer = KlaviyoTestReducer(reducer: initializationReducer)
         let test = Store(initialState: .test, reducer: reducer)
         environment.analytics.store = test
-        _ = environment.analytics.store.send(.initialize("foo"))
-        _ = environment.analytics.store.send(.flushQueue)
-        _ = environment.analytics.store.send(.flushQueue)
         
+        @MainActor func runDebouncedEffect() {
```
ndurell, line 105: For older versions of swift this is needed to get tests to pass. Noticed this in composable arch code.

Tests/KlaviyoSwiftTests/StateChangePublisherTests.swift
```diff
@@ -80,20 +92,24 @@ final class StateChangePublisherTests: XCTestCase {
                 state.initalizationState = .initialized
                 return StateChangePublisher.test.publisher().eraseToEffect()
             case .flushQueue:
-                state = state
+                return .none
```
ndurell, line 95: This code was silly anyway and didn't work in older versions of swift.

ajaysubra (approved):

.github/workflows/swift.yml
```diff
@@ -0,0 +1,28 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
```
jordan-griffin, line 6: Should we have a workflow that runs tests on PRs too?

.github/workflows/swift.yml
```diff
@@ -0,0 +1,28 @@
...
+    branches:
+      - '*'
+  workflow_dispatch:
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  library:
```
jordan-griffin, line 17: nit: Might be nice to give this job a name as well, similar to the `name` fields you added to the individual steps below.

.github/workflows/swift.yml
ndurell: It ran on my PR?

ndurell: But I wonder if main is correct - I think this one uses master right?

jordan-griffin: Ya I was surprised the repo used main, master would be in the case for most older repos like this, right? in the Android SDK we have two workflows, one runs on master, the other runs on feature branches

ndurell: I'll fix to use master. I think I'd rather just run on PRs vs branches right?

jordan-griffin: Ya just on PRs makes sense to me! Thats what I meant sorry

PR #56, opened 2023-01-18
My suggestion for how to have a unified host URL without much code change

evan-masseau:


evan-masseau:
now that both places just need a base url without `api` in the path, this approach works

ndurell (approved):
Thanks!

PR #57, opened 2023-01-20
[TP149126] - Handle pending profile requests

ndurell:
This will allow developers to log pending profile request. 

* Introduces a new internal enum for storing pending "requests".
* Pending requests are applied after initialization.


ajaysubra (approved):

evan-masseau (approved):
Makes sense to me

PR #59, opened 2023-01-24
New API for the iOS SDK

ndurell:
Implementation of the new API for iOS.

1. Creates KlaviyoSDK struct
2. Hooked up old methods to new struct
3. Marking methods on Klaviyo Object as `@objc` for objc support.

TODO: set profile property.

### DIFF
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -143,6 +143,9 @@ public class Klaviyo: NSObject  {
                 .create(event: Event(attributes: .init(name: .OpenedPush,
                                                        properties: properties,
                                                        profile: [:])))
+            if let url = body["url"] as? String, let url = URL(string: url) {
+                UIApplication.shared.open(url)
+            }
         }
         
     }
@@ -443,9 +446,13 @@ public struct KlaviyoSDK {
             create(event: Event(attributes: .init(name: .OpenedPush, properties: properties, profile: [:])))
             Task {
                 await MainActor.run {
+                    if let url = body["url"] as? String, let url = URL(string: url) {
+                        UIApplication.shared.open(url)
+                    }
                     completionHandler()
                 }
             }
+      
             return true
         }
         return false
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -312,23 +312,23 @@ public struct KlaviyoSDK {
         // This seems tricky to implement with Any - might need to restrict to something equatable, encodable....
     }
     @_spi(KlaviyoPrivate)
-    func create(event: Event) {
+    public func create(event: Event) {
         dispatchOnMainThread(action: .enqueueEvent(event))
     }
     @_spi(KlaviyoPrivate)
-    func set(pushToken: Data) {
+    public func set(pushToken: Data) {
         let apnDeviceToken = pushToken.map { String(format: "%02.2hhx", $0) }.joined()
         dispatchOnMainThread(action: .setPushToken(apnDeviceToken))
     }
     @_spi(KlaviyoPrivate)
-    func handle(remoteNotification userInfo: [AnyHashable : Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+    public func handle(remoteNotification userInfo: [AnyHashable : Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
         if let properties = userInfo as? [String: Any], let body = properties["body"] as? [String: Any], let _ = body["_k"] {
             create(event: Event(attributes: .init(metric: .init(name: .OpenedPush), properties: properties, profile: [:])))
             completionHandler(.noData)
         }
     }
     @_spi(KlaviyoPrivate)
-    func handle(notificationResponse: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
+    public func handle(notificationResponse: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
         if let properties = notificationResponse.notification.request.content.userInfo as? [String: Any],
            let body = properties["body"] as? [String: Any], let _ = body["_k"] {
             create(event: Event(attributes: .init(metric: .init(name: .OpenedPush), properties: properties, profile: [:])))
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -137,12 +137,14 @@ public class Klaviyo: NSObject  {
      */
     @objc
     public func handlePush(userInfo: NSDictionary) {
-        guard let userInfo = userInfo as? [AnyHashable: Any] else {
-            return
+        if let properties = userInfo as? [String: Any],
+           let body = properties["body"] as? [String: Any], let _ = body["_k"] {
+            KlaviyoSDK()
+                .create(event: Event(attributes: .init(name: .OpenedPush,
+                                                       properties: properties,
+                                                       profile: [:])))
         }
-        _ = Klaviyo.sdkInstance.handle(remoteNotification: userInfo) { _ in
-            // Empty implementation
-        }
+        
     }
     
     /**
@@ -338,7 +340,7 @@ public struct KlaviyoSDK {
     /// - Parameter profile: a profile object to send to Klaviyo
     @_spi(KlaviyoPrivate)
     public func set(profile: Profile) {
-        dispatchOnMainThread(action: .enqueueProile(profile))
+        dispatchOnMainThread(action: .enqueueProfile(profile))
     }
     
     /// Reset all profile data that was logged to Klaviyo. This includes the anonymous identifiers as well as the push token.
@@ -413,22 +415,6 @@ public struct KlaviyoSDK {
         return self
     }
     
-    /// Track a remote push notification open event in Klaviyo
-    /// - Parameters:
-    ///   - remoteNotification: the remote notificaiton that was opened
-    ///   - fetchCompletionHandler: a completion handler that will be called with a result for Klaviyo notifications
-    /// - Returns: true if the notificaiton originated from Klaviyo, false otherwise.
-    @_spi(KlaviyoPrivate)
-
-    public func handle(remoteNotification userInfo: [AnyHashable : Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) -> Bool {
-        if let properties = userInfo as? [String: Any], let body = properties["body"] as? [String: Any], let _ = body["_k"] {
-            create(event: Event(attributes: .init(name: .OpenedPush, properties: properties, profile: [:])))
-            completionHandler(.noData)
-            return true
-        }
-        return false
-    }
-    
     /// Track a notificationResponse open event in Klaviyo
     /// - Parameters:
     ///   - remoteNotification: the remote notificaiton that was opened
@@ -439,7 +425,11 @@ public struct KlaviyoSDK {
         if let properties = notificationResponse.notification.request.content.userInfo as? [String: Any],
            let body = properties["body"] as? [String: Any], let _ = body["_k"] {
             create(event: Event(attributes: .init(name: .OpenedPush, properties: properties, profile: [:])))
-            completionHandler()
+            Task {
+                await MainActor.run {
+                    completionHandler()
+                }
+            }
             return true
         }
         return false
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -280,51 +280,54 @@ public class Klaviyo: NSObject  {
 }
 
 
-struct KlaviyoSDK {
-    func initialize(with apiKey: String) {
+@_spi(KlaviyoPr)
+public struct KlaviyoSDK {
+    
+    @_spi(KlaviyoPrivate)
+    public func initialize(with apiKey: String) {
         dispatchOnMainThread(action: .initialize(apiKey))
     }
-    
-    func set(profile: Profile) {
+    @_spi(KlaviyoPrivate)
+    public func set(profile: Profile) {
         dispatchOnMainThread(action: .enqueueProile(profile))
     }
-    
-    func resetProfile() {
+    @_spi(KlaviyoPrivate)
+    public func resetProfile() {
         dispatchOnMainThread(action: .resetProfile)
     }
-    
-    func set(email: String) {
+    @_spi(KlaviyoPrivate)
+    public func set(email: String) {
         dispatchOnMainThread(action: .setEmail(email))
     }
-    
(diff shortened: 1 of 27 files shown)


Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -121,9 +138,10 @@ public class Klaviyo: NSObject  {
      
      - Parameter userInfo: NSDictionary containing the push notification text & metadata
      */
+    @objc
     public func handlePush(userInfo: NSDictionary) {
-        if let body = userInfo["body"] as? NSDictionary, let _ = body["_k"] {
-            trackEvent(eventName: KLPersonOpenedPush, properties: userInfo)
+        Klaviyo.sdkInstance.handle(remoteNotification: userInfo as! [AnyHashable : Any]) { result in
+            // Empty implementation
```
evan-masseau, line 144: Is this supposed to be empty? In the prior version didn't this method trigger `$opened_push`

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -259,13 +277,56 @@ public class Klaviyo: NSObject  {
                 , "Property values must be of NSString, NSNumber, NSNull, NSDictionary, NSDate, or NSURL. Got: \(String(describing: properties[k as! NSCopying]))")
         }
     }
+}
+
+
+struct KlaviyoSDK {
```
evan-masseau, line 283: I'm mixing up languages here -- does this need to be marked internal or private so they can't instantiate it themselves or is that what is is by default in swift.

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -259,13 +277,56 @@ public class Klaviyo: NSObject  {
...
+        dispatchOnMainThread(action: .enqueueEvent(event))
+    }
+    
+    func set(pushToken: Data) {
+        let apnDeviceToken = pushToken.map { String(format: "%02.2hhx", $0) }.joined()
+        dispatchOnMainThread(action: .setPushToken(apnDeviceToken))
+    }
+    
+    func handle(remoteNotification userInfo: [AnyHashable : Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+        if let properties = userInfo as? [String: Any], let body = properties["body"] as? [String: Any], let _ = body["_k"] {
```
evan-masseau, line 319: Gotta get better about writing these things down, but didn't we determine we don't want to trigger $opened_push here, only in the UNNofiticationResponse method, as long as they implement both?

evan-masseau: I also cut a TP ticket about putting de-duping $opened_push by message+push token in that event processor.

evan-masseau (commented):
Seems logical so far, are you going to keep working on this PR to add the `setProfileAttribute` method?
We should probably review our two READMEs so they have similar structure etc. The iOS one will also need to have a migration section for 2.0.

Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -272,6 +275,38 @@ struct KlaviyoReducer: ReducerProtocol {
...
+            }
+            let profile = KlaviyoAPI.KlaviyoRequest.KlaviyoEndpoint.CreateProfilePayload.Profile(profile: profile,
+                                                                                                 anonymousId: anonymousId)
+            let payload = KlaviyoAPI.KlaviyoRequest.KlaviyoEndpoint.CreateProfilePayload(data: profile)
+            let endpoint = KlaviyoAPI.KlaviyoRequest.KlaviyoEndpoint.createProfile(payload)
+            let request = KlaviyoAPI.KlaviyoRequest(apiKey: apiKey, endpoint: endpoint)
+            state.queue.append(request)
+            return .none
+        case .resetProfile:
+            state.email = nil
```
evan-masseau, line 305: I think we also ought to clear push token in `resetProfile` ?

evan-masseau: So that the developer has to call `setPushToken` again on the new profile, for consistent behavior.

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -121,9 +138,10 @@ public class Klaviyo: NSObject  {
      
      - Parameter userInfo: NSDictionary containing the push notification text & metadata
      */
+    @objc
     public func handlePush(userInfo: NSDictionary) {
-        if let body = userInfo["body"] as? NSDictionary, let _ = body["_k"] {
-            trackEvent(eventName: KLPersonOpenedPush, properties: userInfo)
+        Klaviyo.sdkInstance.handle(remoteNotification: userInfo as! [AnyHashable : Any]) { result in
```
ajaysubra, line 143: Is it possible to avoid forced type cast here?

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -92,27 +106,30 @@ public class Klaviyo: NSObject  {
      
      - Parameter apiKey: string representation of the Klaviyo API Key
      */
+    @objc
```
ajaysubra, line 109: Is it sufficient to just mark the class as open to Objective-C instead of marking each method?

Sources/KlaviyoSwift/Klaviyo.swift
ajaysubra: Internal is default in swift.

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -259,13 +277,56 @@ public class Klaviyo: NSObject  {
...
+    
+    func handle(remoteNotification userInfo: [AnyHashable : Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+        if let properties = userInfo as? [String: Any], let body = properties["body"] as? [String: Any], let _ = body["_k"] {
+            create(event: Event(attributes: .init(metric: .init(name: .OpenedPush), properties: properties, profile: [:])))
+            completionHandler(.noData)
         }
     }
     
+    func handle(notificationResponse: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void) {
+        if let properties = notificationResponse.notification.request.content.userInfo as? [String: Any],
```
ajaysubra, line 326: should we call the completion handler outside of the `if let` condition else if any one of the if let condition is not satisfied the closure will never get called?

evan-masseau: I think the idea is we only call the completion handler if it the notification originated from Klaviyo... that said we should probably return a bool so they know whether we called the completion handler

ndurell: Agreed on the bool!

Sources/KlaviyoSwift/Klaviyo.swift
jordan-griffin: I'm curious about this too, I would assume the decorator on the class would scope all contained methods too

Sources/KlaviyoSwift/Klaviyo.swift
jordan-griffin: +1 this caught my eye too

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -259,13 +273,67 @@ public class Klaviyo: NSObject  {
                 , "Property values must be of NSString, NSNumber, NSNull, NSDictionary, NSDate, or NSURL. Got: \(String(describing: properties[k as! NSCopying]))")
         }
     }
+}
+
+
+@_spi(KlaviyoPrivate)
```
jordan-griffin, line 279: Just to make sure I follow:

`@_spi` is used to mark this class and its methods as internal and not intended for public usage by developers right?

ndurell: Yah I put it here because I'm not ready to expose the api yet but we would rip this particular one off when we are ready.

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: You need it on both I believe. https://developer.apple.com/forums/thread/129282

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: Fixed it.

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: Makes sense!

Sources/KlaviyoSwift/Klaviyo.swift
jordan-griffin: Gotcha that makes total sense to me! Thanks!

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: Pushing a fix now.

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: I'm calling into my own API. Our old SDK api doesn't take in a completion handler. We will have to intro a new objc api to take a completion handler.

Sources/KlaviyoSwift/StateManagement.swift
ndurell: I believe I addressed this...

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: I think you might want to have both in case you don't implement both options. I don't know why you would do that but having the extra api gives them the option to not have to implement both.

evan-masseau: ok. as long as we're deduping i guess it doesnt matter.

ndurell: That said potentially we could update the docs though to give options vs implementing both but I thought it wouldn't call both.

ndurell: Ok will re-test!

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -259,13 +279,175 @@ public class Klaviyo: NSObject  {
...
     
-    private func dispatchOnMainThread(action: KlaviyoAction) {
-        Task {
-            await MainActor.run {
-                _ = environment.analytics.store.send(action)
-            }
+    /// Returns the email for the current user, if any.
+    @_spi(KlaviyoPrivate)
+    public var email: String? {
+        get {
```
ajaysubra, line 313: Nit: Can we drop the `get`?

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -259,13 +279,176 @@ public class Klaviyo: NSObject  {
...
+        let apnDeviceToken = pushToken.map { String(format: "%02.2hhx", $0) }.joined()
+        dispatchOnMainThread(action: .setPushToken(apnDeviceToken))
+        return self
+    }
+    
+    /// Track a notificationResponse open event in Klaviyo
+    /// - Parameters:
+    ///   - remoteNotification: the remote notificaiton that was opened
+    ///   - fetchCompletionHandler: a completion handler that will be called with a result for Klaviyo notifications
+    /// - Returns: true if the notificaiton originated from Klaviyo, false otherwise.
```
ajaysubra, line 438: should both of these be combined? Like does it makes sense to have both the completion handler and the function return indicate the success of the operation?

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: Resolved this!

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: Yah good point!

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: This completion handler is meant to be the one given to us by the delegate method. If you do not call it you get a warning. However if we aren't handling the callback we should call it. So unfortunately no we can't combine them.

ajaysubra: ah I thought so.

Sources/KlaviyoSwift/Klaviyo.swift
evan-masseau: So in swift a var with just get defined cannot be assigned at all, right?

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -259,13 +279,176 @@ public class Klaviyo: NSObject  {
...
+    @_spi(KlaviyoPrivate)
+    @discardableResult
+    public func initialize(with apiKey: String) -> KlaviyoSDK {
+        dispatchOnMainThread(action: .initialize(apiKey))
+        return self
+    }
+    
+    /// Set a profile in your Klaviyo account.
+    /// Future SDK calls will use this data when making api requests to Klaviyo.
+    /// NOTE: this will trigger a reset of existing
```
evan-masseau, line 353: This feels like a sentence fragment, "this will reset the existing currently tracked profile" maybe?

evan-masseau: also, can you link to another method, like @see resetProfile ?

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -259,13 +279,176 @@ public class Klaviyo: NSObject  {
...
+    /// - Returns: a KlaviyoSDK instance
+    @_spi(KlaviyoPrivate)
+    @discardableResult
+    public func set(email: String) -> KlaviyoSDK {
+        dispatchOnMainThread(action: .setEmail(email))
+        return self
+    }
+    
+    /// Set the current user's phone number.
+    /// Note that the phone number should be in a format that Klaviyo accepts.
```
evan-masseau, line 380: NOTE

evan-masseau (approved):
works for me, just left a few comments on the comments.

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: haha!

Sources/KlaviyoSwift/Klaviyo.swift
ndurell: Not sure...

ndurell: Done.

Sources/KlaviyoSwift/Klaviyo.swift
ajaysubra: So what Noah initially had was a regular computed property which had a getter and it comes with an optional setter. However, if you know you are never going to set a value to it you can use read only computed properties that only have getters and the way you do that is by dropping the 'get`.

PR #61, opened 2023-02-27
Removed .provisional from the example to setup push 

ajaysubra:
Removed .provisional from the example to setup push since it's more of a niche usecase and also aligning closer to [apple docs](https://developer.apple.com/documentation/usernotifications/asking_permission_to_use_notifications).

Also, having `.provisional` will not bring up the alert controller for developers to ask users permission of push which may not be the behaviour they intend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,7 @@ Add the below code to your application wherever you would like to prompt users t
 
     let center = UNUserNotificationCenter.current()
     center.delegate = self as? UNUserNotificationCenterDelegate
-    let options: UNAuthorizationOptions = [.alert, .sound, .badge, .provisional]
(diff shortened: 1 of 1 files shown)


ajaysubra:
> I'm more on the fence on this one. I think it might be beneficial to recommend this by default. Though our example apps don't have to use it.

Yup. I will add some comments about this when I get to adding deep linking stuff to the readme today or tomorrow.

ajaysubra:
Closing this PR as I added a note about provisional push notifications in #64 so that users are aware of provisional notifications and have the option to chose that if desired. However, at some point we should have more clear documentation on this on our support blogs.

PR #68, opened 2023-03-21
Add Linting and Formatting to the Swift SDK

ndurell:
Using pre-commit to run linting and formatting on every pull and on master.


Tests/KlaviyoSwiftTests/Vendor/ComposableArchitecture/TestStore.swift
```diff
@@ -20,7 +20,7 @@
  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
```
ajaysubra, line 20: should we exclude vendor code?

Sources/KlaviyoSwift/Vendor/ReachabilitySwift.swift
```diff
@@ -40,10 +40,10 @@ let ReachabilityChangedNotification = NSNotification.Name("ReachabilityChangedNo
 func callback(reachability:SCNetworkReachability, flags: SCNetworkReachabilityFlags, info: UnsafeMutableRawPointer?) {
```
ajaysubra, line 40: same question about vendor code here as well.

ajaysubra (approved):
👍🏽 with a question about vendor code.

Tests/KlaviyoSwiftTests/Vendor/ComposableArchitecture/TestStore.swift
ndurell: I did for most things. I can exclude here too but seemed harmless for now.

PR #69, opened 2023-03-21
Moving example iOS code into SDK

ajaysubra:
Moving code that was in [this](https://github.com/klaviyo/KlaviyoSwiftExamples/) repo back to the SDK so that it's more easily discoverable by developers.


jordan-griffin:
Have you run these examples to ensure they are operating as expected from their new location? (following the little README you added to install the setup I guess).

Also, is there any changes from the original code we should know about, or is this pretty much a straight rip from the original repo?

jordan-griffin:
Oh, and maybe some screenshots to show what the examples look like might be good context?

Just want to make sure these are "up to standard" if we're gonna be exposing them to customers to use as a baseline to integrate with Klaviyo's SDK

Examples/KlaviyoSwiftExamples/Shared/CheckOutViewController.swift
```diff
@@ -6,67 +6,70 @@
...
         cart.saveCart()
 
         // Trigger "Checkout Completed" Event
-        //TODO: check with Noah why SDK doesn't take in swift dictionary
+        // TODO: check with Noah why SDK doesn't take in swift dictionary
         let propertiesDictionary = NSMutableDictionary()
         propertiesDictionary[Klaviyo.sharedInstance.KLEventTrackPurchasePlatform] = "iOS \(UIDevice.current.systemVersion)"
         propertiesDictionary["Total Price"] = cartTotal
-        propertiesDictionary["Items Purchased"] = uniqueItemsArray.compactMap { $0.name }
+        propertiesDictionary["Items Purchased"] = uniqueItemsArray.compactMap(\.name)
```
ajaysubra, line 70: @ndurell This lint change is causing a compiler error, may be because the example app isn't using the same version of swift?
image

ajaysubra: Essentially these two changes are causing the compiler on the example app to fail - 
image

Are you seeing any issues on the SDK side?

ndurell: Nope...what version of swift are you using?

ajaysubra:
> Oh, and maybe some screenshots to show what the examples look like might be good context?
> 
> Just want to make sure these are "up to standard" if we're gonna be exposing them to customers to use as a baseline to integrate with Klaviyo's SDK

Added a demo. Personally, I think the views can be improved but code wise I did my best given the time to [modernize](https://github.com/klaviyo/KlaviyoSwiftExamples/pull/6) it.

ajaysubra:
> Have you run these examples to ensure they are operating as expected from their new location? (following the little README you added to install the setup I guess).
> 
> Also, is there any changes from the original code we should know about, or is this pretty much a straight rip from the original repo?

I tested the SPM path. Things that I'm testing today - 
1. Cocoapods install 
2. Push (by adding this app to our apple developer)

This is just the same code that was in the example repo. Only other change are running linting changes.

ajaysubra:
> > Have you run these examples to ensure they are operating as expected from their new location? (following the little README you added to install the setup I guess).
> > Also, is there any changes from the original code we should know about, or is this pretty much a straight rip from the original repo?
> 
> I tested the SPM path. Things that I'm testing today -
> 
> 1. Cocoapods install
> 2. Push (by adding this app to our apple developer)
> 
> This is just the same code that was in the example repo. Only other change are running linting changes.

Tested both and it looks great!

PR #70, opened 2023-03-21
Deprecate the old sdk and expose the new one.

ndurell:
* Also update the new one to include push token


Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -19,6 +19,8 @@ private func dispatchOnMainThread(action: KlaviyoAction) {
 
 // MARK: Objective-C
 
+@available(
+    iOS, deprecated: 9999, message: "Use either KlaviyoSDK for swift bindings or KlaviyoObjc for Objective-C bindings.")
```
jordan-griffin, line 23: if you use the `deprecated` decorator without specifying a version number does it not deprecate for all versions?

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
```diff
@@ -12,7 +12,7 @@ import UIKit
 
 var environment = KlaviyoEnvironment.production
 
-let PRODUCTION_HOST = Klaviyo.sharedInstance.KlaviyoServerURLString
+let PRODUCTION_HOST = "https://a.klaviyo.com"
```
jordan-griffin, line 15: Wait why is this changing? You didnt change that constant

Sources/KlaviyoSwift/Klaviyo.swift
ajaysubra: Also would it make sense to use `renamed` as that seems to let folks use the fix it option in xcode vs. having to manually find the method and replace it.

image

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
ndurell: This will eventually go away and it's not needed anymore. Changing host is now accomplished through a different mechanism.

jordan-griffin (approved):

Examples/Package.swift
ndurell, line 1: This is done to hide the examples directory.

PR #71, opened 2023-03-23
use better flush interval when network is unreachable

ndurell:
When the network become unreachable the network timer got set to zero. 
If the timer ever restarted under these conditions it trigger many flushes.
In addition some of the cancellation ids were incorrect which would have prevented duplicate timers from being created.

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement.swift
@@ -179,7 +179,7 @@ struct KlaviyoReducer: ReducerProtocol {
                     KlaviyoAction.flushQueue
                 }
                 .eraseToEffect()
-                .cancellable(id: Timer.self, cancelInFlight: true)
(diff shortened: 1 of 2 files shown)


Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -228,8 +228,8 @@ struct KlaviyoReducer: ReducerProtocol {
             }
             switch networkStatus {
             case .notReachable:
-                state.flushInterval = 0
-                return EffectPublisher.cancel(ids: [RequestId.self, Timer.self])
+                state.flushInterval = Double.infinity
```
evan-masseau, line 231: I read this as double infinity at first... Like infinity wasn't enough, double it.

evan-masseau (approved):
Makes sense!

Sources/KlaviyoSwift/StateManagement.swift
ndurell: 😂

PR #72, opened 2023-03-23
Release version 1.7.2

ndurell:


fastlane/report.xml
```diff
@@ -5,44 +5,42 @@
 
 
 
-      <testcase classname="fastlane.lanes" name="0: default_platform" time="0.000366">
```
ajaysubra, line 8: what are these changes?

ndurell: I think fast lane does this

ajaysubra: okay.

ajaysubra (approved):

PR #75, opened 2023-03-24
Create .github/workflows/stale.yml

ndurell:


.github/workflows/stale.yml
```diff
@@ -0,0 +1,25 @@
...
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-issue-stale: -1
+        days-before-issue-close: -1
+        days-before-pr-stale: 16
+        days-before-pr-close: 14
+        stale-pr-message: 'This PR has not seen any updates in the last 16 days. Without further action this PR will be closed in 14 days. To disable further staleness checks add the `evergreen` label.'
+        close-pr-message: 'This PR has seen no updates in the past 30 days and will now be closed. Restore the branch and create a new PR if you wish to continue work.'
+        stale-pr-label: 'stale'
+        exempt-pr-labels: 'evergreen'
```
ajaysubra, line 24: Wonder what would be an example of evergreen?

ajaysubra (approved):

.github/workflows/stale.yml
ndurell: Something you want to keep around for a while. This is from the app repo.

PR #76, opened 2023-03-24
Pulling in new api in sample app

ajaysubra:
⚠️ Merge only after #70 ⚠️
⚠️ Make sure to update the SDK from using branch to a version number once released ⚠️ 

This is still WIP, I went through and updated all the old API calls to the new one, but will review this on Monday. The idea is to know what the migration is, be able to write the migration script and merge this PR in when the updating to the new API happen.

### DIFF
--- a/Examples/KlaviyoSwiftExamples/CocoapodsExample/Podfile
+++ b/Examples/KlaviyoSwiftExamples/CocoapodsExample/Podfile
@@ -2,5 +2,5 @@ platform :ios, '13.0'
 use_frameworks!
 
 target 'CocoapodsExample' do
-  pod 'KlaviyoSwift', '~> 1.7.1'
+  pod 'KlaviyoSwift', '~> 2.0.1'
 end
--- a/Examples/KlaviyoSwiftExamples/CocoapodsExample/Podfile.lock
+++ b/Examples/KlaviyoSwiftExamples/CocoapodsExample/Podfile.lock
@@ -1,10 +1,10 @@
 PODS:
   - AnyCodable-FlightSchool (0.6.7)
-  - KlaviyoSwift (1.7.1):
+  - KlaviyoSwift (2.0.1):
     - AnyCodable-FlightSchool
 
 DEPENDENCIES:
-  - KlaviyoSwift (~> 1.7.1)
+  - KlaviyoSwift (~> 2.0.1)
 
 SPEC REPOS:
   trunk:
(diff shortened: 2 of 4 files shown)


Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift
```diff
@@ -34,14 +34,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         window?.makeKeyAndVisible()
 
         // STEP2: Setup Klaviyo SDK with api key
-        Klaviyo.setupWithPublicAPIKey(apiKey: "magpcN")
+        KlaviyoSDK().initialize(with: "magpcN")
```
ndurell, line 37: Did you want to change this back to apiKey?

Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift
```diff
@@ -34,14 +34,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
 
         // EXAMPLE: of how to track an event
-        Klaviyo.sharedInstance.trackEvent(eventName: "Opened kLM App")
+        KlaviyoSDK().create(event: .init(name: .CustomEvent("Opened kLM App")))
```
ndurell, line 40: You could keep an instance around if you want too.

Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift
```diff
@@ -182,15 +170,19 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+// STEP6: Add this extension on AppDelegate for additional push notifications handling
 extension AppDelegate: UNUserNotificationCenterDelegate {
     // below method will be called when the user interacts with the push notification
     func userNotificationCenter(
         _ center: UNUserNotificationCenter,
         didReceive response: UNNotificationResponse,
         withCompletionHandler completionHandler: @escaping () -> Void) {
-        Klaviyo.sharedInstance.handlePush(userInfo: response.notification.request.content.userInfo as NSDictionary)
-        completionHandler()
+        // If this notifiation is Klaviyo's notification we'll handle it
```
ndurell, line 180: Nice!

ndurell (approved):
Looks good to me.

Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift
ajaysubra: What is the best practice we want to suggest to our users? Keep an instance or use `KlaviyoSDK()` whenever they want to access the SDK.

Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift
ajaysubra: I personally think it reads better as `KlaviyoSDK().initialize(with: "magpcN")` but it's not clear as to what initializing the SDK with.  Although, `KlaviyoSDK().initialize(apiKey: "magpcN")` doesn't read as well it's clear as to what the SDK requires to initialize.

PR #81, opened 2023-04-27
add patch to support universal links

ndurell:
# Description
This changes the sdk to allow developers to provide an optional deep link handler which when included will not automatically follow urls.

# Check List

- [x] Have you verified that your changes are compitable with all the operating system version this SDK currently supports?

# Manual Test Plan

1. Tested with the example app to verify it does not break compatible with the old api.
2. Test with the test app to ensure universal opens properly using the debugger.

# Supporting Materials

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,19 +243,35 @@ There are two use cases for deep linking that can be relevant here:
 1. When you push a notification to your app with a deep link.
 2. Any other cases where you may want to deep link into your app via SMS, email, web browser etc.
 
-Note that Klaviyo doesn't officially support universal links yet, but since there is no validation on the klaviyo front end for URI schemes, you can include universal links in your push notifications. Ensuring that Klaviyo push works to your expectations with universal links will be the responsibility of your developers.
-
 In order for deep linking to work, there are a few configurations that are needed and these are no different from what are required for handling deep linking in general and [Apple documentation](https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app) on this can be followed in conjunction with the steps highlighted here:
 
-### Step 1: Register the URL scheme
+### Option 1: Modify Open Tracking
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -243,19 +243,35 @@ There are two use cases for deep linking that can be relevant here:
...
+            }
+            completionHandler()
+        }
+    }
+
+```
+Note that the deep link handler will be called back on the main thread. If you want to handle uri schemes in addition to universal links you implement them as described below.
+
+### Option 2: Use URL Schemes
+If you do not need universal link support you can instead implement url schemes for your app and the deepLinkHandler as indicated in Option 1 can be omitted. The Klaviyo SDK will follow all url automatically in this case.
```
evan-masseau, line 264: I think this is a good way to document it.

evan-masseau (approved):
Yeah looks right to me!

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -466,14 +475,18 @@ public struct KlaviyoSDK {
...
                 await MainActor.run {
                     if let url = properties["url"] as? String, let url = URL(string: url) {
-                        UIApplication.shared.open(url)
+                        if let deepLinkHandler = deepLinkHandler {
+                            deepLinkHandler(url)
+                        } else {
+                            UIApplication.shared.open(url)
+                        }
                     }
                     completionHandler()
```
evan-masseau, line 491: Maybe just a nit, but does it make a difference that you're running the completion handler on main thread here vs how you left it to the dev to call completionHandler before?

ndurell: This part was always on the main thread. If you are talking about the code above though I think it should have been on the main thread before. I suspect that function is always called on the main thread anyway but putting it back on it is just a safety measure since devs are likely to call ui related code in the callback.

evan-masseau (approved):

PR #86, opened 2023-06-02
Add device metadata to events' payload

evan-masseau:
# Description
Added device metadata in the same manner as Kenny's android PR [here](https://github.com/klaviyo/klaviyo-android-sdk/pull/76/files) 

This will make sure that all events from the Swift SDK are tagged with metadata. 

# Check List

- [ ] Have you verified that your changes are compitable with all the operating system version this SDK currently supports?

# Manual Test Plan

1. Updated the unit tests

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/AppContextInfo.swift
+++ b/Sources/KlaviyoSwift/AppContextInfo.swift
@@ -24,6 +24,8 @@ private let DEFAULT_DEVICE_MODEL: String = {
     return String(cString: machine)
 }()
 
+private let DEVICE_ID_STORE_KEY = "_klaviyo_device_id"
+
 struct AppContextInfo {
     let executable: String
     let bundleId: String
@@ -34,6 +36,7 @@ struct AppContextInfo {
     let osName: String
     let manufacturer: String
     let deviceModel: String
+    let deviceId: String
 
     var osVersion: String {
         "\(version.majorVersion).\(version.minorVersion).\(version.patchVersion)"
@@ -51,7 +54,8 @@ struct AppContextInfo {
          version: OperatingSystemVersion = DEFAULT_OS_VERSION,
          osName: String = DEFAULT_OS_NAME,
          manufacturer: String = DEFAULT_MANUFACTURER,
-         deviceModel: String = DEFAULT_DEVICE_MODEL) {
+         deviceModel: String = DEFAULT_DEVICE_MODEL,
+         deviceId: String? = nil) {
         self.executable = executable
         self.bundleId = bundleId
         self.appVersion = appVersion
@@ -61,5 +65,14 @@ struct AppContextInfo {
         self.osName = osName
         self.manufacturer = manufacturer
         self.deviceModel = deviceModel
+
+        if let _deviceId = deviceId {
+            self.deviceId = _deviceId
+        } else if let _deviceId = UserDefaults.standard.string(forKey: DEVICE_ID_STORE_KEY) {
+            self.deviceId = _deviceId
+        } else {
+            self.deviceId = UUID().uuidString
+            UserDefaults.standard.set(self.deviceId, forKey: DEVICE_ID_STORE_KEY)
+        }
     }
 }
--- a/Sources/KlaviyoSwift/AppContextInfo.swift
+++ b/Sources/KlaviyoSwift/AppContextInfo.swift
@@ -12,16 +12,29 @@ private let DEFAULT_EXECUTABLE: String = (info?["CFBundleExecutable"] as? String
 private let DEFAULT_BUNDLE_ID: String = info?["CFBundleIdentifier"] as? String ?? "Unknown"
 private let DEFAULT_APP_VERSION: String = info?["CFBundleShortVersionString"] as? String ?? "Unknown"
 private let DEFAULT_APP_BUILD: String = info?["CFBundleVersion"] as? String ?? "Unknown"
+private let DEFAULT_APP_NAME: String = info?["CFBundleName"] as? String ?? "Unknown"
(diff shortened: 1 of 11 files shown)


klaviyoit:
https://klaviyo.tpondemand.com/entity/175638

evan-masseau:
oops, i guess i don't have the precommit installed.

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -49,9 +49,21 @@ public struct Event: Equatable {
 
     public let metric: Metric
     public var properties: [String: Any] {
-        _properties.value as! [String: Any]
+        (_properties.value as! [String: Any]).merging([
```
ndurell, line 52: ooh merging....

Sources/KlaviyoSwift/AppContextInfo.swift
```diff
@@ -30,17 +43,24 @@ struct AppContextInfo {
         "\(osName) \(osVersion)"
     }
 
-    init(excutable: String = DEFAULT_EXECUTABLE,
```
ndurell, line 33: spelling is hard 🤦🏻

Sources/KlaviyoSwift/KlaviyoModels.swift
evan-masseau: is that.. ok?

ndurell: Yup I just didn't know it existed haha.

ndurell (approved):

Sources/KlaviyoSwift/InternalAPIModels.swift
```diff
@@ -121,8 +123,23 @@ extension KlaviyoAPI.KlaviyoRequest {
                     let uniqueId: String
                     init(attributes: KlaviyoSwift.Event,
                          anonymousId: String? = nil) {
+                        let context = KlaviyoAPI.KlaviyoRequest._appContextInfo
```
evan-masseau, line 126: @ndurell FYI I moved where I appended the metadata to the event from the Event model to this "Internal" model, that seemed more appropriate.

evan-masseau:
TODO just noticed we're missing device identifier too, will push another update today

Sources/KlaviyoSwift/AppContextInfo.swift
```diff
@@ -30,17 +46,33 @@ struct AppContextInfo {
...
         self.appBuild = appBuild
+        self.appName = appName
         self.version = version
         self.osName = osName
+        self.manufacturer = manufacturer
+        self.deviceModel = deviceModel
+
+        if let _deviceId = deviceId {
+            self.deviceId = _deviceId
+        } else if let _deviceId = UserDefaults.standard.string(forKey: DEVICE_ID_STORE_KEY) {
```
evan-masseau, line 71: @ajaysubra and @ndurell let me know if this looks like an appropriate way to save our own device UUID for the lifetime of the app install. Namely wondering if you're cool with using user defaults here. 

I was hesitant to use identifierForVendor since it can be duplicated across apps, if the vendor has multiple apps installed on the device.

ajaysubra: 1. minor nit - the unwrapped variable's name can be the same as the variable name of the optional variable in swift.
2. About using `UserDefaults` I mostly try and avoid using it since it's very specific to storing any default setting the user makes like using Celsius over Fahrenheit etc. But I do see your point about using `identifierForVendor`. Reading apple's documentation on this, seems like they intentionally made `identifierForVendor`  limited to the lifetime of the app for privacy reasons. Should we just remain compliant with what apple recommends just to be safe and not risk user customers apps getting reviewed and flagged for doing something outside of what they recommend? I know I'm probably over thinking this but in my experience apple's review can be nit picky at times.

Also, I'm curious what is the issue if the device ID is same across different apps from the same vendor? I would expect it to be the same since it's the same device no?

P.S - found [this](https://nshipster.com/device-identifiers/)  article interesting especially the part about apple building the DeviceCheck framework for Uber.

evan-masseau: I guess i could re-consider my data structure on the backend, but I've defined device ID as unique in the data model, so if there's a clear way to wind up with duplicates (e.g. if a company uses us for two different apps), that presents a bit of a problem on the backend right now.

evan-masseau: Actually i might just be overthinking it.... A company can only have 1 app configured with us now anyway. Whenever we change that, we'll probably have a variety of required changed to the metadata model anyways.

evan-masseau: And yea, in the future it might be _nice_ to be able to link devices across apps if we had a bigger client with multiple apps. Ok gonna change my approach then.

ajaysubra: Yeah, that's a good plan. I think the biggest use cases of multi app users are marketing agencies. When I was at constant contact our iOS app supported multi-user experience where marketing agencies would have multiple accounts setup on our app and switch between them. It was a pain the neck to maintain but was a good lesson in building for scale.

ndurell: Yah slightly concerned about this - but at the same time maybe it's ok. In the case where a customer had multiple apps are we ok with this having the same id? I think it would be ok since we also use company id and we only allow one  app per company. If we started to allow multiple apps per company i'm wondering if it means we would have one device entry for both apps.

evan-masseau:
Closed to consolidate this PR with all other metadata changes.

PR #89, opened 2023-06-16
Added v3 token API containing metadata, including environment

evan-masseau:
# Description

- Replaces V2 identify API with V3 push tokens API call, including metadata
- Added collection of the debug/release environment 
- Collecting the push authorization state and background processing authorization 

# Check List

- [x] Are you changing anything with the public API? [NO]
- [x] Have you verified that your changes are compitable with all the operating system version this SDK currently supports?

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/AppContextInfo.swift
+++ b/Sources/KlaviyoSwift/AppContextInfo.swift
@@ -24,6 +24,8 @@ private let DEFAULT_DEVICE_MODEL: String = {
     return String(cString: machine)
 }()
 
+private let DEVICE_ID_STORE_KEY = "_klaviyo_device_id"
+
 struct AppContextInfo {
     let executable: String
     let bundleId: String
@@ -34,6 +36,7 @@ struct AppContextInfo {
     let osName: String
     let manufacturer: String
     let deviceModel: String
+    let deviceId: String
 
     var osVersion: String {
         "\(version.majorVersion).\(version.minorVersion).\(version.patchVersion)"
@@ -51,7 +54,8 @@ struct AppContextInfo {
          version: OperatingSystemVersion = DEFAULT_OS_VERSION,
          osName: String = DEFAULT_OS_NAME,
          manufacturer: String = DEFAULT_MANUFACTURER,
-         deviceModel: String = DEFAULT_DEVICE_MODEL) {
+         deviceModel: String = DEFAULT_DEVICE_MODEL,
+         deviceId: String? = nil) {
         self.executable = executable
         self.bundleId = bundleId
         self.appVersion = appVersion
@@ -61,5 +65,14 @@ struct AppContextInfo {
         self.osName = osName
         self.manufacturer = manufacturer
         self.deviceModel = deviceModel
+
+        if let _deviceId = deviceId {
+            self.deviceId = _deviceId
+        } else if let _deviceId = UserDefaults.standard.string(forKey: DEVICE_ID_STORE_KEY) {
+            self.deviceId = _deviceId
+        } else {
+            self.deviceId = UUID().uuidString
+            UserDefaults.standard.set(self.deviceId, forKey: DEVICE_ID_STORE_KEY)
+        }
     }
 }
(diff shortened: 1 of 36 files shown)


Sources/KlaviyoSwift/AppContextInfo.swift
```diff
@@ -30,17 +48,76 @@ struct AppContextInfo {
...
+
+    private func provisioningProfile() throws -> [String: Any]? {
+        guard let url = Bundle.main.url(forResource: "embedded", withExtension: "mobileprovision") else {
+            return nil
+        }
+
+        let binaryString = try String(contentsOf: url, encoding: .isoLatin1)
+
+        let scanner = Scanner(string: binaryString)
+        guard scanner.scanUpToString("<plist") != nil, let plistString = scanner.scanUpToString("</plist>"),
```
ndurell, line 115: hmm

evan-masseau: Yeah gotta test this on an ad hoc build to make sure it works as advertised

Sources/KlaviyoSwift/AppContextInfo.swift
```diff
@@ -30,17 +48,76 @@ struct AppContextInfo {
...
+            #else
+            environment = "release"
+            #endif
+        }
+    }
+}
+
+extension UIDevice {
+    public enum PushEnvironment: String {
+        case unknown
```
jordan-griffin, line 89: I know nothing on how this works, but I'm wondering why we would want an `unknown` environment. If APNS gives us nothing, can we assume thats a device running on prod and default to that?

evan-masseau: i used this as a fallback to then use the DEBUG flag

ndurell: I think this is indicating we can't determine the type.

jordan-griffin: Aaah I see above. Thanks - makes perfect sense to me

ndurell: Err the environment.

ndurell: The question is what would we do with these tokens?

evan-masseau: I suspect we'd only hit this if the app that doesn't use push notifications at all. In which case I'll just fall back on using the DEBUG compiler flag. 
As far as what we send the server I'll still only ever send `debug` or `release`

ndurell: ok cool

Tests/KlaviyoSwiftTests/__Snapshots__/EncodableTests/testKlaviyoState.1.json
```diff
@@ -7,17 +7,50 @@
...
+                "data" : {
+                  "attributes" : {
+                    "anonymous_id" : "foo",
+                    "email" : "foo",
+                    "phone_number" : "foo",
+                    "properties" : {
+
+                    }
+                  },
+                  "meta" : {
```
ndurell, line 40: Ahh dang this part is different I think....

ndurell: I'm not sure you can put meta in or not I mean

evan-masseau: wondered about that, sad. it was nice to just re-use the same struct

Sources/KlaviyoSwift/InternalAPIModels.swift
```diff
@@ -160,53 +178,105 @@ extension KlaviyoAPI.KlaviyoRequest {
...
-                let email: String?
-                let phoneNumber: String?
-                let externalId: String?
-                struct Append: Equatable, Codable {
-                    let pushToken: String
-                    enum CodingKeys: String, CodingKey {
-                        case pushToken = "$ios_tokens"
-                    }
-                }
+            var type = "push-token"
```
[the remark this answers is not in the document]
evan-masseau: Oh it gives a warning about how this property can't be decoded if declared as a let. Which i guess is "just" a warning but it makes me want to leave it a var.

Sources/KlaviyoSwift/AppContextInfo.swift
evan-masseau: Confirmed ad-hoc build -> "release" woot

evan-masseau:
Do not merge -- currently points to an alpha revision of the V3 API.

evan-masseau:
closing to consolidate PRs

PR #92, opened 2023-06-26
Adding rich push notifications to Klaviyo SDK

ajaysubra:
# Description

This PR is to add a new target that is exported independently using SPM and Cocoapods to users and will be integrated into the notification service extension that users will create on their mobile apps to handle rich push notifications. 

## Some considerations that we considered are - 

1. We decided against including the logic to download and attach image into the core SDK mainly because 
 * The notification service extensions that users create on their apps are granted limited memory by app and hence bundling our SDK into the extension will unnecessarily bloat up the extension size.
 * There are some application methods that are used in our SDK that are not available to access from an extension such as `UIApplication.shared` which is used to open deep links.
2. We decided to move the logic to download image and attach it into the SDK vs. leaving it to our users for the below reasons - 
 * Lesser code for our users to maintain
 * Easily to debug user issues when reported.

### SPM - 

Since the package is already pulled into user's apps that use Klaviyo, they only need to include this in their `NotificationServiceExtension` frameworks and targets section of this target - 

image

### Cocoapods - 

Unfortunately, Cocoapods doesn't support sub specs for extensions - [GH issue](https://github.com/CocoaPods/CocoaPods/issues/7733). Because of this we export the extension logic as a separate pod which user can import in their notification service extension target as below - 

```
target 'NotificationServiceExtension' do
    pod 'KlaviyoSwiftExtension'
end
``` 

^ is separate from the KlaviyoSwift that the main app target imports 

The full pod file would look something like - 

```
platform :ios, '13.0'
use_frameworks!

target 'CocoapodsExample' do
  pod 'KlaviyoSwift'
end

target 'NotificationServiceExtension' do
    pod 'KlaviyoSwiftExtension'
end
```

Once included in the SDK below is how it would be imported and used in the notification service extension - 

```swift
import UIKit
import UserNotifications
import KlaviyoSwiftExtension

// MARK: notification service extension implementation.

/// When push payload is marked as there being mutable-content this service
/// (more specifically the `didReceiveNotificationRequest` ) is called to perform
/// tasks such as downloading images and attaching it to the notification before it's displayed to the user.
///
/// There is a limited time before which `didReceiveNotificationRequest`  needs to wrap up it's operations
/// else the notification is displayed as received.
///
/// Any property from `UNMutableNotificationContent` can be mutated here before presenting the notification.
class NotificationService: UNNotificationServiceExtension {
    var request: UNNotificationRequest!
    var contentHandler: ((UNNotificationContent) -> Void)?
    var bestAttemptContent: UNMutableNotificationContent?

    override func didReceive(
        _ request: UNNotificationRequest,
        withContentHandler contentHandler: @escaping (UNNotificationContent) -> Void) {
            
        self.request = request
        self.contentHandler = contentHandler
        self.bestAttemptContent = (request.content.mutableCopy() as? UNMutableNotificationContent)

        if let bestAttemptContent = bestAttemptContent {
            KlaviyoExtensionSDK().handleNotificationServiceDidReceivedRequest(
                request: self.request,
                bestAttemptContent: bestAttemptContent,
                contentHandler: contentHandler
            )
        }
    }

    override func serviceExtensionTimeWillExpire() {
        /// Called just before the extension will be terminated by the system.
        /// Use this as an opportunity to deliver your "best attempt" at modified content, otherwise the original push payload will be used.
        if let contentHandler = contentHandler,
           let bestAttemptContent = bestAttemptContent {
            KlaviyoExtensionSDK().handleNotificationServiceExtensionTimeWillExpireRequest(
                request: self.request,
                bestAttemptContent: bestAttemptContent,
                contentHandler: contentHandler
            )
        }
    }
}

```


KlaviyoSwift.podspec
```diff
@@ -1,10 +1,10 @@
 Pod::Spec.new do |s|
   s.name             = "KlaviyoSwift"
   s.version          = "2.0.1"
```
ajaysubra, line 3: Although there is no change in the SDK itself should we update the minor version for the rich push release as it's not really a patch since it's a new feature but at the same time it's not breaking change.

KlaviyoSwift.podspec
```diff
@@ -1,10 +1,10 @@
...
-  s.summary          = "Incorporate Klaviyo's event and person tracking functionality into iOS applications"
+  s.summary          = "Incorporate Klaviyo's event and person tracking and push notifications functionality into iOS applications"
 
   s.description      = <<-DESC
-                        Use the Klaviyo SDK to incorporate Klaviyo's event and person tracking functionality within iOS applications. Written in Swift.'
+                        Use the Klaviyo SDK to incorporate Klaviyo's event and person tracking functionality and push notifications within iOS applications. Written in Swift.'
```
ajaysubra, line 7: Made this update to indicate that push notifications is also included now.

KlaviyoSwiftExtension.podspec
```diff
@@ -0,0 +1,18 @@
...
+  s.version          = "2.0.1"
+  s.summary          = "Incorporate Klaviyo's rich push notifications functionality into your iOS applications"
+
+  s.description      = <<-DESC
+                          Incorporate Klaviyo's rich push notifications functionality into your iOS applications. Written in Swift.'
+                       DESC
+
+  s.homepage         = "https://github.com/klaviyo/klaviyo-swift-sdk"
+  s.license          = { :type => "MIT", :file => "LICENSE" }
+  s.author           = { "Mobile @ Klaviyo" => "mobile@klaviyo.com" }
```
ajaysubra, line 12: Is this email valid?

KlaviyoSwiftExtension.podspec
```diff
@@ -0,0 +1,18 @@
+Pod::Spec.new do |s|
+  s.name             = "KlaviyoSwiftExtension"
```
ajaysubra, line 2: This is the new pod that is now exported exclusively for the notification service extension. Refer PR description for more info.

KlaviyoSwiftExtension.podspec
```diff
@@ -0,0 +1,18 @@
...
+                       DESC
+
+  s.homepage         = "https://github.com/klaviyo/klaviyo-swift-sdk"
+  s.license          = { :type => "MIT", :file => "LICENSE" }
+  s.author           = { "Mobile @ Klaviyo" => "mobile@klaviyo.com" }
+  s.source           = { :git => "https://github.com/klaviyo/klaviyo-swift-sdk.git", :tag => s.version.to_s }
+  s.swift_version = '5.7'
+  s.platform = :ios
+  s.ios.deployment_target = '13.0'
+  s.source_files = 'Sources/KlaviyoSwiftExtension/**/*.swift'
```
ajaysubra, line 17: These are the source files for the rich push feature.

Package.swift
```diff
@@ -9,7 +9,10 @@ let package = Package(
     products: [
         .library(
             name: "KlaviyoSwift",
-            targets: ["KlaviyoSwift"])
+            targets: ["KlaviyoSwift"]),
+        .library(
+            name: "KlaviyoSwiftExtension",
```
ajaysubra, line 14: Not really married to this name but I think this follows the existing convention.

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -0,0 +1,139 @@
+//
```
ajaysubra, line 1: This is the main file that has the rich push related code.

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -0,0 +1,139 @@
...
+    /// This method mainly does two things - downloads the media attached in the payload and then attaches it to the push notification.
+    ///
+    /// NOTE that there is no guarantees that the content handler will be called within the iOS stipulated time in the case where the download of the rich media takes long. iOS will automatically
+    /// present the notification as received from APNS when this happens and the image will not be attached to the notification.
+    ///
+    /// - Parameters:
+    ///   - request: the request received in the delegate `didReceive(_:withContentHandler:)`
+    ///   - bestAttemptContent: this is also received in `didReceive(_:withContentHandler:)` and is the best attempt at mutating the APNS payload before attaching it to the push notification
+    ///   - contentHandler: this is also received in `didReceive(_:withContentHandler:)` and is the closure that needs to be called before the time iOS provides for us to mutate the content. This closure will be called with the `bestAttemptContent` once the image is downloaded and attached.
+    public func handleNotificationServiceDidReceivedRequest(
```
ndurell, line 26: thoughts on making these static and ditching the initializer?

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -0,0 +1,139 @@
...
+         contentHandler: @escaping (UNNotificationContent) -> Void
+    ) {
+        // 1a. get the rich media url from the push notification payload
+        guard let imageURLString = bestAttemptContent.userInfo["rich-media"] as? String else {
+            contentHandler(bestAttemptContent)
+            return
+        }
+
+        // 1b.falling back to .png in case the media type isn't sent from the server.
+        let imageTypeString = bestAttemptContent.userInfo["rich-media-type"] as? String ?? "png"
```
ndurell, line 38: Default to png? Maybe should be a parameter with a default so they can override it?

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -0,0 +1,139 @@
...
+            }
+
+            guard let file = file else {
+                completion(nil)
+                return
+            }
+
+            completion(file)
+        }
+        task.resume()
```
ndurell, line 101: We don't need to hold onto the task right?

ndurell: Was also thinking should we try to use async api where available?

ndurell (approved):

KlaviyoSwiftExtension.podspec
ndurell: I think so yah! We get emails to it sometimes.

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
ajaysubra: Good call. Done. Also, swift format changed the struct to an enum with [this](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#enumnamespaces) rule:
image

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
ajaysubra: Yeah we don't. I removed it. With using async await - I'll probably have it as a fast follow since we've already tested this code now.

KlaviyoSwift.podspec
evan-masseau: yes

evan-masseau: If there's one place I'd lean toward syncing version numbers, its the test apps with their accompanying SDK version

evan-masseau: of course we still can't be perfect, like if we overhauled the UI that'd warrant a version increment, even if the SDK didn't change, so its not a perfect science

evan-masseau: wow ignore all of that, thought i was looking at the test app PR 😮‍💨

evan-masseau: Realizing now we've talked about this since you wrote this comment. But yes I do think rich push should increment the minor version, as a new feature.

evan-masseau (approved):

Package.swift
jordan-griffin: Would there be a situation where we need multiple extensions? Maybe we want to be more clear what this extension does if its just going to be for handling rich push. Otherwise, I think the name seems fine

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -0,0 +1,133 @@
...
+#if !os(macOS)
+import Foundation
+import UserNotifications
+
+public enum KlaviyoExtensionSDK {
+    /// Call this method when you receive a rich push notification in the notification service extension.
+    /// This method should be called from within `didReceive(_:withContentHandler:)` method of `UNNotificationServiceExtension`.
+    /// This method mainly does two things - downloads the media attached in the payload and then attaches it to the push notification.
+    ///
+    /// NOTE that there is no guarantees that the content handler will be called within the iOS stipulated time in the case where the download of the rich media takes long. iOS will automatically
```
jordan-griffin, line 16: ```NOTE that there is no guarantee that the content handler will be called with in the time stipulated by iOS to download the rich media successfully. In the case where the download does not complete, iOS will automatically present the notification as received from APNS without the attached image```

Just thought that sounded a little clearer. What do you think?

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -0,0 +1,133 @@
...
+    ///
+    /// - Parameters:
+    ///   - request: the request received in the delegate `didReceive(_:withContentHandler:)`
+    ///   - bestAttemptContent: this is also received in `didReceive(_:withContentHandler:)` and is the best attempt at mutating the APNS payload before attaching it to the push notification
+    ///   - contentHandler: this is also received in `didReceive(_:withContentHandler:)` and is the closure that needs to be called before the time iOS provides for us to mutate the content. This closure will be called with the `bestAttemptContent` once the image is downloaded and attached.
+    public static func handleNotificationServiceDidReceivedRequest(
+        request: UNNotificationRequest,
+        bestAttemptContent: UNMutableNotificationContent,
+        contentHandler: @escaping (UNNotificationContent) -> Void) {
+        // 1a. get the rich media url from the push notification payload
```
jordan-griffin, line 27: I actually love these little step comments haha

jordan-griffin (approved):

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
ajaysubra: Thanks. Reading what I wrote again now makes me want to go hide under a rock!

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
ajaysubra: Good call. Done!

Package.swift
ajaysubra: Yeah @evan-masseau and I were having the same conversation earlier today and decided to keep the name as is since even in a future where we support multiple extensions we may not want to create a new target for each but instead use this as a general purpose target to support all extensions related support outside of the core SDK.

KlaviyoSwift.podspec
ajaysubra: Ignored. It's been a long day :)

Package.swift
```diff
@@ -23,7 +26,12 @@ let package = Package(
     targets: [
         .target(
             name: "KlaviyoSwift",
-            dependencies: [.product(name: "AnyCodable", package: "AnyCodable")]),
+            dependencies: [.product(name: "AnyCodable", package: "AnyCodable")],
+            path: "Sources/KlaviyoSwift"),
```
ndurell, line 30: Actually what's this about?

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -0,0 +1,134 @@
+//
+//  KlaviyoExtension.swift
+//
+//
+//  Created by Ajay Subramanya on 6/23/23.
+//
+#if !os(macOS)
+import Foundation
```
ndurell, line 8: foundation is not on macos?

ajaysubra: when running the package tests using xcodebuild on CLI there wasn't a way to indicate to run only on iOS and this was a suggestion I found online. I did notice that the package.swift file did have the platform in there but still I was getting compile error about something in here not supported. Checking now to see what happens when I remove this.

Package.swift
ajaysubra: the path?

ndurell: yes.

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
ajaysubra: removed this now. looks like somehow the package defaults and build for a mac target even though the package clearly has the platform as iOS.

Package.swift
ajaysubra:
```
///   - path: The custom path for the target. By default, the Swift Package Manager requires a target's sources to reside at predefined search paths;
///       for example, `[PackageRoot]/Sources/[TargetName]`.
///       Don't escape the package root; for example, values like `../Foo` or `/Foo` are invalid.
```

I think if the name of the target is the same as the name of the directory under `/Source` we don't need this but now that we have two targets and products I was just being more explicit about the path of the source files.

ajaysubra: https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#product

PR #91, opened 2023-06-22
V3 API updates and Device Metadata

evan-masseau:
# Description
- Adds device metadata collection
- Tag all events created by Swift SDK with device metadata.
- Upgrades from using the V2 identify API for push tokens to new V3 Push Tokens API. 
- [ ] TODO - Update the V3 Event API payload structure -- See https://github.com/klaviyo/klaviyo-swift-sdk/pull/93

# Check List

- [x] Are you changing anything with the public API? No
- [x] Have you tested this change on real device? Yes
- [x] Are your changes backwards compatible with previous SDK Versions? Yes
- [x] Have you added unit test coverage for your changes? Yes
- [x] Have you verified that your changes are compatible with all the operating system version this SDK currently supports? Yes

# Manual Test Plan

1. Updated unit tests
2. Tested with tester app against the alpha API release on production
3.

PR #93, opened 2023-06-27
Updates to 7-15 revision data structures

evan-masseau:
Here is what I think the API payloads need to look like for upcoming 7-15 V3 API changes.
Note the PR's target branch: this is on top of this pending PR #91 
I tested these changes against prod on the VPN so that I could target 7-15.pre revision

### DIFF
--- a/Sources/KlaviyoSwift/Version.swift
+++ b/Sources/KlaviyoSwift/Version.swift
@@ -7,5 +7,6 @@
 
 import Foundation
 
-public let __klaviyoSwiftName = "klaviyo-swift-sdk"
+// public let __klaviyoSwiftName = "klaviyo-swift-sdk"
+public let __klaviyoSwiftName = "swift"
 public let __klaviyoSwiftVersion = "2.0.1"
--- a/Tests/KlaviyoSwiftTests/__Snapshots__/EncodableTests/testEventPayload.1.json
+++ b/Tests/KlaviyoSwiftTests/__Snapshots__/EncodableTests/testEventPayload.1.json
@@ -12,14 +12,16 @@
       "profile" : {
         "data" : {
           "attributes" : {
-            "anonymous_id" : "",
+            "anonymous_id" : "anon-id",
+            "email" : "blob@email.com",
             "properties" : {
 
             }
           },
           "meta" : {
             "identifiers" : {
-              "anonymous_id" : ""
+              "anonymous_id" : "anon-id",
+              "email" : "blob@email.com"
             }
           },
           "type" : "profile"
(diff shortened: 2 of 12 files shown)


Tests/KlaviyoSwiftTests/__Snapshots__/EncodableTests/testEventPayload.1.json
```diff
@@ -2,14 +2,28 @@
...
+          "type" : "metric"
+        }
       },
       "profile" : {
-        "email" : "blob@email.com",
-        "location" : {
-          "city" : "blob city"
-        },
-        "stuff" : 2
+        "data" : {
```
evan-masseau, line 13: oh that didnt go as planned

Sources/KlaviyoSwift/Version.swift
```diff
@@ -7,5 +7,6 @@
 
 import Foundation
 
-public let __klaviyoSwiftName = "klaviyo-swift-sdk"
+// public let __klaviyoSwiftName = "klaviyo-swift-sdk"
```
evan-masseau, line 10: temporary until a fix is deployed to prod

Tests/KlaviyoSwiftTests/__Snapshots__/EncodableTests/testEventPayload.1.json
```diff
@@ -2,14 +2,30 @@
...
+          "attributes" : {
+            "anonymous_id" : "anon-id",
+            "email" : "blob@email.com",
+            "properties" : {
+
+            }
+          },
+          "meta" : {
+            "identifiers" : {
```
jason-myers-klaviyo, line 22: meta identifiers is deprecated no longer needed.  You only have to pas the top part now (attributes with identifiers in them, like you have above)

evan-masseau: Oh weird I was just testing against the 7-15 alpha events API and it seemed to be necessary, will look again. 

(also damn that was a quick response, are you omniscient?)

jason-myers-klaviyo: Ah, the change was made to /client/profiles but perhaps /client/events/ was not updated to match

I will investigate and see about removing it from there too.  In the meantime if it seems required, go ahead and leave it in here

jason-myers-klaviyo: Oh, they are removing it but it is still open PR: https://github.com/klaviyo/app/pull/51575

The release is the same as yours though (7-15)

evan-masseau: Awesome, I'm sure that means we shouldn't have our new push tokens API use meta identifiers either then. So glad you were watching this repo thanks!

Issue #100, opened 2023-08-09
Client Push token API is failing when background field is UNAVAILABLE


sairamkotha:
### Description

Push Tokens are not registering on the dashboard when the background refresh is a UNAVAILABLE field inside the push tokens API.
**API:**
https://a.klaviyo.com/client/push-tokens/?company_id=XXXXXX
Request body:
```
{
 {

> "background": "UNAVAILABLE",

      "vendor": "APNs",
      "platform": "ios",
      "enablement_status": "AUTHORIZED",
      "profile": {
        "data": {
          "attributes": {
            "anonymous_id": "41E5D174-6970-4064-8CDA-31132864ABD5",
            "properties": {}
          },
          "type": "profile"
        }
      }
    },
    "type": "push-token"
  }
}

```
**Response:**
```
{
  "errors": [
    {
      "id": "5dc13942-2156-4b17-a2a2-b3b3bd75d858",
      "status": 400,
      "code": "invalid",
      "title": "Invalid input.",
      "detail": "'UNAVAILABLE' is not a valid choice for 'background'.",
      "source": {
        "pointer": "/data/attributes/background"
      },
      "meta": {}
    }
  ]
}
```

Based on the Swift code:
This enum returns UNAVAILABLE and API is returning 400 status code.
enum PushBackground: String {
        case available = "AVAILABLE"
        case unavailable = "UNAVAILABLE"

        static func create(from status: UIBackgroundRefreshStatus) -> PushBackground {
            switch status {
            case .available:
                return PushBackground.available
            default:
                return PushBackground.unavailable
            }
        }
    }

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-swift-sdk/issues) or [discussion](https://github.com/pointfreeco/klaviyo/klaviyo-swift-sdk/discussions).

### Expected behavior

Push tokens API shouldn't fail for background request status

### The Klaviyo Swift SDK version information

2.2.0 version

### Destination operating system

iOS 16.4

### Xcode version information

Xcode: 14.3.1

ndurell:
Hello!
[...]

PR #101, opened 2023-08-09
Add two cases for Push Background, denied & restricted 

sairamkotha:
# Description
Currently, Push tokens api fails when we send unavailable in that API. I have mentioned more details in this issue
https://github.com/klaviyo/klaviyo-swift-sdk/issues/100

# Check List

# Manual Test Plan

1. I used proxy tools and tested by sending background field as restricted or denied or available and API is returning 202 status code.

# Supporting Materials


ndurell (approved):
Looks good! Thanks for fixing!

PR #103, opened 2023-08-09
fill in identifiers for events

ndurell:
# Description
In updating the sdk to support the new event api changes we appear to have dropped the ids from some events in certain cases. It also looks like customer properties on events were dropped as well. This is now fixed and should get applied properly. Coincident with this PR we started seeing flakiness in the CI build. I've added some extra timeouts to help with that. Ideally though we should investigate this more deeply to determine the cause.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -189,8 +189,7 @@ public class Klaviyo: NSObject {
            let body = properties["body"] as? [String: Any], let _ = body["_k"] {
             Self.sdkInstance
                 .create(event: Event(name: .OpenedPush,
-                                     properties: properties,
-                                     profile: [:]))
+                                     properties: properties))
             if let url = properties["url"] as? String, let url = URL(string: url) {
                 Task {
                     await MainActor.run {
@@ -481,7 +480,7 @@ public struct KlaviyoSDK {
     public func handle(notificationResponse: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void, deepLinkHandler: ((URL) -> Void)? = nil) -> Bool {
         if let properties = notificationResponse.notification.request.content.userInfo as? [String: Any],
            let body = properties["body"] as? [String: Any], let _ = body["_k"] {
-            create(event: Event(name: .OpenedPush, properties: properties, profile: [:]))
+            create(event: Event(name: .OpenedPush, properties: properties))
             Task {
                 await MainActor.run {
                     if let url = properties["url"] as? String, let url = URL(string: url) {
--- a/Sources/KlaviyoSwift/KlaviyoModels.swift
+++ b/Sources/KlaviyoSwift/KlaviyoModels.swift
@@ -54,7 +54,7 @@ public struct Event: Equatable {
 
     private let _properties: AnyCodable
     public var profile: [String: Any] {
-        _profile.value as! [String: Any]
+        _profile.value as? [String: Any] ?? [:]
     }
 
     internal var _profile: AnyCodable
@@ -63,8 +63,6 @@ public struct Event: Equatable {
     public let uniqueId: String
     public let identifiers: Identifiers?
 
-    @available(
-        iOS, deprecated: 9999, message: "Please use alternative initializer.")
     public init(name: EventName,
                 properties: [String: Any]? = nil,
                 identifiers: Identifiers? = nil,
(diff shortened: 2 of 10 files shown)


jordan-griffin (approved):

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -189,8 +189,7 @@ public class Klaviyo: NSObject {
            let body = properties["body"] as? [String: Any], let _ = body["_k"] {
             Self.sdkInstance
                 .create(event: Event(name: .OpenedPush,
-                                     properties: properties,
-                                     profile: [:]))
```
ndurell, line 193: This was not needed, was already an optional parameter.

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -54,7 +54,7 @@ public struct Event: Equatable {
 
     private let _properties: AnyCodable
     public var profile: [String: Any] {
-        _profile.value as! [String: Any]
+        _profile.value as? [String: Any] ?? [:]
```
ndurell, line 57: I started seeing a crash here so I made this optional.

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -70,7 +70,16 @@ public struct Event: Equatable {
                 value: Double? = nil,
                 time: Date? = nil,
                 uniqueId: String? = nil) {
-        _profile = AnyCodable(profile ?? [:])
+        var profile = profile
```
ndurell, line 73: This is probably temporary. We move some old parameters from the profile dictionary to the identifers structure so it gets picked up in the api request later. If any of these are absent they were be overwritten by the state value or they will overwrite state.

Tests/KlaviyoSwiftTests/APIRequestErrorHandlingTests.swift
```diff
@@ -9,6 +9,8 @@
 import Foundation
 import XCTest
 
+let TIMEOUT_NANOSECONDS: UInt64 = 10_000_000_000 // 10 seconds
```
ndurell, line 12: This test become flakey so I added some timeout here to reduce flakiness.

Tests/KlaviyoSwiftTests/KlaviyoStateTests.swift
```diff
@@ -45,7 +45,7 @@ final class KlaviyoStateTests: XCTestCase {
     let TEST_INVALID_PROPERTIES_EVENT = [
         "properties": [
             1: "propValue"
-        ],
+        ] as [AnyHashable: String],
```
ndurell, line 48: Cleans up a warning.

Tests/KlaviyoSwiftTests/KlaviyoTestUtils.swift
```diff
@@ -156,7 +157,7 @@ extension StateChangePublisher {
     static let test = { () -> StateChangePublisher in
         StateChangePublisher.debouncedPublisher = { publisher in
             publisher
-                .debounce(for: .seconds(0), scheduler: DispatchQueue.main)
+                .debounce(for: .seconds(0), scheduler: DispatchQueue.immediate)
```
ndurell, line 160: I noticed i had used the main scheduler here. Using immediate should help speed things up.

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -79,7 +79,7 @@ class StateManagementTests: XCTestCase {
         stateChangeSubject.send(completion: .finished)
         lifecycleSubject.send(completion: .finished)
 
-        wait(for: [stateChangeIsSubscribed, lifecycleExpectation], timeout: 1.0)
+        await fulfillment(of: [stateChangeIsSubscribed, lifecycleExpectation])
```
ndurell, line 82: fixes a warning.

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -382,9 +382,23 @@ class StateManagementTests: XCTestCase {
...
         _ = await store.send(.enqueueEvent(event)) {
-            $0.email = "foo"
+            $0.email = "foo@foo.com"
+            $0.phoneNumber = "666BLOB"
+            $0.externalId = "my_user_id"
+            try $0.enqueueRequest(request: .init(apiKey: XCTUnwrap($0.apiKey), endpoint: .createEvent(.init(data: .init(event: event, anonymousId: XCTUnwrap($0.anonymousId))))))
+        }
+    }
+
+    func testEnqueueEventWithIdentifiers() async throws {
```
ndurell, line 394: Added another equivalent test that uses newer api structures.

Tests/KlaviyoSwiftTests/__Snapshots__/EncodableTests/testEventPayload.1.json
```diff
@@ -15,7 +15,11 @@
             "anonymous_id" : "anon-id",
             "email" : "blob@email.com",
             "properties" : {
-
+              "email" : "blob@email.com",
```
ndurell, line 18: This is expected and should have been caught earlier when we updated the api.

Tests/KlaviyoSwiftTests/__Snapshots__/KlaviyoStateTests/testLoadWithRequestMigration.1.json
```diff
@@ -26,7 +26,7 @@
                       "email" : "value",
                       "external_id" : "value",
                       "properties" : {
-
+                        "foo" : "bar"
```
ndurell, line 29: Same here we would have dropped some properties in migrated requests.

PR #115, opened 2023-10-19
fix access of identifiers

ndurell:
# Description

Reported in slack we don't have the right access for Event Identifiers.

# Check List

- [X ] Are you changing anything with the public API?
- [ X] Are your changes backwards compatible with previous SDK Versions?

# Manual Test Plan

1. Run it in the test app before release.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoModels.swift
+++ b/Sources/KlaviyoSwift/KlaviyoModels.swift
@@ -38,9 +38,9 @@ public struct Event: Equatable {
         public let email: String?
         public let phoneNumber: String?
         public let externalId: String?
-        init(email: String? = nil,
-             phoneNumber: String? = nil,
-             externalId: String? = nil) {
+        public init(email: String? = nil,
+                    phoneNumber: String? = nil,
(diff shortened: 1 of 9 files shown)


Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -38,9 +38,9 @@ public struct Event: Equatable {
         public let email: String?
         public let phoneNumber: String?
         public let externalId: String?
-        init(email: String? = nil,
-             phoneNumber: String? = nil,
-             externalId: String? = nil) {
+        public init(email: String? = nil,
+                    phoneNumber: String? = nil,
```
jordan-griffin, line 42: total nit, but the extra indentation seems like a bit much?

jordan-griffin (approved):

Sources/KlaviyoSwift/KlaviyoModels.swift
ndurell: I think that was swiftformat!

ajaysubra (approved):

PR #118, opened 2023-11-16
Improve handling of push token / profile updates

ndurell:
# Description

# Check List

# Manual Test Plan

1.

# Supporting Materials


Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -86,9 +86,6 @@ struct KlaviyoReducer: ReducerProtocol {
 
             state = initialState
             state.initalizationState = .initialized
-            if let request = try? state.buildProfileRequest() {
```
ndurell, line 89: We are no longer sending a request after initialization.

Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -323,21 +336,50 @@ struct KlaviyoReducer: ReducerProtocol {
...
+            let pushTokenData = state.pushTokenData
             state.reset()
+            guard case .initialized = state.initalizationState,
+                  let apiKey = state.apiKey,
+                  let _ = state.anonymousId
+            else {
+                return .none
+            }
+            if let tokenData = pushTokenData {
+                guard let anonymousId = state.anonymousId else {
```
ajaysubra, line 374: Is this redundant on the same code on line 369?

ndurell: Yup good point. We can eliminate the first check now I thinkl

Sources/KlaviyoSwift/InternalAPIModels.swift
```diff
@@ -391,7 +391,7 @@ struct LegacyEvent: Equatable {
 
         if eventName == "$opened_push" {
             // Special handling for $opened_push include push token at the time of open
-            eventProperties["push_token"] = state.pushToken
+            eventProperties["push_token"] = state.pushTokenData?.pushBackground
```
evan-masseau, line 394: Whats this do?

ndurell: that looks like a bug 🤦🏻

evan-masseau: seems like this would result in `push_token: "ENABLED"`

ndurell: Oh maybe you are commenting on an old commit?

evan-masseau: I don't think so... it says `eventProperties["push_token"] = state.pushTokenData?.pushBackground`

ndurell: I changed how the state was arranged a bit so it could be diffed against token/device updates.

evan-masseau: should be `eventProperties["push_token"] = state.pushTokenData?.pushToken` ?

ndurell: Nvm...I thought when I clicked on the link it would take me to the branch but I was looking at master. It's wrong I will fix it!

evan-masseau: That coulda been caught by a unit test, maybe we're missing some detail in push opened event coverage

Sources/KlaviyoSwift/KlaviyoEnvironment.swift
```diff
@@ -73,7 +74,11 @@ struct KlaviyoEnvironment {
             reachabilityService?.currentReachabilityStatus
         },
         randomInt: { Int.random(in: 0...10) },
-        stateChangePublisher: StateChangePublisher().publisher)
+        stateChangePublisher: StateChangePublisher().publisher, raiseFatalError: { msg in
+            #if DEBUG
+            fatalError(msg)
+            #endif
```
ajaysubra, line 80: nice!

Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -88,23 +104,68 @@ struct KlaviyoState: Equatable, Codable {
...
+
+    mutating func enqueueProfileOrTokenRequest() {
+        guard let apiKey = apiKey,
+              let anonymousId = anonymousId else {
+            runtimeWarn("SDK internal error")
+            return
+        }
+        // if we have push data and we are switching emails
+        // we want to associate the token with the new email.
+        if let pushTokenData = pushTokenData {
```
ajaysubra, line 150:
```suggestion
        if let pushTokenData = self.pushTokenData {
```

NIT: so that we are consistent with using `self` when accessing this property.

ajaysubra (approved):

Sources/KlaviyoSwift/KlaviyoState.swift
ndurell: This doesn't match our linting rules 😬

evan-masseau: it definitely does read weird.

ndurell: with syntax highlighting it's not as bad.
Screenshot 2023-11-29 at 4 29 43 PM

evan-masseau: yeah i was thinking that

evan-masseau (approved):
Looks good from just reading thru on GH

Issue #119, opened 2023-11-30
Missing "Added to Cart" event


walkingbrad:
### Description

I'm new to Klaviyo, but neither the iOS nor Android SDKs have the "Added to Cart" event name/type available. You guys have plenty of [documentation](https://developers.klaviyo.com/en/docs/guide_to_integrating_a_platform_without_a_pre_built_klaviyo_integration) indicating it's a normal standard event (including JS examples for triggering it), so I'm confused.

Should we trigger it as a custom event? If so, what should the event name be? Other mobile event names look like `$viewed_product` whereas JS documentation shows event names like `Added to Cart`, so I'm also confused about the formatting differences.

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-swift-sdk/issues) or [discussion](https://github.com/pointfreeco/klaviyo/klaviyo-swift-sdk/discussions).

### Expected behavior

There is an explicit `addedToCart` event name.

### Actual behavior

There is none, while most if not all other events exist.

### The Klaviyo Swift SDK version information

2.2.1

### Destination operating system

iOS 17

### Xcode version information

Xcode 15

jordan-griffin:
Hey, thanks for opening this issue, it's a good call-out that we're missing this from our event options in both our mobile SDKs currently. I've filed away a feature request for this change for a future update to the SDK.

In the interim, you can create an "Added to Cart" event through the custom event type in the SDK with the key "$added_to_cart". I believe that should translate to an "Added to Cart" event in Klaviyo. You can then set properties on this event like you would with any other integration: e.g: [Added to Cart tracking snippet for JavaScript](https://developers.klaviyo.com/en/v1-2/docs/guide-to-integrating-a-platform-without-a-pre-built-klaviyo-integration#added-to-cart-tracking-snippet)

walkingbrad:
@jordan-griffin Awesome! Thanks for the quick reply. This is a totally workable solution until the next update comes. Thanks for the deets!

evan-masseau:
Hello, we've rolled out a new release, `2.5` on swift and `1.4` on android to correct event names and remove some that should not have been added in the first place. We appreciate your bringing this inconsistency to our attention! Relevant code is in these PRs:

Swift SDK:
[...]

PR #126, opened 2023-12-11
fix event names

ndurell:
# Description
We used inccorect event names in some cases. This should fix the names to be more in line with other Klaviyo SDKs. These names will also look better in Klaviyo.

# Check List

# Manual Test Plan

1.

# Supporting Materials


Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -180,18 +180,18 @@ extension Event.EventName {
...
-        case .PlacedOrder: return "$placed_order"
-        case .OrderedProduct: return "$ordered_product"
-        case .CancelledOrder: return "$cancelled_order"
-        case .PaidForOrder: return "$paid_for_order"
-        case .SubscribedToBackInStock: return "$subscribed_to_back_in_stock"
-        case .SubscribedToComingSoon: return "$subscribed_to_coming_soon"
-        case .SubscribedToList: return "$subscribed_to_list"
-        case .SuccessfulPayment: return "$successful_payment"
-        case .FailedPayment: return "$failed_payment"
+        case .ViewedProduct: return "Viewed product"
```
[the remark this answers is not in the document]
ajaysubra: what happens to customers who have setup flows using the older events that use $?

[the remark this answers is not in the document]
kennyklaviyo: maybe i missed a convo but what about opened push?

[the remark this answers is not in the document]
ndurell: I think we expected $opened_push to be like this. Our backend won't pickup "Opened Push".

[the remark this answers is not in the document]
ndurell: We might have to inform those customers that the event name is changed and they can just duplicate that flow to use the new event name.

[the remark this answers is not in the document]
ajaysubra: Would this be a breaking SDK update in that case?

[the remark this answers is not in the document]
ndurell: Unfortunately yes.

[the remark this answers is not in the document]
evan-masseau: Thinking this [file](https://github.com/klaviyo/app/blob/c0a5c9c1dbf47593d6cc9c6b78ed89987d2f1ccc/src/learning/app/services/statistics.py) has the casing rules we should follow?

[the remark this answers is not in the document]
evan-masseau: I think yes, there's a canonical list in `app` that we can reference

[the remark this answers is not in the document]
ndurell: Updated these!

evan-masseau (dismissed):

evan-masseau:
Maybe worth adding "missing" types in the same release https://github.com/klaviyo/klaviyo-swift-sdk/issues/119

evan-masseau:
Would this work for namespacing the event names?

```


    private enum EventName: Equatable {}

    public struct V1 {
        public enum EventName: EventName {
            case OpenedPush
            case ViewedProduct
            case CustomEvent(String)
        }
    }

    public struct V2 {
        public enum EventName: EventName {
            case OpenedPush
            case ViewedProduct
            case CustomEvent(String)
        }
    }
```

and

```


extension Event.V1.EventName {
    var value: String {
        switch self {
        case .OpenedPush: return "$opened_push"
        case .ViewedProduct: return "$viewed_product"
        case let .CustomEvent(value): return "\(value)"
        }
    }
}

extension Event.V2.EventName {
    var value: String {
        switch self {
        case .OpenedPush: return "$opened_push"
        case .ViewedProduct: return "Viewed Product"
        case let .CustomEvent(value): return "\(value)"
        }
    }
}
```

evan-masseau:
Never mind, my XCode was broken, that idea doesn't even compile

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -8,29 +8,30 @@
...
-        public let name: EventName
-
-        public init(name: EventName) {
-            self.name = name
+    public struct V1: Equatable {
+        public enum MetricName: MetricNameProtocol {
+            case OpenedPush
+            case ViewedProduct
+            case StartedCheckout
+            case ActiveOnSite
```
evan-masseau, line 32: Synced with Mark, he suggested we go with "Opened App"

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -176,22 +186,27 @@ public struct Profile: Equatable {
...
+}
+
+extension Event.V1.MetricName {
+    public var value: String {
+        switch self {
+        case .OpenedPush: return "$opened_push"
+        case .ViewedProduct: return "Viewed Product"
+        case .StartedCheckout: return "Started Checkout"
+        case .AddedToCart: return "Added to Cart"
+        case .ActiveOnSite: return "Active on Site"
```
evan-masseau, line 209: "Opened App"

MIGRATION_GUIDE.md
```diff
@@ -3,7 +3,23 @@
 
 This guide outlines how developers can migrate from older versions of our SDK to newer ones.
 
-## Migrating from v1.X.0 to v2.0.0
+## Migrating from v2.3.0 to v2.3.1
+It was recently discovered that the Swift SDK was using legacy event for some common event names like "Added to Cart", as well as some event names that were associated with server actions like "Product Ordered". This would result in some events being logged with names like "$added_to_cart" which is undesirable. In order to bring the Swift SDK in line with other product we have introduced new types to ensure the correct event name gets logged by default. The exmaple shows how to use the event names:
```
mark-piana, line 7: I would tweak this slightly. I also noticed a few typos that need to be fixed.

It was recently discovered that the Swift SDK was using legacy event names for some common events, like "Added to Cart," and some events that are associated with server actions, like "Ordered Product." As a result, if your account used these standard events, they were being logged with names like "$added_to_cart," in contrast to website generated events which are logged as "Added to Cart." In order to bring the Swift SDK in line with Klaviyo's other integrations, we have introduced new types to ensure that the correct event name is logged by default. This example shows how to use the event names:

Sources/KlaviyoSwift/KlaviyoModels.swift
mark-piana: Following up on Evan's comment, based on our chat in standup today, I think we should switch "Active on Site" to "Opened App." That follows Klaviyo's standard naming convention of Verb Noun (e.g., Received Push) and feels generic enough to apply to other potential use cases beyond mobile apps in the future.

MIGRATION_GUIDE.md
```diff
@@ -3,7 +3,23 @@
 
 This guide outlines how developers can migrate from older versions of our SDK to newer ones.
 
-## Migrating from v1.X.0 to v2.0.0
+## Migrating from v2.3.0 to v2.3.1
```
evan-masseau, line 6: By semantic versioning, this release should be 3.0.0, as we made breaking changes to the public API.

ndurell: Yah fair maybe I can remove the old api now too.

ndurell: I'll do it in a separate PR.

evan-masseau: Yeah it would be a good opportunity to bundle in other things like that.

MIGRATION_GUIDE.md
```diff
@@ -3,7 +3,23 @@
...
+It was recently discovered that the Swift SDK was using legacy event for some common event names like "Added to Cart", as well as some event names that were associated with server actions like "Product Ordered". This would result in some events being logged with names like "$added_to_cart" which is undesirable. In order to bring the Swift SDK in line with other product we have introduced new types to ensure the correct event name gets logged by default. The exmaple shows how to use the event names:
+```swift
+let event = Event(metricName: Event.V1.MetricName.AddedToCart)
+```
+
+If you would like to continue using the old names you can do the following:
+```swift
+let event = Event(metricName: Event.Legacy.MetricName.AddedToCart)
+```
+But note that these names will deprecated in a future release. You can also achieve the same result with:
```
jordan-griffin, line 16: small typo but: `will be deprecated`

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -8,29 +8,30 @@
...
-        case SuccessfulPayment
-        case FailedPayment
-        case CustomEvent(String)
+    public struct Legacy: Equatable {
+        public enum MetricName: MetricNameProtocol {
+            case OpenedPush
+            case ViewedProduct
+            case StartedCheckout
+            case ActiveOnSite
+            case AddedToCart
```
ajaysubra, line 22: Are these the only non-custom events we now support? Asking since on RN I've added the old list.

ndurell: Yah we updated it we should make RN match this.

ndurell: Uhh also opened app instead of active of site.

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -8,29 +8,30 @@
...
+            case CustomEvent(String)
+        }
     }
 
-    public struct Metric: Equatable {
-        public let name: EventName
-
-        public init(name: EventName) {
-            self.name = name
+    public struct V1: Equatable {
```
ajaysubra, line 27: Sorry late to this PR - do we need the V1 layer?

ndurell: for RN no...

ajaysubra: No I meant in general - do we need this since I feel like it gives the impression that these properties are things that change frequently. This is a nit comment nothing blocking.

ndurell: Oh agreed. We needed a way to get people to update their code. This seemed like the easiest way to make that happen. If we leave it as is someone might end up with duplicate event names. If they want to preserve the old string name they can use v1.

ajaysubra: That makes sense.

ajaysubra (approved):

ajaysubra:
> This PR needs to be split up now, into a non-breaking deprecation for a minor version release, and a breaking change for major release.

I can help with this if you have not already started on it.

evan-masseau:
Superceded by #138 and #139

Issue #122, opened 2023-12-06
Stuck in bad state when passing improper phone number format


walkingbrad:
### Description

If we call `set(profile:)` with a profile containing a phone number that the API doesn't like (e.g. empty string, though presumably other values as well), the SDK seems to get stuck in a bad state that prevents profile syncing from working, even if we later set the phone number to `nil` in code. I notice warning logs that show a 400 http error from the Klaviyo API indicating that the phone number is invalid. This persists even after I then attempt to set the phone number to `nil`.

Besides handling phone number as an empty string, I also worry that this same behavior would occur if the phone number we pass belongs to a country that Klaviyo does not yet have SMS support for. Some conversations I found online from Klaviyo staff indicated that this may be the case.

For what it's worth, I don't believe the Android SDK has the same issue. I pass an empty phone number string there and I think all is well.

My app sets the profile in a reactive manner whenever the "customer stream" emits a new customer value. This occurs immediately on app start, on sign in / register, on profile information update, and on sign out (in which we call `resetProfile()`).

---

I assume `nil` values are interpreted as "ignore", where the internal state management retains previously set non-nil values. So it seems impossible to effectively remove a previously-set field.

It may be worth considering switching from a raw `String` to an enum that allows more granularity. I've seen other tools implement something similar. Example:

```
enum ProfileValue {
  case unset
  case set(String?)
}
```

### Checklist

- [X] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-swift-sdk/issues) or [discussion](https://github.com/pointfreeco/klaviyo/klaviyo-swift-sdk/discussions).

### Expected behavior

Either:
- an empty phone number string is parsed out by the SDK and not sent up to the API
- setting a profile field to `nil` wipes out the previous field value (though I understand this has other implications, so probs not a solution)
- 

### Actual behavior

The Klaviyo SDK holds onto old profile values (the phone number empty string) and does not unset them when we later try to set it to `nil`, resulting in all future API calls to fail.

### Steps to reproduce

1. Set profile phone number to empty string
2. Notice API call fail with an invalid phone number error
3. Set profile phone number to nil
4. Notice API call continue to fail for the same initial reason

### The Klaviyo Swift SDK version information

2.2.1

### Destination operating system

iOS 17

[...]

PR #127, opened 2023-12-12
Fix for SDK stuck in bad state when email or phone number fails API validation

ajaysubra:
# Description

We had a customer open a [issue](https://github.com/klaviyo/klaviyo-swift-sdk/issues/122) where they pointed out that the app is stuck in a bad state whenever the email or phone number fails validation. This code aims to do two things - 
1. If the phone number fails API validation we will erase the state and dequeue this request.
2. If the email is invalid, we do the same, reset the state and dequeue the request.

We try to keep the SDK fairly light weight and dequeue the request vs. trying again. 

# Check List

# Manual Test Plan

Case 1: Invalid phone number 

1. Add an invalid phone number on a test app using the SDK
2. After this request fails, send a event request using a test event 
3. This event succeeded vs. it used to fail in the past

Case 2: Invalid email
Same as above except using a invalid email instead of a phone number.

### DIFF
--- a/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
+++ b/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
@@ -67,7 +67,7 @@ func handleRequestError(
         if let invalidFields, !invalidFields.isEmpty {
             return .resetStateAndDequeue(request, invalidFields)
         } else {
-            return .dequeCompletedResults(request)
+            return .deQueueCompletedResults(request)
         }
 
     case let .networkError(error):
@@ -82,23 +82,23 @@ func handleRequestError(
 
     case let .internalError(data):
         runtimeWarn("An internal error occurred msg: \(data)")
-        return .dequeCompletedResults(request)
+        return .deQueueCompletedResults(request)
 
     case let .internalRequestError(error):
         runtimeWarn("An internal request error occurred msg: \(error)")
-        return .dequeCompletedResults(request)
+        return .deQueueCompletedResults(request)
 
     case let .unknownError(error):
         runtimeWarn("An unknown request error occured \(error)")
-        return .dequeCompletedResults(request)
+        return .deQueueCompletedResults(request)
 
     case .dataEncodingError:
         runtimeWarn("A data encoding error occurred during transmission.")
-        return .dequeCompletedResults(request)
+        return .deQueueCompletedResults(request)
 
     case .invalidData:
         runtimeWarn("Invalid data supplied for request. Skipping.")
-        return .dequeCompletedResults(request)
+        return .deQueueCompletedResults(request)
 
     case .rateLimitError:
         var requestRetryCount = 0
@@ -122,6 +122,6 @@ func handleRequestError(
 
     case .missingOrInvalidResponse:
         runtimeWarn("Missing or invalid response from api.")
-        return .dequeCompletedResults(request)
+        return .deQueueCompletedResults(request)
     }
 }
--- a/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
+++ b/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
@@ -125,21 +125,3 @@ func handleRequestError(
         return .dequeCompletedResults(request)
     }
 }
-
-struct ErrorResponse: Codable {
-    let errors: [ErrorDetail]
-}
-
-struct ErrorDetail: Codable {
-    let id: String
-    let status: Int
-    let code: String
-    let title: String
-    let detail: String
-    let source: ErrorSource
-    let meta: [String: String]
-}
-
-struct ErrorSource: Codable {
-    let pointer: String
-}
--- a/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
+++ b/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
@@ -40,18 +40,14 @@ private func getDelaySeconds(for count: Int) -> Int {
 }
 
 private func parseError(_ data: Data) -> [InvalidField]? {
-    var invalidFields: [InvalidField]?
-    do {
-        let errorResponse = try JSONDecoder().decode(ErrorResponse.self, from: data)
-
-        invalidFields = errorResponse.errors.compactMap { error in
-            InvalidField.getInvalidField(sourcePointer: error.source.pointer)
-        }
-    } catch {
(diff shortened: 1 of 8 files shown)


Sources/KlaviyoSwift/APIRequestErrorHandling.swift
```diff
@@ -26,7 +62,13 @@ func handleRequestError(
...
         let responseString = String(data: data, encoding: .utf8) ?? "[Unknown]"
         environment.logger.error("An http error occured status code: \(statuscode) data: \(responseString)")
-        return .dequeCompletedResults(request)
+
+        let invalidFields = parseError(data)
+        if let invalidFields, !invalidFields.isEmpty {
+            return .resetStateAndDequeue(request, invalidFields)
+        } else {
+            return .dequeCompletedResults(request)
+        }
```
evan-masseau, line 71: Trivial/spelling: I've noticed we used both `dequeue` and `deque` spellings. I'd advocate for picking a consistent spelling, and also camel-casing it like `deQueue` to distinguish from `deque`/`dequeue` which is a type of queue with two ends ([double-ended queue](https://www.geeksforgeeks.org/difference-between-queue-and-deque-queue-vs-deque/)).

evan-masseau:
Makes sense to me. I do wonder if we should do anything more than just drop it out of the queue, like emit a warning or error message. I know last spring we did discuss whether to include any kind of phone or email validation and decided to leave that up to the server at the time.

Sources/KlaviyoSwift/APIRequestErrorHandling.swift
```diff
@@ -12,12 +12,48 @@ struct ErrorHandlingConstants {
...
+}
+
 private func getDelaySeconds(for count: Int) -> Int {
     let delay = Int(pow(2.0, Double(count)))
     let jitter = environment.randomInt()
     return min(delay + jitter, ErrorHandlingConstants.maxBackoff)
 }
 
+private func parseError(_ data: Data) -> [InvalidField]? {
+    var invalidFields: [InvalidField]?
```
ndurell, line 43: Not clear from the diff but is this happening on a background thread?

Sources/KlaviyoSwift/APIRequestErrorHandling.swift
ajaysubra: Good catch! and thanks for the suggestion. Using camelCase and In the interesting of keeping this PR readable I created a separate PR with just the rename changes [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/128). Let me know if it looks good.

ajaysubra:
> Makes sense to me. I do wonder if we should do anything more than just drop it out of the queue, like emit a warning or error message. I know last spring we did discuss whether to include any kind of phone or email validation and decided to leave that up to the server at the time.

Documenting for viz - We discussed this internally and decided to silently fail for now mainly due to not have targeted error codes from the API. We do at some point expect to retry based on more accurate error codes from the API.

Sources/KlaviyoSwift/APIRequestErrorHandling.swift
ajaysubra: It's the same queue that `handleRequestError` is called which seems to have [user-initiated QoS](https://developer.apple.com/documentation/dispatch/dispatchqos/1780759-userinitiated). 
image

Sources/KlaviyoSwift/APIRequestErrorHandling.swift
```diff
@@ -12,12 +12,48 @@ struct ErrorHandlingConstants {
...
+    do {
+        let errorResponse = try JSONDecoder().decode(ErrorResponse.self, from: data)
```
ndurell, line 45: There is a decode method (I think) in the environment which might aid you in testing.

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -196,3 +196,21 @@ extension Event.EventName {
...
+}
+
+struct ErrorDetail: Codable {
+    let id: String
+    let status: Int
+    let code: String
+    let title: String
+    let detail: String
+    let source: ErrorSource
+    let meta: [String: String]
```
ndurell, line 211: Is the meta always this shape? Also if we aren't reading maybe better to leave it out until we need really it.

Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -481,6 +486,18 @@ struct KlaviyoReducer: ReducerProtocol {
             pendingProfile[key] = value
             state.pendingProfile = pendingProfile
             return .none
+
+        case let .resetStateAndDequeue(request, invalidFields):
```
ndurell, line 490: I might have modified the deque action like this instead to include invalidFields. You are effectively chaining two actions together when you could update the state in one go.

Tests/KlaviyoSwiftTests/APIRequestErrorHandlingTests.swift
```diff
@@ -35,6 +35,50 @@ class APIRequestErrorHandlingTests: XCTestCase {
         }
     }
 
+    func testSendRequestHttpFailureForPhoneNumberResetsStateAndDequesRequest() async throws {
```
ndurell, line 38: great tests!

ndurell (approved):

evan-masseau (approved):
LGTM, were you going to merge in the deque/dequeue branch into this one also?

Sources/KlaviyoSwift/APIRequestErrorHandling.swift
ajaysubra: Ah good call. Using the one from environment and it helped simplify the code a bit as well.

Sources/KlaviyoSwift/StateManagement.swift
ajaysubra: Yeah I initially had it that way but decided to create a separate action for this mainly because there are other code paths that call into the dequeue action and I didn't want to add code that is specific to a very small edge case in there.

ajaysubra:
> LGTM, were you going to merge in the deque/dequeue branch into this one also?

Yup, was just going to address Noah comments and then go for the merge. Doing it now.

Sources/KlaviyoSwift/KlaviyoModels.swift
ajaysubra: Good call, removed this since we don't use it.

PR #129, opened 2024-01-02
Fixing a user reported crash on device model code

ajaysubra:
# Description

Added a check to the device model code to avoid a reported crash. Checking the size before copying over the data to the pointer.

# Check List

# Manual Test Plan

Checked on a test app and device model showed up fine on the server for an iPhone 13 mini. 

```json
{
  "SDK Version": "2.3.0",
  "Device ID": "XXXX",
  "Total Price": 10.99,
  "OS Version": "17.1.2",
  "OS Name": "iOS",
  "App ID": "com.klaviyo.testiosforbfcm",
  "Device Manufacturer": "Apple",
  "App Name": "testingKlaviyo",
  "Device Model": "iPhone14,4",
  "App Version": "1.0",
  "Items Purchased": [
    "Hot Dog",
    "Fries",
    "Shake"
  ],
  "SDK Name": "swift",
  "Push Token": "XXX",
  "App Build": "1",
  "$value": 10.99
}
```

Reproduced the crash by commenting the line where the size is assigned - 
image

### DIFF
--- a/Sources/KlaviyoSwift/AppContextInfo.swift
+++ b/Sources/KlaviyoSwift/AppContextInfo.swift
@@ -20,10 +20,14 @@ struct AppContextInfo {
     private static let defaultOSName = "iOS"
     private static let defaultDeviceModel: String = {
         var size = 0
+        var deviceModel = ""
         sysctlbyname("hw.machine", nil, &size, nil, 0)
         var machine = [CChar](repeating: 0, count: size)
-        sysctlbyname("hw.machine", &machine, &size, nil, 0)
-        return String(cString: machine)
+        if size > 0 {
+            sysctlbyname("hw.machine", &machine, &size, nil, 0)
+            deviceModel = String(cString: machine)
+        }
+        return deviceModel
     }()
 
     private static let deviceIdStoreKey = "_klaviyo_device_id"
--- a/Sources/KlaviyoSwift/AppContextInfo.swift
+++ b/Sources/KlaviyoSwift/AppContextInfo.swift
@@ -22,8 +22,8 @@ struct AppContextInfo {
         var size = 0
         var deviceModel = ""
         sysctlbyname("hw.machine", nil, &size, nil, 0)
-        var machine = [CChar](repeating: 0, count: size)
         if size > 0 {
+            var machine = [CChar](repeating: 0, count: size)
             sysctlbyname("hw.machine", &machine, &size, nil, 0)
             deviceModel = String(cString: machine)
         }


evan-masseau:
did the iphone 13 mini repro the crash?

evan-masseau (commented):
It looks right. Obvs it'd be great if we could repro the crash too, but IDK if that's possible.

Sources/KlaviyoSwift/AppContextInfo.swift
```diff
@@ -20,10 +20,14 @@ struct AppContextInfo {
     private static let defaultOSName = "iOS"
     private static let defaultDeviceModel: String = {
         var size = 0
+        var deviceModel = ""
         sysctlbyname("hw.machine", nil, &size, nil, 0)
         var machine = [CChar](repeating: 0, count: size)
```
evan-masseau, line 25: nit: move `var machine` declaration inside the if

Sources/KlaviyoSwift/AppContextInfo.swift
```diff
@@ -20,10 +20,14 @@ struct AppContextInfo {
     private static let defaultOSName = "iOS"
     private static let defaultDeviceModel: String = {
         var size = 0
+        var deviceModel = ""
```
kennyklaviyo, line 23: does our backend support empty string values for device model?

ajaysubra:
> did the iphone 13 mini repro the crash?

Yes.

image

Sources/KlaviyoSwift/AppContextInfo.swift
ajaysubra: Good question. I tried sending a empty string and it seems to accept w/o any issue. May be instead of empty string we can have it as unknown?
```json
{
  "SDK Version": "2.3.0",
  "App Version": "1.0",
  "SDK Name": "swift",
  "OS Version": "17.1.1",
  "Total Price": 10.99,
  "Push Token": null,
  "App ID": "com.klaviyo.testiosforbfcm",
  "Items Purchased": [
    "Hot Dog",
    "Fries",
    "Shake"
  ],
  "App Build": "1",
  "App Name": "testingKlaviyo",
  "Device Manufacturer": "Apple",
  "Device ID": "XXXXX",
  "Device Model": "",
  "OS Name": "iOS",
  "$value": 10.99
}
```

evan-masseau: Probably best if it is empty string. If the model or manufacturer are empty, the device metadata service will just skip those, with no effect on other metadata fields. If you sent "Unknown" then we'd have DeviceModel records specifically named "Unknown"

ndurell (approved):

evan-masseau (approved):

PR #134, opened 2024-01-04
remove initializer for setting identifiers in events

ndurell:
# Description

We no longer want to allow setting identifiers, time, and profile in events. This can create confusing bugs for developers and brings it in line with the Android SDK.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoModels.swift
+++ b/Sources/KlaviyoSwift/KlaviyoModels.swift
@@ -63,6 +63,7 @@ public struct Event: Equatable {
     public let uniqueId: String
     public let identifiers: Identifiers?
 
+    @available(*, deprecated, renamed: "init(name:properties:value:uniqueId:)", message: "This initializer has been deprecated. Setting profile properties should be set prior to logging events.")
     public init(name: EventName,
                 properties: [String: Any]? = nil,
                 identifiers: Identifiers? = nil,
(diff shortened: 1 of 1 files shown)


ndurell:
This will be released as a minor point release first and then removed in upcoming 3.0.0.

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -63,6 +63,7 @@ public struct Event: Equatable {
     public let uniqueId: String
     public let identifiers: Identifiers?
 
+    @available(*, deprecated, renamed: "init(name:properties:value:uniqueId:)", message: "This initializer has been deprecated. Setting profile properties should be set prior to logging events.")
```
ajaysubra, line 66:
```suggestion
    @available(*, deprecated, renamed: "init(name:properties:value:uniqueId:)", message: "This initializer has been deprecated. Profile properties should be set prior to logging events.")
```

ajaysubra (approved):

Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -87,6 +88,19 @@ public struct Event: Equatable {
...
+    public init(name: EventName,
+                properties: [String: Any]? = nil,
+                value: Double? = nil,
+                uniqueId: String? = nil) {
+        metric = .init(name: name)
+        _profile = AnyCodable([:])
+        _properties = AnyCodable(properties ?? [:])
+        identifiers = nil
+        self.value = value
+        time = environment.analytics.date()
```
kennyklaviyo, line 101: shouldn't this be `self.identifiers` and `self.time`?

ndurell: Not sure I'm totally following but you don't have to use self if it's clear what you are setting. I think I have a swiftlint rule that will remove this but I can try setting it.

evan-masseau (approved):

PR #136, opened 2024-01-09
Fix broken invalid url handling on iOS 17

ajaysubra:
# Description

In iOS 17 apple changes how url's string initializer works and this was causing one of our tests that validates invalid urls to fail. I reverted the URL initializer to not use the new decoding and sticking with the pre iOS 17 logic of keeping invalid characters from decoding. This is fine since the string we are using the URL initializer is a static one and the idea is to align the code with the tests we had. 

[Change](https://ihor.pro/handling-a-new-url-parsing-behavior-in-ios-17-c08598d28659) in iOS 17 for more context.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/.github/workflows/swift.yml
+++ b/.github/workflows/swift.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: macos-13
     strategy:
       matrix:
-        xcode: ['14.3']
+        xcode: ['15.1']
         config: ['debug', 'release']
     steps:
       - uses: actions/checkout@v3
--- a/Sources/KlaviyoSwift/KlaviyoAPI.swift
+++ b/Sources/KlaviyoSwift/KlaviyoAPI.swift
@@ -103,7 +103,11 @@ extension KlaviyoAPI.KlaviyoRequest {
     var url: URL? {
         switch endpoint {
         case .createProfile, .createEvent, .registerPushToken, .unregisterPushToken:
-            return URL(string: "\(environment.analytics.apiURL)/\(path)/?company_id=\(apiKey)")
+            if #available(iOS 17.0, *) {
+                return URL(string: "\(environment.analytics.apiURL)/\(path)/?company_id=\(apiKey)", encodingInvalidCharacters: false)
+            } else {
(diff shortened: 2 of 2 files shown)


Sources/KlaviyoSwift/KlaviyoAPI.swift
```diff
@@ -103,7 +103,11 @@ extension KlaviyoAPI.KlaviyoRequest {
     var url: URL? {
         switch endpoint {
         case .createProfile, .createEvent, .registerPushToken, .unregisterPushToken:
-            return URL(string: "\(environment.analytics.apiURL)/\(path)/?company_id=\(apiKey)")
+            if #available(iOS 17.0, *) {
+                return URL(string: "\(environment.analytics.apiURL)/\(path)/?company_id=\(apiKey)", encodingInvalidCharacters: false)
```
ndurell, line 107: Nice

ndurell (approved):

.github/workflows/swift.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: macos-13
     strategy:
       matrix:
-        xcode: ['14.3']
+        xcode: ['15.1']
```
kennyklaviyo, line 22: is this change related?

evan-masseau: Yeah, since the breaking code is related to a new swift version, he had to update the runner XCode version to access that new swift version.

evan-masseau (approved):

evan-masseau:
@ajaysubra assume you won't mind, i updated this against master so we can see if CI 100% is fixed now.

ajaysubra:
> hm CI is hung on 14.3 tests, which I think you removed. Might have to force-merge this one, maybe those tests are cached or something?

Thanks for updating with master. You are probably right that GH is caching something. Manually triggered the jobs 🤞🏽

kennyklaviyo (approved):

ajaysubra:
Actions is passing but it's not updating the PR so I'm going to just merge this bypassing the checks.
image

PR #140, opened 2024-01-11
Dropped support for iOS 13

ajaysubra:
# Description

This is going out in the next major version and we'll be dropping support for iOS 13 and min deployment version support will be iOS 14.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/KlaviyoSwift.podspec
+++ b/KlaviyoSwift.podspec
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.source           = { :git => "https://github.com/klaviyo/klaviyo-swift-sdk.git", :tag => s.version.to_s }
   s.swift_version = '5.7'
   s.platform = :ios
-  s.ios.deployment_target = '13.0'
+  s.ios.deployment_target = '14.0'
   s.source_files = 'Sources/KlaviyoSwift/**/*.swift'
   s.dependency     'AnyCodable-FlightSchool'
 end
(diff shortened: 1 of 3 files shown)


evan-masseau (approved):

evan-masseau:
If this is standard practice then we may as well do it in a major release. 
I do have slight reservation that we aren't really doing it for a _positive reason_ like to attain new features, which makes it hard to justify to a developer who maybe has an app that still supports older OSes. But perhaps that is a very small/non-existent segment.

kennyklaviyo (approved):

KlaviyoSwift.podspec
```diff
@@ -13,7 +13,7 @@ Pod::Spec.new do |s|
   s.source           = { :git => "https://github.com/klaviyo/klaviyo-swift-sdk.git", :tag => s.version.to_s }
   s.swift_version = '5.7'
   s.platform = :ios
-  s.ios.deployment_target = '13.0'
+  s.ios.deployment_target = '14.0'
```
kennyklaviyo, line 16: > If this is standard practice then we may as well do it in a major release.
> I do have slight reservation that we aren't really doing it for a positive reason like to attain new features, which makes it hard to justify to a developer who maybe has an app that still supports older OSes. But perhaps that is a very small/non-existent segment.

If this is true, we'd lose 0.7% of devices
https://iosref.com/ios-usage

kennyklaviyo: @evan-masseau

ajaysubra: Had a discussion internally and decided to leave this at 13 for now.

ajaysubra:
Had a discussion internally and decided to leave this at 13 for now.

PR #143, opened 2024-01-12
Getting ready for release 3.0.0 | cocoapods 

ajaysubra:
# Description

Updating cocoapods version to 3.0..0

# Check List

# Manual Test Plan

1.

# Supporting Materials


MIGRATION_GUIDE.md
```diff
@@ -3,6 +3,35 @@
 
 This guide outlines how developers can migrate from older versions of our SDK to newer ones.
 
+## Migrating to v3.0.0
```
jordan-griffin, line 6: Is this section needed if it just references migrating to v2.4.0 anyway?

jordan-griffin: Oh actually I thought this was in the setup readme, I see you actually put it in a standalone migration guide. I think thats more sensible!

MIGRATION_GUIDE.md
```diff
@@ -3,6 +3,35 @@
...
+
+```swift
+// Old code: using one of the legacy enum cases
+let event = Event(name: .ViewedProduct)
+
+// New code: update to new case with -Metric suffix
+let event = Event(name: .ViewedProductMetric)
+```
+
+If you would are using any of the old names and need to continue using them, you can use the custom enum e.g.
```
jordan-griffin, line 30: `If you are using any of the old event names and need to continue using them, you can use the custom enum e.g.`

Just removing the `would` from the sentence. I think it was left in there by accident

jordan-griffin (approved):

MIGRATION_GUIDE.md
ajaysubra: Evan added the migration guide :) 
This PR was suppose to just be just updating SDK version in code to 3.0.0 and because master was ahead of the release branch there was some file that got included here like the one you commented on.

MIGRATION_GUIDE.md
ajaysubra: Done

PR #144, opened 2024-01-12
remove deprecated event initializer

ndurell:
# Description

Removes access to the event initializer (still used internally).

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoModels.swift
+++ b/Sources/KlaviyoSwift/KlaviyoModels.swift
@@ -45,11 +45,6 @@ public struct Event: Equatable {
     }
 
     private let _properties: AnyCodable
-    var profile: [String: Any] {
-        _profile.value as? [String: Any] ?? [:]
-    }
-
-    let _profile: AnyCodable
     public var time: Date
     public let value: Double?
     public let uniqueId: String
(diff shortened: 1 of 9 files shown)


Sources/KlaviyoSwift/KlaviyoModels.swift
```diff
@@ -26,7 +26,7 @@ public struct Event: Equatable {
         }
     }
 
-    public struct Identifiers: Equatable {
+    struct Identifiers: Equatable {
```
ndurell, line 29: These should have been deprecated...

Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -508,23 +508,10 @@ extension KlaviyoState {
 
 extension Event {
     func updateStateAndEvent(state: inout KlaviyoState) -> Event {
-        let email = identifiers?.email ?? state.email
-        let phoneNumber = identifiers?.phoneNumber ?? state.phoneNumber
-        let externalId = identifiers?.externalId ?? state.externalId
         let identifiers = Identifiers(
```
ndurell, line 511: This no longer updates the state.

Sources/KlaviyoSwift/KlaviyoModels.swift
evan-masseau: I suppose you could still release a 2.6 where you deprecate it. Kinda pedantic.

evan-masseau (approved):

Sources/KlaviyoSwift/KlaviyoModels.swift
ndurell: That was my thought too. I don't think you would use one without the other.

PR #145, opened 2024-01-19
Fixed opened push not getting called when app is terminated

ajaysubra:
# Description

We were handling pending requests but when an app is terminated and launched again from a push notification open it triggers  the open event and this open event is added to the pending requests since the SDK is not yet initialized (when the app is relaunched after terminating it's initialized again). However, between the SDK going from initializing to initialized the requests that were added to pending requests were ignored which is what was causing the opened push events from not getting enqueued and eventually sent out. 

The fix was to just use the pending requests from state and enqueue each of those instead of using the state that was taken from the disk.  

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement.swift
@@ -144,11 +144,8 @@ struct KlaviyoReducer: ReducerProtocol {
             }
             state.initalizationState = .initializing
             state.apiKey = apiKey
-            // carry over pending events
-            let pendingRequests = state.pendingRequests
             return .run { send in
-                var initialState = loadKlaviyoStateFromDisk(apiKey: apiKey)
-                initialState.pendingRequests = pendingRequests
+                let initialState = loadKlaviyoStateFromDisk(apiKey: apiKey)
                 await send(.completeInitialization(initialState))
             }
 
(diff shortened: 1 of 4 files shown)


Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -144,11 +144,8 @@ struct KlaviyoReducer: ReducerProtocol {
             }
             state.initalizationState = .initializing
             state.apiKey = apiKey
-            // carry over pending events
```
ajaysubra, line 147: We are now doing this before completing initialization since we know that in certain cases requests may be added `state.pendingRequests` after this action has completed.

Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -165,12 +162,14 @@ struct KlaviyoReducer: ReducerProtocol {
             if let externalId = state.externalId {
                 initialState.externalId = externalId
             }
-            let queuedRequests = state.queue
-            initialState.queue += queuedRequests
+            // For any requests that get added between initilizing and initilized.
+            // Ex: when the app is invoked from a push notification after being killed from the app switcher.
+            let pendingRequests = state.pendingRequests
```
ajaysubra, line 167: This is the line that actually fixes the issue. When we don't have this, the requests that are added to `state.pendingRequests` after the initialization action were erased from state without enqueuing them.

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -441,4 +441,31 @@ class StateManagementTests: XCTestCase {
             try $0.enqueueRequest(request: .init(apiKey: XCTUnwrap($0.apiKey), endpoint: .createEvent(.init(data: .init(event: event, anonymousId: XCTUnwrap($0.anonymousId))))))
         }
     }
+
+    func testEnqueueEventWhenInitilizingSendsEvent() async throws {
```
ajaysubra, line 445: Test to validate the solution works.

ndurell: Awesome test!

ndurell (approved):

evan-masseau (approved):
I have a sort of second hand understanding of the root cause, and this fix makes sense to me based on that understanding!

Issue #151, opened 2024-01-31
Unable to build SDK


harishvitta:
### Description

I am having an issue showing inside the framework. Having build issues
<img width="1274" alt="Screenshot 2024-01-31 at 1 34 27 PM" src="https://github.com/klaviyo/klaviyo-swift-sdk/assets/786101/0b94be75-f279-4a81-b8da-f97bccc7479a">

### Checklist

- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-swift-sdk/issues) or [discussion](https://github.com/pointfreeco/klaviyo/klaviyo-swift-sdk/discussions).

harishvitta:
@ajaysubra Can you please have a look into this issue?

ndurell:
Can you also provide version details on xcode?

harishvitta:
Xcode: 14.3.1

ndurell:
Thanks and sorry for this will try to repro here and get a fix to you shortly.

harishvitta:
Thank you. I have downloaded the latest version of Xcode, which resolved my issue. But people who are on older versions will face this issue.

ajaysubra:
@harishvitta Thanks again for reporting this to us. We have since fixed this issue and will be releasing it next week. Closing this issue for now.

ajaysubra:
@harishvitta As mentioned in the previous message we've made the fix and published two releases - `2.5.2` and `3.0.2` containing the fix. These releases should work on xcode 14.x. Thanks again for reporting the issue to us.

harishvitta:
@ajaysubra Thank you for quick fix.

PR #152, opened 2024-01-31
Fix - SDK not building on xcode 14 and lower

ajaysubra:
# Description

An issue was reported to us about build issues on xcode 14.3.1 [here](https://github.com/klaviyo/klaviyo-swift-sdk/issues/151) and this would be the case on older versions of xcode as well.

This was related to a change I had made to [fix](https://github.com/klaviyo/klaviyo-swift-sdk/pull/136) one of the broken tests in newer versions of xcode because of a change in the way URL initialization in iOS 17 onwards.

Although the fix solve the test from failing it added a run time check for iOS 17 but since xcode 14.x comes with SDKs for iOS 16.x, this new API that was introduced in 17 would not be available and hence cause a compile time failure. 

I found [this](https://forums.swift.org/t/do-we-need-something-like-if-available/40349/32) swift forum proposal to add a compile time check of SDKs but there was no resolution and most suggestions were not a clean fix.

I spent some time looking at the test itself and the idea of the test is to test for invalid urls and with the change in the way urls are encoded (URL automatically percent- and IDNA-encodes invalid characters to help create a valid URL [source](https://developer.apple.com/documentation/foundation/url/3126806-init)) the string we had would have been a valid URL and hence decided to actually used an invalid URL I.E empty string but this alone would not fix the issue since the URL failable initializer appends other strings to `environment.analytics.apiURL` so added a check to be sure that we have a non empty api URL in order to form a URL instance. 

# Check List

# Manual Test Plan

1.

# Supporting Materials


Sources/KlaviyoSwift/KlaviyoAPI.swift
```diff
@@ -103,11 +103,10 @@ extension KlaviyoAPI.KlaviyoRequest {
     var url: URL? {
         switch endpoint {
         case .createProfile, .createEvent, .registerPushToken, .unregisterPushToken:
-            if #available(iOS 17.0, *) {
-                return URL(string: "\(environment.analytics.apiURL)/\(path)/?company_id=\(apiKey)", encodingInvalidCharacters: false)
-            } else {
+            if !environment.analytics.apiURL.isEmpty {
```
ndurell, line 106: Not exactly the same thing as before but it'll work! The compile time check didn't work out huh?

ndurell (approved):

Sources/KlaviyoSwift/KlaviyoAPI.swift
ajaysubra: Yeah, I added why I went this way in the description.

kennyklaviyo (approved):

PR #156, opened 2024-02-08
Merged requesting and collecting tokens under the same section

ajaysubra:
# Description

⚠️  README ⚠️ 

Realized that the SDK doesn't refresh the push enablement status automatically and hence my prior instructions were incorrect. Update them to now set the push token only after requesting push permissions. 

This won't allow for silent push to work in some cases as our SDK currently only allows setting token after requesting push permission. I can think of two options here - 

1. Having a seperate method for setting the push enablement status so that we can get the token on app launch and then allow the user to request push permission at any point in their app as they see fit and update our SDK when they have the permissions (or not).
2. The SDK listens for app foregrounded life cycle events and gets the push enablement status and updates the register token endpoint whenever the enablement status has changed. I think I prefer this over the first approach.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,8 +18,7 @@
 - [Push Notifications](#push-notifications)
   - [Prerequisites](#prerequisites)
   - [Setup](#setup)
-  - [Collecting Push Tokens](#collecting-push-tokens)
-  - [Request Push Notification Permission](#request-push-notification-permission)
+  - [Requesting Push Notification Permission & Collecting tokens](#requesting-push-notification-permission--collecting-tokens)
   - [Receiving Push Notifications](#receiving-push-notifications)
     - [Tracking Open Events](#tracking-open-events)
     - [Deep Linking](#deep-linking)
@@ -52,16 +51,16 @@ KlaviyoSwift is available via [Swift Package Manager](https://swift.org/package-
 2. Select the **Package Dependencies** tab and click on the **add** button below the packages list.
 3. Enter the URL of the Swift SDK repository `https://github.com/klaviyo/klaviyo-swift-sdk` in the text field. This should bring up the package on the screen.
 4. For the dependency rule dropdown select - **Up to Next Major Version** and leave the pre-filled versions as is.
-5. Click **Add Package**.
+5. Click **Add Package** on the next two prompts.
 </details>
 
 <details>
 
 <summary>CocoaPods</summary>
 
-KlaviyoSwift is available through [CocoaPods](https://cocoapods.org/?q=klaviyo).
+KlaviyoSwift is available through [CocoaPods](https://cocoapods.org/pods/KlaviyoSwift).
 
-1. To install, add the following line to your Podfile:
+1. To install, add the following line to your Podfile under your app target:
 
 ```ruby
 pod "KlaviyoSwift"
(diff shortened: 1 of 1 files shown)


jordan-griffin (approved):

evan-masseau:
I don't understand why you have to merge the readme sections, and I think that is confusing because the two don't have to be done together. Can't we just update the code sample about requesting permission to include calling `registerForRemoteNotifications` from the callback, so that they will update token state after permission is set?

README.md
```diff
@@ -255,13 +236,20 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 	        print("error = ", error)
 	    }
 
-	    // Enable or disable features based on the authorization status.
-	    // if you didn't register for remote notifications above you can call `registerForRemoteNotifications` here
-	    // Klaviyo SDK will automatically update the authorization status on next app launch
+        // you can check if the permission was granted or not but it's not necessary
```
evan-masseau, line 239: Nit:
- Fix indentations
- Comment about silent push seems inappropriate for a readme

ajaysubra:
> think that is confusing because the two don't have to be done together.

> think that is confusing because the two don't have to be done together.
Agree in theory but as it currently stands with our SDK they are tightly coupled since you can only set token after requesting permission else the token will never be refreshed.

> Can't we just update the code sample about requesting permission to include calling registerForRemoteNotifications from the callback, so that they will update token state after permission is set?

`registerForRemoteNotifications` is now indeed called from the closure after requesting permissions. About keeping the two sections separate - I guess I could but when I think about how devs would implement this, and how closely they are tied it made sense to keep them together. 

Having said all that, if we act on my suggestions above I do think we can separate them out and I think we should since if we end up at some future point supporting silent push the tight coupling doesn't work since we'd need to token as soon as someone initializes our app and also allows users to request permission when they see fit.

README.md
ajaysubra: I realized the indentation and fix it but you are right about silent push. Removed it.

README.md
```diff
@@ -255,13 +236,19 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
...
 	    }
 
-	    // Enable or disable features based on the authorization status.
-	    // if you didn't register for remote notifications above you can call `registerForRemoteNotifications` here
-	    // Klaviyo SDK will automatically update the authorization status on next app launch
-	}
+        // you can check if the permission was granted or not but it's not necessary especially if you are using provisional push notifications
+        DispatchQueue.main.async {
+            // this will trigger the didRegisterForRemoteNotificationsWithDeviceToken method to be called where you can collect the push token
+            UIApplication.shared.registerForRemoteNotifications()
```
evan-masseau, line 242: If `registerForRemoteNotifications` is called outside this block, should we advise them to call `setPushToken` directly here? Maybe that doesn't make any difference.

evan-masseau:
I'd prefer to keep the separate sections, rather than merge them now and separate them again later. And I think we'd be better served to write this readme _as if_ we support silent push already, because when we do add silent push, customers would be better positioned to take advantage right away if they've already been collecting tokens prior to the permission request anyway.

README.md
ajaysubra: You can't do that since you don't have the token here. The token is only available in the application delegate method - `didRegisterForRemoteNotificationsWithDeviceToken`

evan-masseau: Oh right 🤦‍♂️

ajaysubra:
> I'd prefer to keep the separate sections, rather than merge them now and separate them again later. And I think we'd be better served to write this readme _as if_ we support silent push already, because when we do add silent push, customers would be better positioned to take advantage right away if they've already been collecting tokens prior to the permission request anyway.

💯 agree and kinda the sentiment when I wrote the description of the PR above. Just that we'd have to implement one of the suggestions I made for us to be able to do that. The reason to merge them now is because it easier for devs to implement it in one shot vs. jumping between sections since they are closely tied. `registerForRemoteNotifications` has to be called after permission is requested and that triggers `didRegisterForRemoteNotificationsWithDeviceToken` delegate method to be called. I'm flexible if you want us to seperate it but I do think as things are now with our SDK, it's easier for devs to do it in one go.

evan-masseau:
If it helps I can make a separate PR with what I'm suggesting to see if it makes sense, or if I'm still missing something.

ajaysubra:
Closing in favor of #157

PR #157, opened 2024-02-09
[CHNL-5340] Minor update to push permission request

ajaysubra:
# Description

Can close #156 in favor of this. On speaking with @evan-masseau realized that we we don't need to combine collecting tokens with request permission as when the app is launched we instruct users to register for push token and that should trigger calling setToken. setToken method will update the push enablement status and we should be good. 

Finally, I will cut a ticket for us so that we can refresh the enablement status on the SDK just as a safety measure.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,9 +255,9 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 	        print("error = ", error)
 	    }
 
-	    // Enable or disable features based on the authorization status.
-	    // if you didn't register for remote notifications above you can call `registerForRemoteNotifications` here
-	    // Klaviyo SDK will automatically update the authorization status on next app launch
+	    // Irrespective of the authorization status call `registerForRemoteNotifications` here so that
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -255,9 +255,9 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 	        print("error = ", error)
 	    }
 
-	    // Enable or disable features based on the authorization status.
-	    // if you didn't register for remote notifications above you can call `registerForRemoteNotifications` here
-	    // Klaviyo SDK will automatically update the authorization status on next app launch
+	    // Irrespective of the authorization status call `registerForRemoteNotifications` here so that
+	    // the `didRegisterForRemoteNotificationsWithDeviceToken` delegate is called. Doing this
+	    // will make sure that Klaviyo always has the latest push authorization status.
```
evan-masseau, line 260: Include the code to call it?

evan-masseau (approved):

ndurell (approved):

PR #158, opened 2024-02-09
[CHNL-979] Unify SPM and pods instructions

ajaysubra:
# Description

Unifying the instructions for SPM and pods which were previously scattered between rich push and the initial installation step.

# Check List

# Manual Test Plan

1.

# Supporting Materials


mark-piana (approved):
Looks good to me!

PR #163, opened 2024-03-12
CHNL-6305:  Fix set profile properties method

ajaysubra:
# Description

* Fixed set profile attributes not being set when a token request is made. Token request is made whenever set token is called.
* Edge case: Token request was being made repeatedly since the pending profile was never set to nil. Fixed this as well. 

# Check List

# Manual Test Plan

1. Set profile with most properties and then set profile properties and profile properties override the earlier stuff set with the profile.
2. Setting profile properties first and then set profile should override the previously set profile properties. 
3. Tested w/o setting token and making sure profile request was successful.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoState.swift
+++ b/Sources/KlaviyoSwift/KlaviyoState.swift
@@ -84,6 +84,7 @@ struct KlaviyoState: Equatable, Codable {
         }
     }
 
+    // state related stuff
     var apiKey: String?
     var email: String?
     var anonymousId: String?
@@ -190,42 +191,53 @@ struct KlaviyoState: Equatable, Codable {
         var attributes = profile.data.attributes
         var location = profile.data.attributes.location ?? .init()
         var properties = profile.data.attributes.properties.value as? [String: Any] ?? [:]
+        let updatedProfile = Profile.updateProfileWithProperties(dict: pendingProfile)
 
-        // Optionally (if not already specified) overwrite attributes and location with pending profile information.
-        for (key, value) in pendingProfile {
-            switch key {
-            case .firstName:
-                attributes.firstName = attributes.firstName ?? value.value as? String
-            case .lastName:
-                attributes.lastName = attributes.lastName ?? value.value as? String
-            case .address1:
-                location.address1 = location.address1 ?? value.value as? String
-            case .address2:
-                location.address2 = location.address2 ?? value.value as? String
-            case .title:
-                attributes.title = attributes.title ?? value.value as? String
-            case .organization:
-                attributes.organization = attributes.organization ?? value.value as? String
-            case .city:
-                location.city = location.city ?? value.value as? String
-            case .region:
-                location.region = location.region ?? value.value as? String
-            case .country:
-                location.country = location.country ?? value.value as? String
-            case .zip:
-                location.zip = location.zip ?? value.value as? String
-            case .image:
-                attributes.image = attributes.image ?? value.value as? String
-            case .latitude:
-                location.latitude = location.latitude ?? value.value as? Double
-            case .longitude:
-                location.longitude = location.longitude ?? value.value as? Double
-            case let .custom(customKey: customKey):
-                properties[customKey] = properties[customKey] ?? value.value
-            }
+        if let firstName = updatedProfile.firstName {
+            attributes.firstName = attributes.firstName ?? firstName
         }
+        if let lastName = updatedProfile.lastName {
+            attributes.lastName = attributes.lastName ?? lastName
+        }
+        if let title = updatedProfile.title {
+            attributes.title = attributes.title ?? title
+        }
+        if let organization = updatedProfile.organization {
+            attributes.organization = attributes.organization ?? organization
+        }
+        if !updatedProfile.properties.isEmpty {
+            attributes.properties = AnyCodable(properties.merging(updatedProfile.properties, uniquingKeysWith: { _, new in new }))
+        }
+
+        if let address1 = updatedProfile.location?.address1 {
+            location.address1 = location.address1 ?? address1
+        }
+        if let address2 = updatedProfile.location?.address2 {
+            location.address2 = location.address2 ?? address2
+        }
+        if let city = updatedProfile.location?.city {
+            location.city = location.city ?? city
+        }
+        if let region = updatedProfile.location?.region {
+            location.region = location.region ?? region
+        }
+        if let country = updatedProfile.location?.country {
+            location.country = location.country ?? country
+        }
+        if let zip = updatedProfile.location?.zip {
+            location.zip = location.zip ?? zip
+        }
+        if let image = updatedProfile.image {
+            attributes.image = attributes.image ?? image
+        }
+        if let latitude = updatedProfile.location?.latitude {
+            location.latitude = location.latitude ?? latitude
+        }
+        if let longitude = updatedProfile.location?.longitude {
+            location.longitude = location.longitude ?? longitude
+        }
+
         attributes.location = location
-        attributes.properties = AnyCodable(properties)
         self.pendingProfile = nil
 
         return .init(data: .init(attributes: attributes))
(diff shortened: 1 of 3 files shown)


Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -190,42 +191,53 @@ struct KlaviyoState: Equatable, Codable {
         var attributes = profile.data.attributes
         var location = profile.data.attributes.location ?? .init()
         var properties = profile.data.attributes.properties.value as? [String: Any] ?? [:]
+        let updatedProfile = Profile.updateProfileWithProperties(dict: pendingProfile)
 
-        // Optionally (if not already specified) overwrite attributes and location with pending profile information.
```
ajaysubra, line 194: This is the change I don't mind reverting as I don't think it improves the code. I made this change so that the logic to go over a pending profile dictionary and create a profile (that I use in creation of token request) can be reused.

Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -190,42 +191,53 @@ struct KlaviyoState: Equatable, Codable {
...
-            case .image:
-                attributes.image = attributes.image ?? value.value as? String
-            case .latitude:
-                location.latitude = location.latitude ?? value.value as? Double
-            case .longitude:
-                location.longitude = location.longitude ?? value.value as? Double
-            case let .custom(customKey: customKey):
-                properties[customKey] = properties[customKey] ?? value.value
-            }
+        if let firstName = updatedProfile.firstName {
```
ajaysubra, line 196: @ndurell as you identified this logic is so that if any stuff was set in the attributes of a profile they are preserved.

Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -190,42 +191,53 @@ struct KlaviyoState: Equatable, Codable {
...
+        }
+        if let latitude = updatedProfile.location?.latitude {
+            location.latitude = location.latitude ?? latitude
+        }
+        if let longitude = updatedProfile.location?.longitude {
+            location.longitude = location.longitude ?? longitude
+        }
+
         attributes.location = location
-        attributes.properties = AnyCodable(properties)
```
ajaysubra, line 228: Merging the two dictionaries now.

`attributes.properties = AnyCodable(properties.merging(updatedProfile.properties, uniquingKeysWith: { _, new in new }))`

Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -278,6 +290,54 @@ struct KlaviyoState: Equatable, Codable {
 
         return pushTokenData != newPushTokenData
     }
+
+    func buildProfileRequest(apiKey: String, anonymousId: String, properties: [String: Any] = [:]) -> KlaviyoAPI.KlaviyoRequest {
```
ajaysubra, line 294: Moved this over from the state management file over to here just so that all the state methods were in this file. Found it easy to read this way but not super married if you had a valid reason for them to be in the state management file.

Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -278,6 +290,54 @@ struct KlaviyoState: Equatable, Codable {
...
+        if let pendingProfile = pendingProfile {
+            profile = Profile.updateProfileWithProperties(
+                email: email,
+                phoneNumber: phoneNumber,
+                externalId: externalId,
+                dict: pendingProfile)
+            self.pendingProfile = nil
+        } else {
+            profile = Profile(email: email, phoneNumber: phoneNumber, externalId: externalId)
+        }
```
ajaysubra, line 321: This is the main change that fixes the issue. Earlier (before moving this code here from the state management file) we were not using the `pendingProfile` var from the state when making a token request and profile was constructed as it is in the else code block. `pendingProfile` is set whenever a user calls`KlaviyoSDK().set(profileAttribute: , value: )` and whenever we flush we see if it's nil and if not we enqueue a token or a profile request.

Also, since `pendingProfile` was never set to `nil` during a token request earlier, we kept making token request indefinitely whenever a profile property was set using the `setProfileAttribute` method. This is definitely a super edge case since most folks use the `set<Id>` methods or the `setProfile` method. This is now fixed by setting the ` self.pendingProfile = nil` on line 318.

Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -217,7 +217,7 @@ struct KlaviyoReducer: ReducerProtocol {
                 state.pendingRequests.append(.pushToken(pushToken, enablement))
                 return .none
             }
-            guard state.shouldSendTokenUpdate(newToken: pushToken, enablement: enablement) else {
+            if !state.shouldSendTokenUpdate(newToken: pushToken, enablement: enablement) {
```
ajaysubra, line 220: This is a personal preference change. I find it easier to read if else for conditionals over guard but happy to revert since I don't have a good reason to change it.

Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -466,45 +466,6 @@ extension Store where State == KlaviyoState, Action == KlaviyoAction {
         reducer: KlaviyoReducer())
 }
 
-extension KlaviyoState {
```
ajaysubra, line 469: moved to state file.

Sources/KlaviyoSwift/KlaviyoState.swift
evan-masseau: Thanks for this explanation, definitely not familiar enough with the iOS sdk code to have known this.

evan-masseau (changes requested):
Makes sense with your annotations, two questions remain in my mind:
- Can the snapshot tests for push token request reflect profile attributes, so that this important case is covered by tests
- Does the event API suffer the same sort of problem? If I set profile attributes and created an event, would the profile attributes get sent to the backend appropriately?

ajaysubra:
> Makes sense with your annotations, two questions remain in my mind:
> 
> * Can the snapshot tests for push token request reflect profile attributes, so that this important case is covered by tests
> * Does the event API suffer the same sort of problem? If I set profile attributes and created an event, would the profile attributes get sent to the backend appropriately?

1. Yeah, I'm looking into writing unit tests, which is why I left the PR in the draft state. Sorry I should have mentioned that in the PR.
2. That's a good question. Events is handled separately from profile attributes updates. What I mean by that is, as long as there is something in `pendingProfile` when flushing the queue we will make a profile/token request until `pendingProfile` is nil. Lines 247-249. 

image

In order words, you could make an events request either before or after setting profile attributes, the event request will not have the profile attributes. Instead there will be a profile/token request that is made after the profile attributes are set. This is the current setup, there could be a world where this is captured in the events request and we can avoid a network call but I'll keep that out of the scope of this PR.

evan-masseau:
Sorry, I didn't even notice this was a draft! I knew you'd said something about tests somewhere.  
Sounds good about the events question, just wanted to be sure we weren't already merging profile changes into events calls too. Definitely not something I expected you to _add_ now.

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -355,45 +355,71 @@ class StateManagementTests: XCTestCase {
...
-            (.zip, "02067"),
-            (.country, "Mexico"),
-            (.latitude, 23.0),
-            (.longitude, 46.0),
-            (.title, "King"),
-            (.organization, "Klaviyo"),
-            (.firstName, "Jeffrey"),
-            (.lastName, "Lebowski"),
-            (.image, "foto.png"),
+        let profileAttributes: [(Profile.ProfileKey, Any)] = [
```
ajaysubra, line 358: Updated this existing test since it was already doing what we wanted except for two things - 
1. Making sure that `pendingProfile` was cleared after adding it to the token request 
2. Making sure that the token request payload contains the profile properties set using the setProfileAttributes method.

Added both of these into this test.

evan-masseau: Yeah this works for me, thanks!

evan-masseau (approved):

jordan-griffin (approved):
Not the best person to review in this area, but I think this looks solid

PR #165, opened 2024-03-22
add new headers

ndurell:
* mobile header indicating traffic coming from mobile sdk
* retry attempt header indicating number of retries

# Description

# Check List

# Manual Test Plan

1.

# Supporting Materials


Sources/KlaviyoSwift/KlaviyoAPI.swift
```diff
@@ -49,6 +49,7 @@ struct KlaviyoAPI {
             requestFailed(request, error, 0.0)
             return .failure(.internalRequestError(error))
         }
+        urlRequest.allHTTPHeaderFields?["X-Klaviyo-Retry-Attempt"] = "\(attemptNumber)/50"
```
ndurell, line 52: maybe could use a constant here.

evan-masseau: Just to be super sure, this header should be formatted `1/50` -> `50/50` ?

ndurell: yah that was my understanding...could totally have missed that detail though!

evan-masseau: Makes sense to me, just want to be sure we don't accidentally start at 0 (or end at 49)

ndurell: Yah I'm going to test it through and ensure it does the desired thing.

evan-masseau: I wonder if the backend could be able to reply with a header to tell you to increase your max attempts.

ndurell: @jason-myers-klaviyo let us know if there are specific requirements but IIRC you wanted it to send the last request with 50/50 right?

jason-myers-klaviyo: Yes, I think that's right.  If we see a 50/50 and we return a 429, we'll know that the request will not be tried again.

50/50 meaning "we're on attempt number 50, and we will stop after 50 (meaning we won't go again after this one"

And 1/50 meaning "we're on the first attempt of 50 possible attempts"

evan-masseau:
For viz, the related [android PR](https://github.com/klaviyo/klaviyo-android-sdk/pull/149)

Sources/KlaviyoSwift/KlaviyoAPI.swift
evan-masseau: Lets name this header `X-Klaviyo-Attempt-Count` then?

jason-myers-klaviyo: I think we landed on

`X-Klaviyo-Retry-Attempt`

Earlier

evan-masseau: Yeah its a nitpick, but just responding to the above thread, that name is a little ambiguous: is `1/50` the first _retry_ or first _attempt_.

jason-myers-klaviyo: I see, no issues here then

evan-masseau (changes requested):
Rename to `X-Klaviyo-Attempt-Count`

ajaysubra:
> Rename to `X-Klaviyo-Attempt-Count`

Done. Let me know if it looks good.

Sources/KlaviyoSwift/KlaviyoAPI.swift
ajaysubra: Starting from 1 now. Still doing some testing on my end but so far it's looking good.

Also, renamed header to `X-Klaviyo-Attempt-Count`

ajaysubra (approved):

Tests/KlaviyoSwiftTests/__Snapshots__/KlaviyoAPITests/testSuccessfulResponseWithEvent.1.txt
```diff
@@ -9,7 +9,10 @@
   ▿ httpMethod: Optional<String>
     - some: "POST"
   ▿ allHTTPHeaderFields: Optional<Dictionary<String, String>>
-    - some: 0 key/value pairs
+    ▿ some: 1 key/value pair
+      ▿ (2 elements)
+        - key: "X-Klaviyo-Retry-Attempt"
+        - value: "0/50"
```
evan-masseau, line 15: When an actual device sends, the first send will be 1/50, ya?
I had to do this in my tests too, just want to confirm.

ajaysubra: Yeah I tested this using a proxy, it starts at 1. I will update the tests.

evan-masseau (approved):
LGTM once tests are fixed

PR #166, opened 2024-03-23
use retry after instead of calculating delay

ndurell:
# Description

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
+++ b/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
@@ -33,10 +33,9 @@ enum InvalidField: Equatable {
     }
 }
 
-private func getDelaySeconds(for count: Int) -> Int {
-    let delay = Int(pow(2.0, Double(count)))
+private func addJitter(to value: Int) -> Int {
     let jitter = environment.randomInt()
-    return min(delay + jitter, ErrorHandlingConstants.maxBackoff)
+    return value + jitter
 }
 
 private func parseError(_ data: Data) -> [InvalidField]? {
@@ -112,7 +111,9 @@ func handleRequestError(
         case let .retryWithBackoff(requestCount, totalCount, _):
             requestRetryCount = requestCount + 1
             totalRetryCount = totalCount + 1
-            nextBackoff = retryAfter ?? getDelaySeconds(for: totalRetryCount)
+            let exponentialBackOff = Int(pow(2.0, Double(totalRetryCount)))
+
+            nextBackoff = addJitter(to: retryAfter ?? exponentialBackOff)
         }
         return .requestFailed(
             request, .retryWithBackoff(
--- a/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
+++ b/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
@@ -100,7 +100,7 @@ func handleRequestError(
         environment.emitDeveloperWarning("Invalid data supplied for request. Skipping.")
         return .deQueueCompletedResults(request)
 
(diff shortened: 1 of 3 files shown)


Sources/KlaviyoSwift/APIRequestErrorHandling.swift
```diff
@@ -112,7 +112,7 @@ func handleRequestError(
         case let .retryWithBackoff(requestCount, totalCount, _):
             requestRetryCount = requestCount + 1
             totalRetryCount = totalCount + 1
-            nextBackoff = getDelaySeconds(for: totalRetryCount)
+            nextBackoff = retryAfter ?? getDelaySeconds(for: totalRetryCount)
```
evan-masseau, line 115: Should we apply jitter to the `retryAfter` too?

evan-masseau: I suppose if we needed jitter, the server could apply it.

Sources/KlaviyoSwift/KlaviyoAPI.swift
```diff
@@ -70,7 +70,8 @@ struct KlaviyoAPI {
 
         if httpResponse.statusCode == 429 {
             requestRateLimited(request)
-            return .failure(KlaviyoAPIError.rateLimitError)
+            let retryAfter = httpResponse.allHeaderFields["Retry-After"] as? Int
```
evan-masseau, line 73: may not affect ios, but on android this response headers' value was a map of lists of strings, `Map<String, List<String>>`

Sources/KlaviyoSwift/APIRequestErrorHandling.swift
evan-masseau: @jason-myers-klaviyo do you have any opinion on whether the SDKs should apply a jitter on top of the `Retry-After` value? Does the `Retry-After` just represent the time till the steady state rate limit "resets" or is there more advanced logic to it?

jason-myers-klaviyo: Jitter longer than the seconds given would be fine and up to you.  Nice to have I think

Jitter shorter will cause another 429 I think

jason-myers-klaviyo: It is just the time until the steady windows reset IIRC

evan-masseau: Oh ya definitely only _add_ jitter

evan-masseau (changes requested):
Add jitter (1-10s?) to server's `Retry-After` value

ajaysubra:
> Add jitter (1-10s?) to server's `Retry-After` value

Added jitter to `Retry-After`. Will test it out in the morning.

ajaysubra (approved):

Sources/KlaviyoSwift/KlaviyoAPI.swift
evan-masseau: Were you able to test that this way to parse `Retry-After` header does work on live 429 response from production?

evan-masseau (approved):
LGTM, just confirming if its been tested against a live 429 response from production

Sources/KlaviyoSwift/KlaviyoAPI.swift
ajaysubra: I didn't have a chance to test it out last night but tested now and it was indeed not returning the header value. Switched to using `value(forHTTPHeaderField:` which works.

PR #168, opened 2024-03-25
Updated readme with badge count note

ajaysubra:
# Description

Updated the readme to include code sample of decrementing badge count.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
     }
 }
 ```
- When tracking opened push notification, you can also decrement the badge count on the app icon by adding the following code to the `userNotificationCenter:didReceive:withCompletionHandler` method:
+When tracking opened push notification, you can also decrement the badge count on the app icon by adding the following code to the `userNotificationCenter:didReceive:withCompletionHandler` method:
 
 ```swift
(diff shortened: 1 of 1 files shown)


jordan-griffin (approved):
Might want an approval from someone that understands Swift better

ndurell (approved):

README.md
```diff
@@ -310,6 +310,45 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
     }
 }
 ```
+When tracking opened push notification, you can also decrement the badge count on the app icon by adding the following code to the `userNotificationCenter:didReceive:withCompletionHandler` method:
```
mark-piana, line 313: Should we make it slightly more clear when this will decrement? Maybe "When tracking opened push notification (i.e., when a user taps/opens a push notification), you can also..."

ajaysubra: Hmm. I was just thinking that since this is within the section of "Tracking open events" and the first line reads as - "When a user taps on a push notification,..." it would already be clear since it is within that context but let me know if you think otherwise.

mark-piana: Oh that's a good point. I didn't see that we had "when a user taps on a push notification" in the section above. What you have looks good!

PR #169, opened 2024-03-29
Added empty check for email, phone and external id

ajaysubra:
# Description

Have empty identifiers primarily email was causing a lot of 400s on our APIs as well as consuming unnecessary user bandwidth. We were vary of adding any validation in our SDK as it's mainly a pass through but felt this would be a good fix to the problem described above.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoState.swift
+++ b/Sources/KlaviyoSwift/KlaviyoState.swift
@@ -182,9 +182,17 @@ struct KlaviyoState: Equatable, Codable {
     }
 
     mutating func updateStateWithProfile(profile: Profile) {
-        email = profile.email ?? email
-        phoneNumber = profile.phoneNumber ?? phoneNumber
-        externalId = profile.externalId ?? externalId
+        if let profileEmail = profile.email, !profileEmail.isEmpty, profileEmail != email {
+            email = profileEmail
+        }
+
+        if let profilePhoneNumber = profile.phoneNumber, !profilePhoneNumber.isEmpty, profilePhoneNumber != phoneNumber {
+            phoneNumber = profilePhoneNumber
+        }
+
+        if let profileExternalId = profile.externalId, !profileExternalId.isEmpty, profileExternalId != externalId {
+            externalId = profileExternalId
+        }
     }
 
     mutating func updateRequestAndStateWithPendingProfile(profile: CreateProfilePayload) -> CreateProfilePayload {
--- a/Sources/KlaviyoSwift/KlaviyoState.swift
+++ b/Sources/KlaviyoSwift/KlaviyoState.swift
@@ -121,6 +121,7 @@ struct KlaviyoState: Equatable, Codable {
 
     mutating func updateEmail(email: String) {
         if email.isEmpty || email == self.email {
+            logDevWarning(for: "email")
             return
         }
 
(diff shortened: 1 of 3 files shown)


Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -190,7 +193,7 @@ struct KlaviyoState: Equatable, Codable {
         }
         var attributes = profile.data.attributes
         var location = profile.data.attributes.location ?? .init()
-        var properties = profile.data.attributes.properties.value as? [String: Any] ?? [:]
+        let properties = profile.data.attributes.properties.value as? [String: Any] ?? [:]
```
ajaysubra, line 196: This is unrelated. I messed this up in the previous release so fixing it now.

ndurell: nice - I fixed this in a pr too.

ndurell (approved):
great tests

kennyklaviyo (approved):

Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -182,9 +182,17 @@ struct KlaviyoState: Equatable, Codable {
     }
 
     mutating func updateStateWithProfile(profile: Profile) {
-        email = profile.email ?? email
-        phoneNumber = profile.phoneNumber ?? phoneNumber
-        externalId = profile.externalId ?? externalId
+        if let profileEmail = profile.email, !profileEmail.isEmpty, profileEmail != email {
```
kennyklaviyo, line 185: does this allow setting the email to nil?

evan-masseau (commented):
LGTM only nit is to log a warning wherever we're rejecting a value per [slack chat](https://klaviyo.slack.com/archives/C03FDDX9BU1/p1712072128112739)

ajaysubra:
> LGTM only nit is to log a warning wherever we're rejecting a value per [slack chat](https://klaviyo.slack.com/archives/C03FDDX9BU1/p1712072128112739)

Done - 
image

ajaysubra:
> > LGTM only nit is to log a warning wherever we're rejecting a value per [slack chat](https://klaviyo.slack.com/archives/C03FDDX9BU1/p1712072128112739)
> 
> Done - image

I'm testing this. Might not need some of the warnings on the profile setter

Sources/KlaviyoSwift/KlaviyoState.swift
```diff
@@ -487,3 +503,14 @@ extension Profile {
         return profile
     }
 }
+
+extension String {
+    fileprivate func isNotEmptyOrSame(as state: String?, identifier: String) -> Bool {
```
ajaysubra, line 508: Since I was doing this thrice for three of the identifier setters and then again thrice when setting a profile, I instead created an fileprivate extension on string and calling this to keep the comparison logic in just one location. Call site now looks like - `email.isNotEmptyOrSame(as: self.email, identifier: "email")`

ajaysubra:
@evan-masseau @ndurell @kennyklaviyo sorry made another quick refactor since the comparison logic was all over the place. Tested this using the test app and it looks good.

Sources/KlaviyoSwift/KlaviyoState.swift
evan-masseau: yeah this is nice.

evan-masseau (approved):

PR #172, opened 2024-04-10
Fixed retry after logic to fire at retry after intervals

ajaysubra:
# Description

The existing logic in master was still causing retries to happen every 10secs on wifi since the backoff was not being set for the switch case of just `retry`. This is the case that is set in the `flushQueue` action once the back off interval is exhaused and a call is made. 

Additionally, in the first 429, the `retry` is the default value set to `state.retryInfo` and not having the backoff set in `handleRequestError.rateLimitError` was causing the network request to fire every 10sec on wifi.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
+++ b/Sources/KlaviyoSwift/APIRequestErrorHandling.swift
@@ -102,7 +102,9 @@ func handleRequestError(
     case let .rateLimitError(retryAfter):
         var requestRetryCount = 0
         var totalRetryCount = 0
-        var nextBackoff = 0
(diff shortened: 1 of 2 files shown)


ndurell (approved):

evan-masseau (approved):
~Can't exactly tell what was wrong~ I would never have caught this from just reading the code, glad you caught it with manual testing!

ajaysubra:
> ~Can't exactly tell what was wrong~ I would never have caught this from just reading the code, glad you caught it with manual testing!

Thanks to Proxyman and good ol' print statements!

PR #173, opened 2024-04-11
Added retry after to the testing code

ajaysubra:
# Description

Did this so that we can test this using the internal test app

# Check List

# Manual Test Plan

1.

# Supporting Materials


ndurell (approved):

evan-masseau (approved):

evan-masseau (approved):
is the test app able to access the request/response headers?

ajaysubra:
> is the test app able to access the request/response headers?

Unfortunately just the request header (and not the session one which is where we add the `X-Klaviyo-Mobile=1`). As for the response, we don't send back the entire headers just the status code, duration and few other things. For now this is enough to validate retry-after logic in the test app but we can expose the headers if needed in the future.

PR #174, opened 2024-04-11
Adding privacy manifest to KlaviyoSwift and KlaviyoSwiftExtension 

ajaysubra:
# Description

Apple requires SDK to add a privacy manifest listing, 
1. If the SDK does tracking.
2. If it does tracking, what domains does it use for that purpose
3. What required reason APIs does the SDK use 
4. What data points does the SDK collect. 

The privacy manifest we added here is a least common denominator of what the SDK does, if developers are doing anything in addition to this in their specific apps like gathering emails/phone numbers they need to listed in your applications privacy manifest.

Some material from Apple about this -
 
[Privacy manifest files](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files?language=objc)
[Describing data use in privacy manifests](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_data_use_in_privacy_manifests#4250555) 
[WWDC 23 session on this topic](https://developer.apple.com/videos/play/wwdc2023/10060/)

# Check List

# Manual Test Plan

Tested this using our test app and this is how the report looks like - 

image

# Supporting Materials

Developers can do the same when releasing their apps using archive - 

image

### DIFF
--- a/KlaviyoSwift.podspec
+++ b/KlaviyoSwift.podspec
@@ -15,5 +15,6 @@ Pod::Spec.new do |s|
   s.platform = :ios
   s.ios.deployment_target = '13.0'
   s.source_files = 'Sources/KlaviyoSwift/**/*.swift'
+  s.resource_bundles = {"KlaviyoSwift" => ["Sources/KlaviyoSwift/PrivacyInfo.xcprivacy"]}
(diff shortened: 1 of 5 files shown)


evan-masseau (approved):
LGTM

kennyklaviyo (approved):

PR #181, opened 2024-05-30
Allowing opened push event to be transmitted when SDK is not initialized 

ajaysubra:
# Description

## Background

The SDK can be in three states of initialization -
* Initialized - The SDK is initialized and any requests it gets is queued immediately and flushed at certain intervals.
* Initializing - The SDK is initializing (reading from disk etc) and any requests that the SDK gets is held in memory and enqueued once the SDK is fully initialized. 
* un-initialized - The SDK will ignore the request and log a warning message to the developer (which they probably won't see if the app is terminated when testing locally).

The issue with react native is that by the time react native is booted from the native code the opened event request has already hit our SDK and the SDK is in the un-initialized state so the request is simply ignored. This logic has been in place since Dec 2023 on iOS and is part of requirement at that point and also now that developers should always initialize the SDK before interacting with it. When we implemented initialize in react native we didn't test this edge case.

While we still want our devs to initialize the SDK before calling any of it's methods. As long as they stick to interfacing with the SDK on just one of the layers (native or RN) they shouldn't encounter any issues. There is an exception to this with calling the SDK when a user opens push and this has to happen on the native side because of the application delegation pattern on iOS.

We want to keep the SDK simple in terms of logic and not queue up things which we don't know which company it belong to. For instance,  if we allow devs to set profile and then initialize there is a chance that the profile wasn't intended to be associated with the company that was set later. In short, it keeps us from making too many assumptions on the SDK.

Finally, since we are the ones queueing up the opened push event when the devs calls our SDK to handle a push opened action from the user, it seems reasonable make an exception just in this case since we know for react native it is possible that the SDK may not be initialized in time to handle sending opened push event.

## Solution 

When the SDK is un-initialized and receives an opened push event, keep it in state until the SDK is initialized and then queue it up. 

# Check List

# Manual Test Plan

1. Test all SDK interactions w/o initializing it locally: This should not enqueue any request other than opened push event ✅
2. Test all SDK interactions after initializing and make sure this change didn't impact any existing use cases ✅

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement.swift
@@ -92,11 +92,11 @@ enum KlaviyoAction: Equatable {
 
     var requiresInitialization: Bool {
         switch self {
-        case let .enqueueEvent(event):
-            // if event metric is opened push we DON'T require initilization in all other event metric cases we DO.
-            return event.metric.name != .OpenedPush
+        // if event metric is opened push we DON'T require initilization in all other event metric cases we DO.
+        case let .enqueueEvent(event) where event.metric.name == .OpenedPush:
+            return false
 
-        case .setEmail, .setPhoneNumber, .setExternalId, .setPushToken, .enqueueProfile, .setProfileProperty, .resetProfile, .resetStateAndDequeue:
+        case .setEmail, .setPhoneNumber, .setExternalId, .setPushToken, .enqueueProfile, .setProfileProperty, .resetProfile, .resetStateAndDequeue, .enqueueEvent:
             return true
 
         case .initialize, .completeInitialization, .deQueueCompletedResults, .networkConnectivityChanged, .flushQueue, .sendRequest, .stop, .start, .cancelInFlightRequests, .requestFailed:
--- a/Sources/KlaviyoSwift/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement.swift
@@ -92,7 +92,11 @@ enum KlaviyoAction: Equatable {
 
     var requiresInitialization: Bool {
         switch self {
-        case .setEmail, .setPhoneNumber, .setExternalId, .setPushToken, .enqueueEvent, .enqueueProfile, .setProfileProperty, .resetProfile, .resetStateAndDequeue:
+        case let .enqueueEvent(event):
+            // if event metric is opened push we DON'T require initilization in all other event metric cases we DO.
+            return event.metric.name == .OpenedPush ? false : true
+
+        case .setEmail, .setPhoneNumber, .setExternalId, .setPushToken, .enqueueProfile, .setProfileProperty, .resetProfile, .resetStateAndDequeue:
             return true
 
         case .initialize, .completeInitialization, .deQueueCompletedResults, .networkConnectivityChanged, .flushQueue, .sendRequest, .stop, .start, .cancelInFlightRequests, .requestFailed:
@@ -112,6 +116,7 @@ struct KlaviyoReducer: ReducerProtocol {
         if action.requiresInitialization,
            case .uninitialized = state.initalizationState {
             environment.emitDeveloperWarning("SDK must be initialized before usage.")
+            return .none
         }
 
         switch action {
--- a/Sources/KlaviyoSwift/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement.swift
@@ -112,7 +112,6 @@ struct KlaviyoReducer: ReducerProtocol {
         if action.requiresInitialization,
            case .uninitialized = state.initalizationState {
             environment.emitDeveloperWarning("SDK must be initialized before usage.")
-            return .none
         }
 
         switch action {
--- a/Sources/KlaviyoSwift/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement.swift
@@ -94,7 +94,7 @@ enum KlaviyoAction: Equatable {
         switch self {
         case let .enqueueEvent(event):
             // if event metric is opened push we DON'T require initilization in all other event metric cases we DO.
-            return event.metric.name == .OpenedPush ? false : true
+            return event.metric.name != .OpenedPush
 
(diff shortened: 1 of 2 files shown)


Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -92,7 +92,11 @@ enum KlaviyoAction: Equatable {
 
     var requiresInitialization: Bool {
         switch self {
-        case .setEmail, .setPhoneNumber, .setExternalId, .setPushToken, .enqueueEvent, .enqueueProfile, .setProfileProperty, .resetProfile, .resetStateAndDequeue:
+        case let .enqueueEvent(event):
+            // if event metric is opened push we DON'T require initilization in all other event metric cases we DO.
+            return event.metric.name == .OpenedPush ? false : true
```
ajaysubra, line 97: this can be simplified to `return event.metric.name != .OpenedPush` but felt that was not readable.

Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
```diff
@@ -116,7 +116,7 @@ class StateManagementEdgeCaseTests: XCTestCase {
     // MARK: - Set Email
 
     @MainActor
-    func testSetEmailUninitialized() async throws {
```
ajaysubra, line 119: Just updated some test names to indicate the outcome we expect

Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
```diff
@@ -369,15 +369,30 @@ class StateManagementEdgeCaseTests: XCTestCase {
...
+    func testOpenedPushEventUninitializedAddsToPendingRequests() async throws {
+        let store = TestStore(initialState: .init(queue: []), reducer: KlaviyoReducer())
+        let event = Event(name: .OpenedPush)
+        _ = await store.send(.enqueueEvent(event)) {
+            $0.pendingRequests = [.event(event)]
+        }
+    }
+
+    @MainActor
+    func testEnqueueNonOpenedPushEventUninitializedDoesNotAddToPendingRequest() async throws {
```
ajaysubra, line 381: This test goes through all other event metrics other than opened push and makes sure that they don't get through the SDK when it's not initialized.

Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
```diff
@@ -369,15 +369,30 @@ class StateManagementEdgeCaseTests: XCTestCase {
     // MARK: - set enqueue event uninitialized
 
     @MainActor
-    func testEnqueueEventUninitialized() async throws {
+    func testOpenedPushEventUninitializedAddsToPendingRequests() async throws {
```
ajaysubra, line 372: Tests opened push event is added to pending requests when SDK isn't initialized.

Sources/KlaviyoSwift/StateManagement.swift
ndurell: hmm to me that seems more readable maybe

evan-masseau: I'm assuming syntax doesn't allow for something like `case let .enqueueEvent(event) && event.metric.name .OpenedPush :` right? 
Else i agree with Noah on readability

Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
evan-masseau: Nice.

Sources/KlaviyoSwift/StateManagement.swift
ajaysubra: Yeah within a switch that syntax AFAIK isn't possible. Updated to `return event.metric.name != .OpenedPush`. I think I can get used to that.

ndurell: Oh yah this might be possible one sec...

ajaysubra:
```swift
    var requiresInitialization: Bool {
        switch self {
        case let .enqueueEvent(event) where event.metric.name == .OpenedPush:
            // if event metric is opened push we DON'T require initilization in all other event metric cases we DO.
            return false
        
        case .enqueueEvent:
            // if event metric is opened push we DON'T require initilization in all other event metric cases we DO.
            return true

        case .setEmail, .setPhoneNumber, .setExternalId, .setPushToken, .enqueueProfile, .setProfileProperty, .resetProfile, .resetStateAndDequeue:
            return true

        case .initialize, .completeInitialization, .deQueueCompletedResults, .networkConnectivityChanged, .flushQueue, .sendRequest, .stop, .start, .cancelInFlightRequests, .requestFailed:
            return false
        }
    }
```

Something like this?

ajaysubra: May be can combine the`enqueueEvent ` with the rest of the cases that require initilization

ajaysubra:
```swift
    var requiresInitialization: Bool {
        switch self {
        case let .enqueueEvent(event) where event.metric.name == .OpenedPush:
            // if event metric is opened push we DON'T require initilization in all other event metric cases we DO.
            return false
        
        case .setEmail, .setPhoneNumber, .setExternalId, .setPushToken, .enqueueProfile, .setProfileProperty, .resetProfile, .resetStateAndDequeue, .enqueueEvent:
            return true

        case .initialize, .completeInitialization, .deQueueCompletedResults, .networkConnectivityChanged, .flushQueue, .sendRequest, .stop, .start, .cancelInFlightRequests, .requestFailed:
            return false
        }
    }
```

ndurell: yah I like that! forgot about where. yah and move enqueueEvent under the other true case.

ajaysubra: Done!

evan-masseau: Using `where` now? Maybe github is messed up, but i'm not seeing that change

evan-masseau (approved):

ndurell (approved):

PR #183, opened 2024-06-04
Flush requests queue on opened push event

ajaysubra:
# Description
In order to keep engagement metric as up to date as possible we want to try and flush the queue whenever an opened push event request is enqueued. 

# Check List

# Manual Test Plan

Unit tests but also using a proxy - 
1. Increased the flush interval to something very high (100 secs)
2. Enqueued various events and confirmed they didn't flush 
3. Enqueued a opened push event and noticed in the proxy that they all flushed.

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement.swift
@@ -400,7 +400,13 @@ struct KlaviyoReducer: ReducerProtocol {
                                                 endpoint: .createEvent(
                                                     .init(data: .init(event: event, anonymousId: anonymousId))
                                                 )))
-            return .none
+
+            /*
+             if we receive an opened push event we want to flush the queue right away so that
+             we don't miss any user engagement events. In all other cases we will flush the queue
+             using the flush intervals defined above in `StateManagementConstants`
+             */
+            return event.metric.name == .OpenedPush ? .task { .flushQueue } : .none
 
         case let .enqueueProfile(profile):
             guard case .initialized = state.initalizationState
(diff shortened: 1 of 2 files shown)


Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -400,7 +400,13 @@ struct KlaviyoReducer: ReducerProtocol {
...
                                                     .init(data: .init(event: event, anonymousId: anonymousId))
                                                 )))
-            return .none
+
+            /*
+             if we receive an opened push event we want to flush the queue right away so that
+             we don't miss any user engagement events. In all other cases we will flush the queue
+             using the flush intervals defined above in `StateManagementConstants`
+             */
+            return event.metric.name == .OpenedPush ? .task { .flushQueue } : .none
```
ajaysubra, line 409: This is the main change.

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -501,7 +509,7 @@ class StateManagementTests: XCTestCase {
         let initialState = INITILIZING_TEST_STATE()
         let store = TestStore(initialState: initialState, reducer: KlaviyoReducer())
 
-        let event = Event(name: .OpenedPush)
+        let event = Event(name: .OpenedAppMetric)
```
ajaysubra, line 512: change to non opened push event since it doesn't matter for this test which event we use.

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -485,14 +485,22 @@ class StateManagementTests: XCTestCase {
     // MARK: - Test enqueue event
 
     @MainActor
-    func testEnqueueEvent() async throws {
+    func testEnqueueEvents() async throws {
```
ajaysubra, line 488: Updated this test to test all events and added an exception for opened push at the bottom.

ndurell (approved):

PR #187, opened 2024-08-01
retry on 503

ndurell:
# Description

In cases where 503 is returned by the api we are allowed to retry the request. This adds the 503 status to the list of retryables. It also adds some missing test coverage for retryable stuff.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoAPI.swift
+++ b/Sources/KlaviyoSwift/KlaviyoAPI.swift
@@ -68,7 +68,7 @@ struct KlaviyoAPI {
             return .failure(.missingOrInvalidResponse(response))
         }
 
-        if httpResponse.statusCode == 429 {
+        if [429, 503].contains(httpResponse.statusCode) {
             let retryAfter = Int(httpResponse.value(forHTTPHeaderField: "Retry-After") ?? "0")
             requestRateLimited(request, retryAfter)
             return .failure(KlaviyoAPIError.rateLimitError(retryAfter))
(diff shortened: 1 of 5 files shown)


Sources/KlaviyoSwift/KlaviyoAPI.swift
```diff
@@ -68,7 +68,7 @@ struct KlaviyoAPI {
             return .failure(.missingOrInvalidResponse(response))
         }
 
-        if httpResponse.statusCode == 429 {
+        if [429, 503].contains(httpResponse.statusCode) {
```
ajaysubra, line 71: Since we won't have the `Retry-After` for 503 will this retry immediately or use exponential back off?

ndurell: I believe exp backoff.

ndurell: `let nextBackoff = addJitter(to: retryAfter ?? exponentialBackOff)`

ndurell: ```            let retryAfter = Int(httpResponse.value(forHTTPHeaderField: "Retry-After") ?? "0")``` but maybe it's not falsy.

ajaysubra: Yeah I think `nextBackoff` might be 0 in this case.

PR #188, opened 2024-08-20
CHNL-11081: Modularize iOS SDK into `KlaviyoCore`, `KlaviyoSwift` and `KlaviyoSwiftExtension`

ajaysubra:
# Description

This PR aims to modularize the SDK into, 
1. `KlaviyoCore` - Base layer module that contains non-business logic like networking, API models, other dependencies like `KlaviyoEnvironment` that have potential for reuse. 
2. `KlaviyoSwift` - Contains most of the business logic for handling analytics (profile identification, events tracking, push token register and unregister). Also some convenience methods for handling push engagement like open tracking etc. In addition this module will house the public facing data models, state management, queuing mechanism etc. 
3. `KlaviyoSwiftExtension` - Nothing changes here.

image

File system changes are as follows,

```
klaviyo-swift-sdk/
├── Sources/
│   ├── KlaviyoCore/
│   ├── KlaviyoSwift/
│   ├── KlaviyoSwiftExtension/
├── Tests/
│   ├── KlaviyoCoreTests/
│   ├── KlaviyoSwiftTests/
│   ├── KlaviyoSwiftExtensionTests/
└── Package.swift
```

Highlighting some significant changes done as part of this PR, 
1. The API models were nested which was compact and the organization was sufficient but was impacting readability in cases where we needed to reference types that were deeply nested. Broke out these models into their own types independent of `KlaviyoAPI`
2. Moved some of the dependencies in `AnalyticsEnvironment` over to `KlaviyoEnvironment` and create a new environment specifically for `KlaviyoSwift` called `KlaviyoSwiftEnvironment`. This was done to keep dependencies organized under one umbrella instead of two since the use cases of each of them were identical. 
3. Moved the network retry logic over to `KlaviyoCore` since this is something that could potentially be reused. In addition to this, also include retying on HTTP status code 503.
4. Removed the dependencies internal API models had on public facing models. In order to map the public models to internal API models created extensions on public data models to map to internal API models.

# Check List

# Manual Test Plan

1.

# Supporting Materials


Package.swift
```diff
@@ -7,6 +7,9 @@ let package = Package(
     name: "klaviyo-swift-sdk",
     platforms: [.iOS(.v13)],
     products: [
+        .library(
```
ajaysubra, line 10: SPM package changes to create new core module and establish the dependencies mentioned in the PR description.

Note to self - need to update cocoapods as well to mirror SPM changes.

Sources/KlaviyoCore/AppContextInfo.swift
```diff
@@ -7,18 +7,18 @@
 import Foundation
 import UIKit
 
-struct AppContextInfo {
```
ajaysubra, line 10: Moved this file to core and exposing it's members publicly so that `KlaviyoSwift` can access them. This is the case of all the files that were moved to `KlaviyoCore` and need to be accessed from `KlaviyoSwift`

Sources/KlaviyoCore/AppLifeCycleEvents.swift
```diff
@@ -9,18 +9,27 @@ import Combine
...
-struct AppLifeCycleEvents {
-    var lifeCycleEvents: () -> AnyPublisher<KlaviyoAction, Never> = {
+public enum LifeCycleEvents {
+    case terminated
+    case forgrounded
+    case backgrounded
+    case reachabilityChanged(status: Reachability.NetworkStatus)
+}
+
+public struct AppLifeCycleEvents {
```
ajaysubra, line 23: In addition to moving this file out of `KlaviyoSwift` I also had to remove its dependence on `KlaviyoAction` that is part of the TCA architecture in `KlaviyoSwift`. To achieve this created a new type called `LifeCycleEvents` that is then mapped to `KlaviyoAction` in `KlaviyoSwift`

Sources/KlaviyoCore/AppLifeCycleEvents.swift
```diff
@@ -30,23 +39,21 @@ struct AppLifeCycleEvents {
...
         // The below is a bit convoluted since network status can be nil.
         let reachability = environment
             .notificationCenterPublisher(ReachabilityChangedNotification)
-            .compactMap { _ -> KlaviyoAction? in
-                guard let status = environment.reachabilityStatus() else {
-                    return nil
-                }
-                return KlaviyoAction.networkConnectivityChanged(status)
+            .map { _ in
+                // TODO: compact map isn't working, need to fix
```
ajaysubra, line 53: Note to self

Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -0,0 +1,208 @@
...
+
+    public var jsonDecoder: JSONDecoder
+    public static let production = Self(jsonDecoder: KlaviyoEnvironment.decoder)
+
+    public func decode<T: Decodable>(_ data: Data) throws -> T {
+        try jsonDecoder.decode(T.self, from: data)
+    }
+}
+
+@usableFromInline
```
ajaysubra, line 190: note to self: potential to move this out of here.

Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -0,0 +1,208 @@
...
+//  Created by Noah Durell on 9/28/22.
+//
+
+import AnyCodable
+import Combine
+import Foundation
+import UIKit
+#if canImport(os)
+import os
+#endif
```
ajaysubra, line 14: note to self: potential to move this out of here.

Sources/KlaviyoCore/Models/APIModels/CreateEventPayload.swift
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    public var data: Event
+    public init(data: Event) {
+        self.data = data
+    }
+}
+
+extension Dictionary where Key == String, Value == Any {
+    fileprivate func appendMetadataToProperties(pushToken: String?) -> [String: Any]? {
```
ajaysubra, line 115: Earlier this is explicitly called from `KlaviyoAPI` after the `CreateEventPayload` instance was created. Seemed like it was possible to simply include it _silently_ when the model is initialized with properties that the user supplies with public facing `Event` data model.

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -0,0 +1,51 @@
...
+import Foundation
+import KlaviyoCore
+
+extension String {
+    fileprivate func returnNilIfEmpty() -> String? {
+        isEmpty ? nil : self
+    }
+}
+
+extension Profile {
```
ajaysubra, line 17: This is the extension mentioned in the PR description that helps map the public facing API to API models.

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -42,48 +42,6 @@ struct KlaviyoState: Equatable, Codable {
         }
     }
 
-    enum PushEnablement: String, Codable {
```
ajaysubra, line 45: moved to core

Package.swift
ajaysubra: updated sdk to work with cocoapods.

Sources/KlaviyoCore/KlaviyoEnvironment.swift
ajaysubra: moved to logging client

Sources/KlaviyoCore/KlaviyoEnvironment.swift
ajaysubra: moved to logging client

Tests/KlaviyoSwiftTests/KlaviyoStateTests.swift
```diff
@@ -113,33 +113,40 @@ final class KlaviyoStateTests: XCTestCase {
...
-            try! JSONEncoder().encode(KlaviyoState(apiKey: "foo", anonymousId: environment.analytics.uuid().uuidString, queue: [], requestsInFlight: []))
+        environment.dataFromUrl = { _ in
+            try! JSONEncoder().encode(KlaviyoState(
+                apiKey: "foo",
+                anonymousId: environment.uuid().uuidString,
+                queue: [],
+                requestsInFlight: []))
         }
-        environment.analytics.decoder = DataDecoder(jsonDecoder: KlaviyoEnvironment.decoder)
+//        environment.decoder = DataDecoder(jsonDecoder: KlaviyoEnvironment.decoder)
```
ab1470, line 123: Is this intentional?

ab1470 (approved):
To my (admittedly new) eyes, this looks like a great architecture improvement! I left some small comments, but otherwise it looks good to me.

ab1470:
One more question: what's the difference between `KlaviyoEnvironment` and `KlaviyoSwiftEnvironment`?

Sources/KlaviyoSwift/KlaviyoSwiftEnvironment.swift
```diff
@@ -0,0 +1,30 @@
...
+//  KlaviyoSwiftEnvironment.swift
+//
+//
+//  Created by Ajay Subramanya on 8/8/24.
+//
+
+import Combine
+import Foundation
+
+var klaviyoSwiftEnvironment = KlaviyoSwiftEnvironment.production
```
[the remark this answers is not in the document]
ajaysubra: Ideally yes but looks like some tests override some of its properties. 

```
klaviyoSwiftEnvironment.send = {
    test.send($0)
}
```

Sources/KlaviyoSwift/Models/Event.swift
```diff
@@ -0,0 +1,98 @@
...
+         properties: [String: Any]? = nil,
+         identifiers: Identifiers? = nil,
+         value: Double? = nil,
+         time: Date? = nil,
+         uniqueId: String? = nil) {
+        metric = .init(name: name)
+        _properties = AnyCodable(properties ?? [:])
+        self.time = time ?? environment.date()
+        self.value = value
+        self.uniqueId = uniqueId ?? environment.uuid().uuidString
```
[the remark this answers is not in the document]
ajaysubra: 1. This is an internal initializer so I think we are okay with not having comments. 
2. Agree. This makes sense.

Sources/KlaviyoSwift/Models/Event.swift
```diff
@@ -0,0 +1,98 @@
...
+        self.value = value
+        time = environment.date()
+        self.uniqueId = uniqueId ?? environment.uuid().uuidString
+    }
+}
+
+extension Event.EventName {
+    public var value: String {
+        switch self {
+        case .OpenedPush: return "$opened_push"
```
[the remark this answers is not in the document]
ajaysubra: Good question. These are klaviyo canonical events that the API expects to be in a certain format. I agree with you that it's inconsistent but changing it now would involve some work since we'd also need to update the API but also make sure to accept the snake case until all SDKs out in the wild are cut over to using the natural language version.

Sources/KlaviyoSwift/Models/Profile.swift
```diff
@@ -0,0 +1,120 @@
...
+                    timezone: String? = nil) {
+            self.address1 = address1
+            self.address2 = address2
+            self.city = city
+            self.country = country
+            self.latitude = latitude
+            self.longitude = longitude
+            self.region = region
+            self.zip = zip
+            self.timezone = timezone ?? environment.timeZone()
```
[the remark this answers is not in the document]
ajaysubra: minor nit pick push back - would that be a breaking change to the public API? This is unlikely but if someone was passing in nil this would break for them. We may have to do a breaking release if we did this.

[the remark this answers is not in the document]
ab1470: Ah I see, yeah it's going to be a bit of an adjustment for me to be thinking in terms of external SDK consumers. 

This would be an easy fix for any consumers, because it'll throw a compiler error. But maybe not worth making this change if we don't want to bother them with it

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -0,0 +1,51 @@
...
+    func toAPIModel(
+        email: String? = nil,
+        phoneNumber: String? = nil,
+        externalId: String? = nil,
+        anonymousId: String) -> ProfilePayload {
+        ProfilePayload(
+            email: email ?? self.email?.returnNilIfEmpty(),
+            phoneNumber: phoneNumber ?? self.phoneNumber?.returnNilIfEmpty(),
+            externalId: externalId ?? self.externalId?.returnNilIfEmpty(),
```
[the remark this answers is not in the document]
ajaysubra: Yeah initially I had it the way you are suggesting but it would fail this test, 

```swift 
    func testSetProfileWithEmptyStringIdentifiers() async throws {
        let initialState = KlaviyoState(
            apiKey: TEST_API_KEY,
            email: "foo@bar.com",
            anonymousId: environment.uuid().uuidString,
            phoneNumber: "99999999",
            externalId: "12345",
            pushTokenData: .init(pushToken: "blob_token",
                                 pushEnablement: .authorized,
                                 pushBackground: .available,
                                 deviceData: .init(context: environment.appContextInfo())),
            queue: [],
            requestsInFlight: [],
            initalizationState: .initialized,
            flushing: true)

        _ = await store.send(.enqueueProfile(Profile(email: "", phoneNumber: "", externalId: ""))) {
            $0.email = nil // since we reset state
            $0.phoneNumber = nil // since we reset state
            $0.externalId = nil // since we reset state
            $0.enqueueProfileOrTokenRequest()
            $0.pushTokenData = nil
        }
    }
```

I think the test makes sense since when you enqueue a profile (developer identifying a new profile from what is currently in state) we reset out local SDK state and assign whatever the developer is now sending to us but for whatever reason they send an empty string it's going to fail in the API so it's better to convert that to nil than to have make a network call and fail.

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -0,0 +1,51 @@
...
+    var toAPILocation: ProfilePayload.Attributes.Location {
+        ProfilePayload.Attributes.Location(
+            address1: address1,
+            address2: address2,
+            city: city,
+            country: country,
+            latitude: latitude,
+            longitude: longitude,
+            region: region,
+            zip: self.zip,
```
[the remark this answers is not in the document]
ajaysubra: Agree!

Tests/KlaviyoSwiftTests/KlaviyoStateTests.swift
ajaysubra: Sorta! it was there before and the test passed w/o it so kept it around to make sure I'm not missing something. Will take a second look and remove it if it's absolutely not needed.

ajaysubra:
> One more question: what's the difference between `KlaviyoEnvironment` and `KlaviyoSwiftEnvironment`?

Yeah good question. `KlaviyoEnvironment` is the overarching environment which contains all the dependencies that `KlaviyoCore` provides to `KlaviyoSwift`. Whereas, `KlaviyoSwiftEnvironment` contains dependencies that `KlaviyoSwift` requires within it's module. 

In order words, `KlaviyoSwiftEnvironment` contains depencies that `KlaviyoSwift` needs that can't (or doesn't make sense to move to `KlaviyoCore`) be moved over to `KlaviyoCore` since it uses some local types. Ex: `var stateChangePublisher: () -> AnyPublisher<KlaviyoAction, Never>` 

I will acknowledge that the naming of these can be improved to better reflect their intent. Let me know if you can think of something or if we could add some comments to make this clearer to other developers.

Sources/KlaviyoCore/AppContextInfo.swift
```diff
@@ -7,18 +7,18 @@
...
-    private static let info = Bundle.main.infoDictionary
-    private static let defaultExecutable: String = (info?["CFBundleExecutable"] as? String) ??
+public struct AppContextInfo {
+    private let info = Bundle.main.infoDictionary
```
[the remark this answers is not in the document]
ajaysubra: Nice catch and thanks for the suggestions. Followed (2).

Sources/KlaviyoCore/AppLifeCycleEvents.swift
```diff
@@ -9,18 +9,27 @@ import Combine
...
-enum LifeCycleErrors: Error {
+public enum LifeCycleErrors: Error {
     case invalidReachaibilityStatus
 }
 
-struct AppLifeCycleEvents {
-    var lifeCycleEvents: () -> AnyPublisher<KlaviyoAction, Never> = {
+public enum LifeCycleEvents {
+    case terminated
+    case forgrounded
```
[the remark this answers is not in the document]
ajaysubra: Don't be sorry. It's my bad that I can't spell to save my life! 

Fixed it and pushed. Didn't commit your changes since I made some locally and didn't want to deal with pulling and pushing. Thanks for the commit suggestions nonetheless.

Sources/KlaviyoSwift/Models/Event.swift
```diff
@@ -0,0 +1,98 @@
...
+        }
+    }
+
+    public let metric: Metric
+    public var properties: [String: Any] {
+        _properties.value as! [String: Any]
+    }
+
+    private let _properties: AnyCodable
+    public var time: Date
```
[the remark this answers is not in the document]
ajaysubra: yeah you are right. Fixed it. Nice catch.

Sources/KlaviyoSwift/Models/Event.swift
```diff
@@ -0,0 +1,98 @@
...
+        _properties = AnyCodable(properties ?? [:])
+        self.time = time ?? environment.date()
+        self.value = value
+        self.uniqueId = uniqueId ?? environment.uuid().uuidString
+        self.identifiers = identifiers
+    }
+
+    /// Create a new event to track a profile's activity, the SDK will associate the event with any identified/anonymous profile in the SDK state
+    /// - Parameters:
+    ///   - name: Name of the event. Must be less than 128 characters., pick from `Event.EventName` which can also contain custom events
```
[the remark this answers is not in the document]
ajaysubra: Nice find. I didn't know you could do this. Commited your change.

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -0,0 +1,51 @@
...
+import KlaviyoCore
+
+extension String {
+    fileprivate func returnNilIfEmpty() -> String? {
+        isEmpty ? nil : self
+    }
+}
+
+extension Profile {
+    func toAPIModel(
```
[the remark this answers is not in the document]
ajaysubra: Good call. It was covered with some existing tests but you are right. It's easier to test this mapping in isolation. Added in some tests to a new file - `KlaviyoModelsTest`

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -0,0 +1,51 @@
...
+            title: title,
+            image: image,
+            location: location?.toAPILocation,
+            properties: properties,
+            anonymousId: anonymousId)
+    }
+}
+
+extension Profile.Location {
+    var toAPILocation: ProfilePayload.Attributes.Location {
```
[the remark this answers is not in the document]
ajaysubra: same as above.

Sources/KlaviyoCore/AppContextInfo.swift
```diff
@@ -32,36 +32,36 @@ struct AppContextInfo {
...
-    let appVersion: String
-    let appBuild: String
-    let appName: String
-    let version: OperatingSystemVersion
-    let osName: String
-    let manufacturer: String
-    let deviceModel: String
-    let deviceId: String
-    let environment: String
+    public let executable: String
```
ndurell, line 35: We talked about this offline. Just make sure it's not exposed publicly outside the sdk. I think it's not!

Sources/KlaviyoCore/Models/APIModels/CreateEventPayload.swift
ndurell: Ok just fyi might result in some stale data and slightly different behavior. We will essentially be capturing a point in time vs the most up to date device info.

KlaviyoSwift.podspec
```diff
@@ -14,7 +14,7 @@ Pod::Spec.new do |s|
   s.swift_version = '5.7'
   s.platform = :ios
   s.ios.deployment_target = '13.0'
-  s.source_files = 'Sources/KlaviyoSwift/**/*.swift'
+  s.source_files = 'Sources/**/**/*.swift'
```
ndurell, line 17: Should we do a list here or a regex. I'm wondering if this would capture tests too. Maybe that has a different root directory.

Sources/KlaviyoCore/Models/KlaviyoAPIError.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  KlaviyoAPIError.swift
+//
+//
+//  Created by Ajay Subramanya on 8/8/24.
+//
+
+import Foundation
+
+public enum KlaviyoAPIError: Error {
```
ndurell, line 10: This has to be public?

Sources/KlaviyoCore/Networking/SDKRequestIterator.swift
```diff
@@ -128,7 +128,6 @@ public struct SDKRequest: Identifiable, Equatable {
     }
 }
 
-@_spi(KlaviyoPrivate)
```
ndurell, line 131: Hmm does the test app not need this?

Sources/KlaviyoCore/Utils/Version.swift
ndurell, line 1: Hmm should this move? We are technically versioning KlaviyoSwift not core.

Sources/KlaviyoCore/Vendor/ReachabilitySwift.swift
```diff
@@ -48,16 +47,39 @@ func callback(reachability:SCNetworkReachability, flags: SCNetworkReachabilityFl
     }
 }
 
-class Reachability {
-
-    typealias NetworkReachable = (Reachability) -> ()
-    typealias NetworkUnreachable = (Reachability) -> ()
+public class Reachability {
```
ndurell, line 50: Was this all from linting?

Sources/KlaviyoSwift/Vendor/ComposableArchitecture/Misc.swift
```diff
@@ -51,7 +51,7 @@ final class Box<Wrapped> {
 @inline(__always)
 func runtimeWarn(
   _ message: @autoclosure () -> String,
-  category: String? = __klaviyoSwiftName,
+  category: String? = "",
```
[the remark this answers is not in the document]
ndurell: Hmm will it not have a category in the logs now?

Tests/KlaviyoCoreTests/ArchivalUtilsTests.swift
```diff
@@ -105,7 +106,6 @@ class ArchivalSystemTest: XCTestCase {
     let TEST_URL = filePathForData(apiKey: "foo", data: "people")
 
     override func setUpWithError() throws {
-        environment = KlaviyoEnvironment.production
```
ndurell, line 108: why remove this?

Tests/KlaviyoCoreTests/__Snapshots__/EncodableTests/testEventPayloadWithoutMetadata.1.json
```diff
@@ -21,23 +21,11 @@
         }
       },
       "properties" : {
-        "App Build" : "1",
```
ndurell, line 24: Expected yes?

Tests/KlaviyoCoreTests/__Snapshots__/EncodableTests/testEventPayloadWithoutMetadata.1.json
```diff
@@ -4,7 +4,7 @@
       "metric" : {
         "data" : {
           "attributes" : {
-            "name" : "blob"
+            "name" : "test"
```
ndurell, line 7: Maybe a little weird...did you change this test data?

Tests/KlaviyoCoreTests/__Snapshots__/EncodableTests/testEventPayloadWithoutMetadata.1.json
```diff
@@ -21,23 +21,11 @@
...
       },
       "properties" : {
-        "App Build" : "1",
-        "App Name" : "FooApp",
-        "App Version" : "1.2.3",
-        "Application ID" : "com.klaviyo.fooapp",
-        "blob" : "blob",
-        "Device Manufacturer" : "Orange",
-        "Device Model" : "jPhone 1,1",
-        "hello" : {
```
ndurell, line 31: Yah something might be wrong here...you are missing custom properties too..

Tests/KlaviyoSwiftTests/APIRequestErrorHandlingTests.swift
```diff
@@ -266,15 +267,15 @@ class APIRequestErrorHandlingTests: XCTestCase {
         initialState.requestsInFlight = [request]
         let store = TestStore(initialState: initialState, reducer: KlaviyoReducer())
 
-        environment.analytics.klaviyoAPI.send = { _, _ in .failure(.rateLimitError(nil)) }
+        environment.klaviyoAPI.send = { _, _ in .failure(.rateLimitError(backOff: 30)) }
```
ndurell, line 270: Did you mean to change this? Are you supporting 503 handling here?

Tests/KlaviyoSwiftTests/__Snapshots__/KlaviyoStateTests/testValidStateFileExists.1.txt
```diff
@@ -3,15 +3,35 @@
     - some: "00000000-0000-0000-0000-000000000001"
   ▿ apiKey: Optional<String>
     - some: "foo"
-  - email: Optional<String>.none
-  - externalId: Optional<String>.none
+  ▿ email: Optional<String>
```
ndurell, line 6: expected?

ndurell (approved):

Sources/KlaviyoCore/AppContextInfo.swift
ajaysubra: so looks these particular properties didn't need to be public but other similar public methods here are accessible outside. Will see how I can try and hide them if I can.

KlaviyoSwift.podspec
ajaysubra: yeah tests are a different directory. 

```
klaviyo-swift-sdk/
├── Sources/
│   ├── KlaviyoCore/
│   ├── KlaviyoSwift/
│   ├── KlaviyoSwiftExtension/
│   └── KlaviyoUI/
├── Tests/
│   ├── KlaviyoCoreTests/
│   ├── KlaviyoSwiftTests/
│   ├── KlaviyoSwiftExtensionTests/
│   └── KlaviyoUITests/
└── Package.swift

```

Sources/KlaviyoCore/Networking/SDKRequestIterator.swift
ajaysubra: Hmm yeah I had the same question, turns out it doesn't as long as it can access the `requestIterator`

ndurell: Ok but we aren't exposing anything in the public api that we shouldn't be right?

Sources/KlaviyoCore/Utils/Version.swift
ajaysubra: Yeah, I thought about this a bit and ended up moving for these reasons, 
* This is used in `KlaviyoEnvironment`, it can be passed in from KlaviyoSwift. 
* If we had a different module for a different use case, I wonder if we would version that differently? The way I'm thinking is this is the SDK version instead of any one module in the SDK. 

Let me know what you think?

Sources/KlaviyoCore/AppLifeCycleEvents.swift
ajaysubra: I was having compiler errors earlier but switched back to compact map and it's working now.

Sources/KlaviyoCore/Models/APIModels/CreateEventPayload.swift
ajaysubra: Oh I didn't think about it that way at all. That makes a lot of sense - I'll revert this logic.

Sources/KlaviyoSwift/Vendor/ComposableArchitecture/Misc.swift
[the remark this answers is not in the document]
ajaysubra: I moved this to LoggerClient and this is what is used in the environment's `emitDeveloperWarning`. Please let me know if I'm missing something?

```swift
import Foundation
#if canImport(os)
import os
#endif

public struct LoggerClient {
    public init(error: @escaping (String) -> Void) {
        self.error = error
    }

    public var error: (String) -> Void
    public static let production = Self(error: { message in os_log("%{public}s", type: .error, message) })
}

@usableFromInline
@inline(__always)
func runtimeWarn(
    _ message: @autoclosure () -> String,
    category: String? = environment.sdkName(),
    file: StaticString? = nil,
    line: UInt? = nil) {
    #if DEBUG
    let message = message()
    let category = category ?? "Runtime Warning"
    #if canImport(os)
    os_log(
        .fault,
        log: OSLog(subsystem: "com.apple.runtime-issues", category: category),
        "%@",
        message)
    #endif
    #endif
}

```

Tests/KlaviyoSwiftTests/KlaviyoStateTests.swift
ajaysubra: Didn't look like it was needed. Removed it.

Sources/KlaviyoCore/Models/KlaviyoAPIError.swift
ajaysubra: yeah since it's used in `KlaviyoAPI` 
`public var send: (KlaviyoRequest, Int) async -> Result<Data, KlaviyoAPIError>`

Sources/KlaviyoCore/Vendor/ReachabilitySwift.swift
ajaysubra: Sorry it's a little confusing. I updated `Reachability` and certain properties within it as well we use this in something that KlaviyoSwift uses like, 

```swift 
public enum LifeCycleEvents {
    case terminated
    case foregrounded
    case backgrounded
    case reachabilityChanged(status: Reachability.NetworkStatus)
}
```

Since I made this change, I think lint formatting the entire file.

Tests/KlaviyoCoreTests/ArchivalUtilsTests.swift
ajaysubra: I don't remember now but may be the test was passing and indeed using the production environment in `unarchiveFromFile` but might be good to explicitly specify it just in case some other test overrides the `environment` variable and renders this test as flaky. Will add it back in.

Tests/KlaviyoCoreTests/__Snapshots__/EncodableTests/testEventPayloadWithoutMetadata.1.json
ajaysubra: Yeah looks like,  looks like when I switched out the models I used a different name - 

Older tests 

```swift
func testEventPayloadWithoutMetadata() throws {
    let event = Event.test
    let createEventPayload = KlaviyoAPI.KlaviyoRequest.KlaviyoEndpoint.CreateEventPayload(data: .init(event: event, anonymousId: "anon-id"))
    assertSnapshot(matching: createEventPayload, as: .json(KlaviyoEnvironment.encoder))
}


extension Event {
    static let SAMPLE_PROPERTIES = [
        "blob": "blob",
        "stuff": 2,
        "hello": [
            "sub": "dict"
        ],
        "Application ID": "com.klaviyo.fooapp",
        "App Version": "1.2.3",
        "App Build": "1",
        "App Name": "FooApp",
        "OS Version": "1.1.1",
        "OS Name": "iOS",
        "Device Manufacturer": "Orange",
        "Device Model": "jPhone 1,1"
    ] as [String: Any]
    static let test = Self(name: .CustomEvent("blob"), properties: SAMPLE_PROPERTIES)
}
```

Current tests 

```swift
    func testEventPayloadWithoutMetadata() throws {
        let createEventPayload = CreateEventPayload(data: CreateEventPayload.Event(name: "test", anonymousId: "anon-id"))
        assertSnapshot(matching: createEventPayload, as: .json(KlaviyoEnvironment.encoder))
    }
```

Tests/KlaviyoCoreTests/__Snapshots__/EncodableTests/testEventPayloadWithoutMetadata.1.json
ajaysubra: yeah looks like the old event that was being used had them but I don't seem to be passing that custom properties to the new initializer. WIll add it in.

Tests/KlaviyoSwiftTests/APIRequestErrorHandlingTests.swift
ajaysubra: Sorry, I don't think I'm following. Yes there is logic to handle 503 now, I should probably add a test for that, but not if that's what you are asking?

ajaysubra: Oh I see your point now. Essentially, you are saying this test should be like the rate limit error tests since we are retrying on 503. I'll look into it.

ajaysubra: Okay looked into this and turns out this test also covers the 503 case. For both 429 and 503, `environment.klaviyoAPI.send ` returns `KlaviyoAPIError.rateLimitError(backOff: nextBackOffWithJitter)`.

Tests/KlaviyoCoreTests/__Snapshots__/EncodableTests/testEventPayloadWithoutMetadata.1.json
ajaysubra: Added in properties to events. Now it looks like this, 

```json
{
  "data" : {
    "attributes" : {
      "metric" : {
        "data" : {
          "attributes" : {
            "name" : "test"
          },
          "type" : "metric"
        }
      },
      "profile" : {
        "data" : {
          "attributes" : {
            "anonymous_id" : "anon-id",
            "properties" : {

            }
          },
          "type" : "profile"
        }
      },
      "properties" : {
        "App Build" : "1",
        "App ID" : "com.klaviyo.fooapp",
        "App Name" : "FooApp",
        "App Version" : "1.2.3",
        "blob" : "blob",
        "Device ID" : "fe-fi-fo-fum",
        "Device Manufacturer" : "Orange",
        "Device Model" : "jPhone 1,1",
        "hello" : {
          "sub" : "dict"
        },
        "OS Name" : "iOS",
        "OS Version" : "1.1.1",
        "Push Token" : "",
        "SDK Name" : "swift",
        "SDK Version" : "3.2.0",
        "stuff" : 2
      },
      "time" : "2009-02-13T23:31:30Z",
      "unique_id" : "00000000-0000-0000-0000-000000000001"
    },
    "type" : "event"
  }
}

```

Tests/KlaviyoCoreTests/__Snapshots__/EncodableTests/testEventPayloadWithoutMetadata.1.json
ajaysubra: Replaced this with a test called `testEventPayload` since we are including metadata when creating the event request object and also added sample properties that got removed back and now the JSON looks like, 

```json

{
  "data" : {
    "attributes" : {
      "metric" : {
        "data" : {
          "attributes" : {
            "name" : "test"
          },
          "type" : "metric"
        }
      },
      "profile" : {
        "data" : {
          "attributes" : {
            "anonymous_id" : "anon-id",
            "properties" : {

            }
          },
          "type" : "profile"
        }
      },
      "properties" : {
        "App Build" : "1",
        "App ID" : "com.klaviyo.fooapp",
        "App Name" : "FooApp",
        "App Version" : "1.2.3",
        "blob" : "blob",
        "Device ID" : "fe-fi-fo-fum",
        "Device Manufacturer" : "Orange",
        "Device Model" : "jPhone 1,1",
        "hello" : {
          "sub" : "dict"
        },
        "OS Name" : "iOS",
        "OS Version" : "1.1.1",
        "Push Token" : "",
        "SDK Name" : "swift",
        "SDK Version" : "3.2.0",
        "stuff" : 2
      },
      "time" : "2009-02-13T23:31:30Z",
      "unique_id" : "00000000-0000-0000-0000-000000000001"
    },
    "type" : "event"
  }
}

```

ajaysubra: Thanks for calling these out!

Sources/KlaviyoCore/Models/APIModels/CreateEventPayload.swift
ajaysubra: Based on conversation on slack keeping this as is as we want the metadata to reflect the state of the device when the event was enqueued vs. dequeued.

PR #193, opened 2024-08-28
CHNL-4403: Adding private method

ajaysubra:
# Description

Adding a private method to set the SDK name and version. 

# Check List

# Manual Test Plan

1.

# Supporting Materials


Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -111,6 +115,9 @@ public struct KlaviyoEnvironment {
     public var klaviyoAPI: KlaviyoAPI
     public var timer: (Double) -> AnyPublisher<Date, Never>
 
+    public var SDKName: String
```
ndurell, line 118: for consistency I might make these methods with default implementations.

Sources/KlaviyoCore/Models/APIModels/CreateEventPayload.swift
```diff
@@ -120,8 +120,8 @@ extension Dictionary where Key == String, Value == Any {
             "Device Model": context.deviceModel,
             "OS Name": context.osName,
             "OS Version": context.osVersion,
-            "SDK Name": __klaviyoSwiftName,
-            "SDK Version": __klaviyoSwiftVersion,
+            "SDK Name": environment.SDKName,
```
ndurell, line 123: I think we use version and sdk name in the user agent too - should it be updated there?

ndurell (approved):
This is actually a nice change since we won't have to recreate artifacts on every sdk release.

Sources/KlaviyoCore/Models/APIModels/CreateEventPayload.swift
ajaysubra: Thanks for that callout. Yeah I missed that and updated it now.

Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -111,6 +115,9 @@ public struct KlaviyoEnvironment {
...
+    public var SDKVersion: String
```
ab1470, line 119: How about changing to `sdkName` and `sdkVersion`, to be consistent with Swift naming conventions?

Sources/KlaviyoCore/Networking/PrivateMethods.swift
```diff
@@ -0,0 +1,27 @@
...
+//  Created by Ajay Subramanya on 8/29/24.
+//
+
+import Foundation
+
+@_spi(KlaviyoPrivate)
+/// Used to override SDK defailts for INTERNAL USE ONLY
+/// - Parameter url: The  URL to use for Klaviyo client APIs, This is used internally to test the SDK against different backends, DO NOT use this in your apps.
+/// - Parameter name: The name of the SDK, defaults to swift but react native will pass it's own name. DO NOT override this in your apps as our backend will not accept unsupported values here and your network requests will fail.
+/// - Parameter version: The version of the swift SDK default to hard coded values here but react native will pass it's own values here. DO NOT override this in your apps.
```
ab1470, line 14:
```suggestion
/// Used to override SDK defailts for INTERNAL USE ONLY
/// - Parameter url: The  URL to use for Klaviyo client APIs, This is used internally to test the SDK against different backends, DO NOT use this in your apps.
/// - Parameter name: The name of the SDK, defaults to swift but react native will pass it's own name. DO NOT override this in your apps as our backend will not accept unsupported values here and your network requests will fail.
/// - Parameter version: The version of the swift SDK default to hard coded values here but react native will pass it's own values here. DO NOT override this in your apps.
@_spi(KlaviyoPrivate)
```
nit: I was playing around and it looks like the @_spi tag needs to go below the documentation in order for the documentation to show up properly. See these screenshots:

<table>
  <tr>
    <td>@_spi tag above documentation</td>
     <td>@_spi tag below documentation</td>
  </tr>
  <tr>
    <td></td>
    <td></td>
  </tr>
 </table>

Sources/KlaviyoCore/Networking/PrivateMethods.swift
```diff
@@ -0,0 +1,27 @@
...
+public func overrideSDKDefaults(url: String? = nil, name: String? = nil, version: String? = nil) {
```
ab1470, line 15: would it be useful to show a warning to any callers of this function, telling them that it's for internal use only? If it would be useful, we could add something like
``` swift
@available(*, deprecated, message: "This function is for internal use only, and should not be used in production")
```
Screenshot 2024-08-29 at 14 40 52

ab1470 (approved):

Sources/KlaviyoCore/KlaviyoEnvironment.swift
ajaysubra: Changed these but also changed `apiUrl`.

Sources/KlaviyoCore/Networking/PrivateMethods.swift
ndurell: wow great find!

ajaysubra: Weird, I used the xcode shortcut (cmd + opt + /) to generate comments and it pushed the @_spi above the comments. Thanks for looking into this locally. I'll update it.

Sources/KlaviyoCore/Networking/PrivateMethods.swift
ajaysubra: Nice I like this. Adding it in now, we will see this on our test app but we can live with that.

PR #194, opened 2024-08-28
[CHNL-5395] automatically track notification permission status

ab1470:
# Description

With this PR, the SDK will check the user's push enablement status every time the host app is foregrounded. If the status has changed, the SDK will send the new status to the server.

This addresses [CHNL-5395](https://klaviyo.atlassian.net/browse/CHNL-5395).

# Check List

# Supporting Materials
This video shows that the app updates the enablement status when it's foregrounded
https://github.com/user-attachments/assets/bb1ffd56-a9d8-4465-962e-5cd487eef625


Sources/KlaviyoSwift/StateManagement.swift
```diff
@@ -230,6 +233,15 @@ struct KlaviyoReducer: ReducerProtocol {
             state.enqueueRequest(request: request)
             return .none
 
+        case let .setPushEnablement(enablement):
+            guard let pushToken = state.pushTokenData?.pushToken else {
+                return .none
+            }
+
+            return .run { send in
```
ndurell, line 241: You might be missing some coverage on this bit. You could add a test that send this directly or just change the state so that you have a push token in one of the tests.

ab1470: should be addressed now, thanks

ndurell (approved):

ajaysubra (approved):
Thanks for adding in the demo video. Looks great.

PR #195, opened 2024-09-03
[CHNL-6170] remove opened push event from public api

ab1470:
# Description

This PR removes the `OpenedPush` case from the `Event.EventName` enum, and replaces it with an internal case `._openedPush`. This addresses [CHNL-6170](https://klaviyo.atlassian.net/browse/CHNL-6170).

(note: I'm requesting to merge this into @ajaysubra's modularization branch, not into `master`)

# Check List

# Supporting Materials

We're able to switch over the internal `._openedPush` case when accessing it internally within the SDK codebase:
Screenshot 2024-09-03 at 08 51 04

...but we'll get a compiler error when we attempt to access the `._openedPush` case from an external consumer of the SDK:
Screenshot 2024-09-03 at 08 50 28

[CHNL-6170]: https://klaviyo.atlassian.net/browse/CHNL-6170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/KlaviyoSwift/Models/Event.swift
```diff
@@ -11,12 +11,15 @@ import KlaviyoCore
...
 public struct Event: Equatable {
     public enum EventName: Equatable {
-        case OpenedPush
         case OpenedAppMetric
         case ViewedProductMetric
         case AddedToCartMetric
         case StartedCheckoutMetric
         case CustomEvent(String)
+
+        internal static var _openedPush: EventName {
```
ajaysubra, line 20: Neat. Any chance you tested this with `client/events` and confirmed that it works? I.E we should we seeing a canonical opened push event in the profile for which we are tracking the open. Ex: 

image

ab1470: Yes, just validated that we're successfully tracking the Opened Push event

ajaysubra: Thanks for checking.

ajaysubra (approved):

ndurell (approved):
Looks good I assume this would have to be at least a minor version upgrade if not major right since we are breaking the api. We didn't want to deprecate the old names first right?

ajaysubra:
Oh yeah totally forgot about that. Maybe we can push a patch release with a deprecation warning on current major version?

> Looks good I assume this would have to be at least a minor version upgrade if not major right since we are breaking the api. We didn't want to deprecate the old names first right?

PR #201, opened 2024-09-17
Fixing removing white space identifiers 

ajaysubra:
# Description

When identifiers were just a bunch of whitespace characters, we were still making a network call and getting 400'ed. This fixes that by trimming white spaces and newline before checking if the identifiers is empty. 

# Check List

# Manual Test Plan

1. Add an empty identifier to the profile (white spaces would work too) - email, phone number or external id
2. Tap on create profile 
3. Notice that a request will be made (since we don't have dedup logic for profile requests) but the request will not contain the empty identifier that was added.

### DIFF
--- a/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
+++ b/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
@@ -10,7 +10,7 @@ import KlaviyoCore
 
 extension String {
     fileprivate func returnNilIfEmpty() -> String? {
-        isEmpty ? nil : self
+        trimmingCharacters(in: .whitespacesAndNewlines).isEmpty ? nil : self
     }
 }
 


evan-masseau (approved):
Minor, but I log a warning message whenever we make a modification to a value, such as whitespace removal

ajaysubra:
> Minor, but I log a warning message whenever we make a modification to a value, such as whitespace removal

We already do before we get here in the state management file.

ajaysubra:
> > > Minor, but I log a warning message whenever we make a modification to a value, such as whitespace removal
> > 
> > 
> > We already do before we get here in the state management file.
> 
> However that says we ignore the requested change and not trimming white spaces -
> 
> ```swift
> private func logDevWarning(for identifier: String) {
>     environment.emitDeveloperWarning("""
>     \(identifier) is either empty or same as what is already set earlier.
>     The SDK will ignore this change, please use resetProfile for
>     resetting profile identifiers
>     """)
> }
> ```

Kinda true that we do ignore the identifier that is empty so I think what we have is sufficient without getting into too much detail. LMK what you think.

ndurell (approved):

ab1470 (approved):

PR #204, opened 2024-09-19
Using plist values for sdk name and version

ajaysubra:
# Description

Looking for a plist resource that our react native SDK adds to the host app's bundle. If this plist is found then use the sdk name and version from that plist else use the name and version hardcoded within this SDK.

This is how the plist looks, 

```
<?xml version="1.0" encoding="UTF-8"?>
<!--
   config.plist
   Pods

   Created by Ajay Subramanya on 9/18/24.
   Copyright (c) 2024 Klaviyo. All rights reserved.
-->
<plist version="1.0">
<dict>
    <key>react_native_sdk_name</key>
    <string>react_native</string>
    <key>react_native_sdk_version</key>
    <string>9.8.8</string>
</dict>
</plist>
```

This plist is loaded to the host app's bundle using the react native SDK podspec like below, 

```
require "json"

package = JSON.parse(File.read(File.join(__dir__, "package.json")))

Pod::Spec.new do |s|
  s.name         = package["name"]
  s.version      = package["version"]
  s.summary      = package["description"]
  s.homepage     = package["homepage"]
  s.license      = package["license"]
  s.authors      = package["author"]

  s.platforms    = { :ios => "13.0" }
  s.source       = { :git => "https://github.com/klaviyo/klaviyo-react-native-sdk.git", :tag => "#{s.version}" }
  s.source_files = "ios/**/*.{h,m,mm,swift}"
  s.resources    = ["ios/klaviyo-sdk-configuration.plist"]

  s.pod_target_xcconfig = { "DEFINES_MODULE" => "YES" }

  s.dependency "React-Core"
  s.dependency "KlaviyoSwift", "3.2.0"
end

```

notice the `s.resources    = ["ios/klaviyo-sdk-configuration.plist"]` is what copies this over to the host app's bundle. 

The reason I think this method is better than the private API method is because, we don't have to worry about making sure that the values are overridden at any given point (initializing from native SDK vs react native etc). With this approach, we know that as long as the plist is correctly bundled on react native we'll always have the right value when the `SDKName` and `SDKVersion` variables are accessed from `KlaviyoEnvironment`.

Thanks to @evan-masseau for this idea.

# Manual Test Plan

1. Using the iOS test app, made sure that the hardcoded version of the SDK is being used when making event and token calls.
2. Using the RN test app validated that the values in the plist are being used when making the same network calls. 
3. Alternate and quicker way to test this w/o going through RN is placing the above plist within the iOS test app and naming it `klaviyo-sdk-configuration.plist`

### DIFF
--- a/Sources/KlaviyoCore/KlaviyoEnvironment.swift
+++ b/Sources/KlaviyoCore/KlaviyoEnvironment.swift
@@ -119,7 +119,7 @@ public struct KlaviyoEnvironment {
     public var sdkVersion: () -> String
 
     private static let rnSDKConfig: [String: AnyObject] = {
-        loadPlist(named: "klaviyo-sdk-configuration") ?? [:]
+        loadPlist(named: "klaviyo-react-native-sdk-configuration") ?? [:]
     }()
 
     private static func getSDKName() -> String {
--- a/Sources/KlaviyoCore/KlaviyoEnvironment.swift
+++ b/Sources/KlaviyoCore/KlaviyoEnvironment.swift
@@ -118,6 +118,24 @@ public struct KlaviyoEnvironment {
     public var sdkName: () -> String
     public var sdkVersion: () -> String
 
+    private static let rnSDKConfig: [String: AnyObject] = {
+        loadPlist(named: "klaviyo-sdk-configuration") ?? [:]
+    }()
+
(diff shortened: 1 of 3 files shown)


evan-masseau (approved):

ab1470 (approved):
love this!

PR #206, opened 2024-09-24
Fixed retry not working

ajaysubra:
# Description

There is a logical error in the `if` condition. I am using `if httpResponse.statusCode == 429, httpResponse.statusCode == 503` which checks if `httpResponse.statusCode` is both `429` and `503` at the same time, which is not possible. Instead, I should use an `||` to check if `httpResponse.statusCode` is either `429` or `503`.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
+++ b/Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
@@ -40,7 +40,7 @@ public struct KlaviyoAPI {
             return .failure(.missingOrInvalidResponse(response))
         }
 
-        if httpResponse.statusCode == 429, httpResponse.statusCode == 503 {
+        if httpResponse.statusCode == 429 || httpResponse.statusCode == 503 {
             let exponentialBackOff = Int(pow(2.0, Double(attemptNumber)))
             var nextBackoff: Int = exponentialBackOff
             if let retryAfter = httpResponse.value(forHTTPHeaderField: "Retry-After") {


ab1470 (approved):

ndurell (approved):

ndurell:
good catch!

Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
```diff
@@ -40,7 +40,7 @@ public struct KlaviyoAPI {
             return .failure(.missingOrInvalidResponse(response))
         }
 
-        if httpResponse.statusCode == 429, httpResponse.statusCode == 503 {
+        if httpResponse.statusCode == 429 || httpResponse.statusCode == 503 {
```
ndurell, line 43: Hmm we didn't a have that failed?

ajaysubra: Are you asking why we did have a failing test?

ajaysubra: If yes, you are right, some unit test should have caught this. Cut a ticket to add some unit tests.

ndurell: Yah I'm surprised I thought I added one but maybe it didn't test things enough.

ajaysubra: nah I should have tested when I moved 429 and added 503. Have limited time this week so cut a ticket but hopefully we can add it soon.

PR #211, opened 2024-09-26
Updated KlaviyoCore to 4.0.0 and pointing to the just released tag

ajaysubra:
# Description

The new private pod `KlaviyoCore` now is versioned at 4.0.0 for consistency with the other pods and also to match the major version tag just released to avoid any confusion. Since we just have one repo that we release we only have one tag for all three pods so it's easier matching the versions of the pods. 

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Examples/KlaviyoSwiftExamples/CocoapodsExample/Podfile.lock
+++ b/Examples/KlaviyoSwiftExamples/CocoapodsExample/Podfile.lock
@@ -1,10 +1,10 @@
 PODS:
   - AnyCodable-FlightSchool (0.6.7)
-  - KlaviyoCore (0.1.0):
+  - KlaviyoCore (4.0.0):
     - AnyCodable-FlightSchool
-  - KlaviyoSwift (3.2.0):
+  - KlaviyoSwift (4.0.0):
     - AnyCodable-FlightSchool
-    - KlaviyoCore (~> 0.1.0)
+    - KlaviyoCore (~> 4.0.0)
   - KlaviyoSwiftExtension (3.0.1)
 
 DEPENDENCIES:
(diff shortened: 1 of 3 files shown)


evan-masseau (approved):
0.1.0 -> 4.0.0 what a glow up

Examples/KlaviyoSwiftExamples/CocoapodsExample/Podfile.lock
```diff
@@ -1,10 +1,10 @@
 PODS:
   - AnyCodable-FlightSchool (0.6.7)
-  - KlaviyoCore (0.1.0):
+  - KlaviyoCore (4.0.0):
```
ndurell, line 3: Do we want to recommend people include klaviyocore, it's not a transitive depedency?

ajaysubra: No folks don't need to worry about it, it is a transitive dependency. By pulling in `KlaviyoSwift` you are also pulling in `KlaviyoCore` without having to explicitly mention it in your pod file. Updated the example app from demo [here](https://github.com/klaviyo/klaviyo-swift-sdk/blob/master/Examples/KlaviyoSwiftExamples/CocoapodsExample/Podfile).

ndurell: oh my bad this is the lock!

ajaysubra: Si.

ab1470 (approved):

PR #216, opened 2024-10-03
Update swift.yml

ndurell:
# Description

# Check List

# Manual Test Plan

1.

# Supporting Materials


ab1470 (approved):

ab1470:
After testing locally, I figured out that `testEventPayload()` fails on iOS 18.0+. It passes on anything lower than 18.

PR #220, opened 2024-10-15
removed cleanup methods

ab1470:
Creating a separate branch from `ndurell-patch-1` to try and narrow down exactly what was causing CI failures


ab1470:
This PR removes the "cleanup" code from `APIRequestErrorHandlingTests`, as it was not the cause of the CI failures

ndurell (approved):

PR #221, opened 2024-10-16
bump cocoapods to 1.15.2

ab1470:
This PR bumps the cocoapods version to 1.15.2


njparadis (commented):
I think you might want to cherry pick only the commit with the version bump; also wondering if it's necessary to re-generate the `gem.lock` file in case this change affects  the dependencies cocoapods uses?

ab1470:
> I think you might want to cherry pick only the commit with the version bump

This is resolved already by just updating the branch on master. I rebased this branch off of your test store fix earlier today, anticipating that you'd be merging it into master shortly

> also wondering if it's necessary to re-generate the gemfile.lock file

Looking into this now

ab1470:
@njparadis updated the gemfile.lock, thanks

ndurell (approved):

PR #225, opened 2024-10-18
move all work to main thread with serial executor

ndurell:
# Description

This backports some stuff from the new test store from TCA.
This should move all store tests effects to the main thread.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Tests/KlaviyoSwiftTests/Vendor/ComposableArchitecture/TestStore.swift
+++ b/Tests/KlaviyoSwiftTests/Vendor/ComposableArchitecture/TestStore.swift
@@ -34,6 +34,7 @@ import CustomDump
 import Foundation
 import CasePaths
 import XCTestDynamicOverlay
+import Dispatch
 
 /// A testable runtime for a reducer.
 ///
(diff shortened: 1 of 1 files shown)


Tests/KlaviyoSwiftTests/Vendor/ComposableArchitecture/TestStore.swift
```diff
@@ -1241,13 +1262,13 @@ extension TestStore where ScopedState: Equatable, Action: Equatable {
   ///     expected.
   @MainActor
   @_disfavoredOverload
-  func receive(
+    func receive(
```
ndurell, line 1265: hmm will fix this

Tests/KlaviyoSwiftTests/Vendor/ComposableArchitecture/TestStore.swift
```diff
@@ -831,7 +846,13 @@ extension TestStore where ScopedState: Equatable {
     let previousState = self.reducer.state
     let task = self.store
       .send(.init(origin: .send(self.fromScopedAction(action)), file: file, line: line))
-    await self.reducer.effectDidSubscribe.stream.first(where: { _ in true })
+
+    if uncheckedUseMainSerialExecutor {
+      await Task.yield()
+    } else {
+      await self.reducer.effectDidSubscribe.stream.first(where: { _ in true })
```
ndurell, line 853: This is different in TCA - might backport that change too..

Tests/KlaviyoSwiftTests/Vendor/ComposableArchitecture/TestStore.swift
ab1470: Nit: for consistency, could we re-indent whole file with 4 spaces instead of 2? Select all (Command-A), then re-indent (Control-i)

ab1470 (approved):

Tests/KlaviyoSwiftTests/Vendor/ComposableArchitecture/TestStore.swift
ndurell: Yah I thought we had that set somewhere in the linting but maybe never enforced anything...

ndurell: That might make the diff hard to read.

ab1470: ahh good point

PR #233, opened 2024-10-30
Minor update to user agent format to reflect cross-platform SDK's name override

evan-masseau:
We want to be able to see react native or flutter in nginx logs, which can only see user agent.
We'll still be able differentiate android and iOS by the OS part of the user agent string.
See related [android PR](https://github.com/klaviyo/klaviyo-android-sdk/pull/195)

# Description

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Tests/KlaviyoCoreTests/__Snapshots__/NetworkSessionTests/testCreateEmphemeralSesionHeaders.1.txt
+++ b/Tests/KlaviyoCoreTests/__Snapshots__/NetworkSessionTests/testCreateEmphemeralSesionHeaders.1.txt
@@ -8,7 +8,7 @@
         - "deflate"
     ▿ (2 elements)
       - key: "User-Agent"
-      - value: "FooApp/1.2.3 (com.klaviyo.fooapp; build:1; iOS 1.1.1) klaviyo-ios/4.0.0"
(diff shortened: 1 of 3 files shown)


ndurell (approved):
might have some test failures. Might not if we don't have any user agent test with a different sdkName.

evan-masseau:
💯 Tests doing what tests are meant to do

Tests/KlaviyoCoreTests/__Snapshots__/NetworkSessionTests/testCreateEmphemeralSesionHeaders.1.txt
```diff
@@ -8,7 +8,7 @@
         - "deflate"
     ▿ (2 elements)
       - key: "User-Agent"
-      - value: "FooApp/1.2.3 (com.klaviyo.fooapp; build:1; iOS 1.1.1) klaviyo-ios/4.0.0"
```
evan-masseau, line 11: @ndurell for consistency with the sdk name, i did have to update the snapshot, resetting your review to make sure you're good with that. AFAIK the user agent isn't critical for anything so I think this is OK

ajaysubra (approved):
nice

ab1470 (approved):

PR #234, opened 2024-10-30
[CHNL-13250] dismiss webview through form actions

ab1470:
# Description

This PR adds a `webViewDidClose()` delegate method to detect when the form emits a `window.close()` javascript event. The method will call the ViewModel's `dismiss()` method, which will be handled after we've wired up the SDK to display the IAM view in the host app.

# Check List

# Supporting Materials

Here's a video showing a message being printed to the Xcode console when the user taps a "Close" button:

https://github.com/user-attachments/assets/23e71442-3a42-403a-8fff-203a025b3b51


Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewController.swift
```diff
@@ -91,6 +91,10 @@ class KlaviyoWebViewController: UIViewController, WKUIDelegate {
         }
     }
 
+    func webViewDidClose(_ webView: WKWebView) {
```
ndurell, line 94: So this is called by js?

Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebWrapperViewController.swift
```diff
@@ -0,0 +1,169 @@
...
+        loadSubviews()
+    }
+
+    func loadSubviews() {
+        if let blurEffectView {
+            view.addSubview(blurEffectView)
+            blurEffectView.pin(to: view)
+        }
+
+        if let tintView {
```
ndurell, line 98: Is this the new way to check for nil?

ndurell (approved):

Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebWrapperViewController.swift
ab1470: yeah, no need to do `if let tintView = tintView` anymore

Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewController.swift
ab1470: out of the box, WKWebView will listen for a `window.close()` JS event. We can respond to the close event by implementing this delegate method

ab1470:
Closing this as we've decided on a different approach for MVP. We can revisit this at a future date, if necessary.

PR #235, opened 2024-11-12
[CHNL-12942] Create a hook to reset badge counts

belleklaviyo:
# Description

This creates a new API method devs can call to reset the badge count to 0. It additionally sets the `badgeCount` value in a shared User Defaults to 0. That shared User Defaults will be a presumably set up following README updates to support badge incrementing. In the scenario the dev does not set the caching resource up correctly, this method will still set the badge to 0.

# Check List

# Manual Test Plan

1. In an app with Klaviyo SDK initialized, call `KlaviyoSDK().resetBadgeCount(defaults: appGroupSource)` where `appGroupSource` is the identifier of the app group between the Service Extension and App (per what would be set up in updated README)
2. Send some notification from Push Notification Console to populate the badge count (i.e. `badge = 3` in payload)
3. Observe when SDK call is triggered (i.e. a button or scene phase) the badge count is reset to 0

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -82,6 +82,11 @@ public struct KlaviyoSDK {
     public func resetProfile() {
         dispatchOnMainThread(action: .resetProfile)
     }
+    
+    public func resetBadgeCount(defaults: String) {
+        dispatchOnMainThread(action: .resetBadgeCount(defaults))
+        UIApplication.shared.applicationIconBadgeNumber = 0
+    }
 
     /// Set the current user's email.
     /// - Parameter email: a string contining the users email.
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -49,6 +49,9 @@ enum KlaviyoAction: Equatable {
 
     /// call this to sync the user's local push notification authorization setting with the user's profile on the Klaviyo back-end.
     case setPushEnablement(PushEnablement)
+    
+    /// call to reset the app badge count to 0 as well as update the stored value in the specified User Defaults suite
+    case resetBadgeCount(String)
 
     /// called when the user wants to reset the existing profile from state
     case resetProfile
@@ -100,7 +103,7 @@ enum KlaviyoAction: Equatable {
         case .setEmail, .setPhoneNumber, .setExternalId, .setPushToken, .setPushEnablement, .enqueueProfile, .setProfileProperty, .resetProfile, .resetStateAndDequeue, .enqueueEvent:
             return true
 
-        case .initialize, .completeInitialization, .deQueueCompletedResults, .networkConnectivityChanged, .flushQueue, .sendRequest, .stop, .start, .cancelInFlightRequests, .requestFailed:
+        case .initialize, .completeInitialization, .deQueueCompletedResults, .networkConnectivityChanged, .flushQueue, .sendRequest, .stop, .start, .cancelInFlightRequests, .requestFailed, .resetBadgeCount:
             return false
         }
     }
@@ -479,6 +482,12 @@ struct KlaviyoReducer: ReducerProtocol {
 
             return .none
 
+        case let .resetBadgeCount(defaults):
+            if let userDefaults = UserDefaults(suiteName: defaults) {
+                userDefaults.set(0, forKey: "badgeCount")
+            }
+            return .none
+            
         case .resetProfile:
             guard case .initialized = state.initalizationState
             else {


jasper-hsu-kyvo (commented):
Looks good to me, however adding Andrew to the review because he knows way more about the SDKs than I do (so won't approve sorry) - also looks like you need to rerun pre-commit though and have a formatting issue.

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -82,6 +82,11 @@ public struct KlaviyoSDK {
     public func resetProfile() {
         dispatchOnMainThread(action: .resetProfile)
     }
+    
+    public func resetBadgeCount(defaults: String) {
+        dispatchOnMainThread(action: .resetBadgeCount(defaults))
+        UIApplication.shared.applicationIconBadgeNumber = 0
```
ab1470, line 88: two comments:
1) I think this needs to be handled within the `.resetBadgeCount(defaults)` case in the `KlaviyoReducer.reduce` method
2) setting the badge number using `applicationIconBadgeNumber` has [been deprecated in iOS 17](https://developer.apple.com/documentation/uikit/uiapplication/1622918-applicationiconbadgenumber). We will need to handle the pre-iOS 16 case though, so we should use `if #available` to switch between the old API and the new one. This should work:
```swift
if #available(iOS 16.0, *) {
    UNUserNotificationCenter.current().setBadgeCount(0)
} else {
    UIApplication.shared.applicationIconBadgeNumber = 0
}
```
note that you'll need to add both `import UIKit` and `import UserNotifications` for this code to work

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -82,6 +82,11 @@ public struct KlaviyoSDK {
     public func resetProfile() {
         dispatchOnMainThread(action: .resetProfile)
     }
+    
+    public func resetBadgeCount(defaults: String) {
```
ab1470, line 86:
```suggestion
    public func resetBadgeCount(defaults: UserDefaults) {
```
it would be safer to set the parameter type as `UserDefaults` so that the user can pass in an actual `UserDefaults` object rather than a typo-prone `String`. Then in your `KlaviyoReducer.reduce` method, your `resetBadgeCount` case would look like this:
```swift
case let .resetBadgeCount(defaults):
    defaults.set(0, forKey: "badgeCount")
    return .none
```

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -82,6 +82,11 @@ public struct KlaviyoSDK {
     public func resetProfile() {
         dispatchOnMainThread(action: .resetProfile)
     }
+
```
ab1470, line 85:
```suggestion
    /// Resets the badge number on the application icon to 0.
    /// - Parameter defaults: The `UserDefaults` suite where the badge number is persisted.
```
We should try to add documentation whenever possible, especially for client-facing code

ab1470 (changes requested):
Just wondering, why do we need to persist the current badge number to `UserDefaults`?

ab1470:
Also, I think we'll want to get have some unit tests for these changes

Sources/KlaviyoSwift/Klaviyo.swift
belleklaviyo: yeah decided this will actually be moved into a plist value (that will be documented in the README, separate ticket) so this parameter will no longer be necessary

belleklaviyo:
> Just wondering, why do we need to persist the current badge number to `UserDefaults`?

this is necessary for how we are going to keep track and increment the badge count, details [here](https://klaviyo.atlassian.net/wiki/x/kwBFEQE) but still some things up in the air, might convert to a draft PR in the meantime

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -288,7 +293,11 @@ struct KlaviyoReducer: ReducerProtocol {
                 return .none
             }
 
+            let autoclearing = Bundle.main.object(forInfoDictionaryKey: "Klaviyo_badge_autoclearing") as? Bool ?? true
```
ndurell, line 296: We should move this to the environment to read from the bundle.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -479,6 +488,17 @@ struct KlaviyoReducer: ReducerProtocol {
 
             return .none
 
+        case let .setBadgeCount:
+            if let userDefaults = UserDefaults(suiteName: Bundle.main.object(forInfoDictionaryKey: "Klaviyo_App_Group") as? String) {
```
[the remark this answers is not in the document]
ajaysubra: @belleklaviyo were you planning on using the integer that was passed in?

@ndurell the idea was that this method acts as a pseudo-swizzled method for developers to set badge count with instead of using `UNUserNotificationCenter.current().setBadgeCount` and us trying to swizzle it. Of course it also acts as a way to reset badge counts.

ajaysubra (approved):
LGTM, minor comments from Noah and I.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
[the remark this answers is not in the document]
ndurell: Maybe should separate to a separate action but either should be moved to an effect.

Examples/KlaviyoSwiftExamples/Shared/Info.plist
```diff
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>Klaviyo_badge_autoclearing</key>
+	<false/>
+	<key>Klaviyo_App_Group</key>
+	<string></string>
```
ab1470, line 8: is this value supposed to be empty?

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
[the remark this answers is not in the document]
belleklaviyo: yeah meant to pass in the int, was a typo -- fixed

Examples/KlaviyoSwiftExamples/Shared/Info.plist
belleklaviyo: kinda? it needs to be set up by the dev, not sure if we should put a placeholder value?

ab1470: since it's an example app, I think it should be fully functional out of the box. Will the example app function as expected if this value is blank? If not, I think we should provide a value here. 

It looks like the naming convention for app groups is `group.<bundle identifier>`. So I think this value can be `<string>group.$(PRODUCT_BUNDLE_IDENTIFIER)</string>`

Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -150,7 +151,11 @@ public struct KlaviyoEnvironment {
             let notificationSettings = await UNUserNotificationCenter.current().notificationSettings()
             return PushEnablement.create(from: notificationSettings.authorizationStatus)
         },
-        getBackgroundSetting: { .create(from: UIApplication.shared.backgroundRefreshStatus) },
+        getBackgroundSetting: { .create(from: UIApplication.shared.backgroundRefreshStatus)
+        },
```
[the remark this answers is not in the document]
ajaysubra: pre commit hook not formatting the file?

Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -150,7 +151,11 @@ public struct KlaviyoEnvironment {
...
+        getBadgeAutoClearingSetting: {
+            Bundle.main.object(forInfoDictionaryKey: "Klaviyo_badge_autoclearing") as? Bool ?? true
```
ajaysubra, line 157: if we are unable to get the plist value we assume that they want to auto clear? I think this is fine but might be good to callout in the readme.

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -83,6 +83,12 @@ public struct KlaviyoSDK {
         dispatchOnMainThread(action: .resetProfile)
     }
 
+    /// Sets the badge number on the application icon. Syncs with the persisted count
+    /// stored in the User Defaults suite set up with the App Group
```
ajaysubra, line 87: Should we mention something about how/why this value is used?

ajaysubra (approved):
Minor nit picks else LGTM.

PR #237, opened 2024-11-16
[CHNL-14651] refactor network request

ab1470:
# Description

This PR refactors `KlaviyoRequest` and `KlaviyoEndpoint` so that the `KlaviyoEndpoint` enum now contains the values for `path` and `body`. The `KlaviyoRequest` struct now has no knowledge of individual endpoint details.

This work will make the `KlaviyoRequest` more flexible, and will enable us to add additional HTTP request methods other than "POST". I will make use of these changes in an upcoming PR. 

# Check List

### DIFF
--- a/Sources/KlaviyoCore/Networking/KlaviyoEndpoint.swift
+++ b/Sources/KlaviyoCore/Networking/KlaviyoEndpoint.swift
@@ -9,14 +9,7 @@
 import AnyCodable
 import Foundation
 
-protocol Endpoint: Equatable, Codable {
-    var httpMethod: RequestMethod { get }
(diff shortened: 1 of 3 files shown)


ndurell (approved):
Not sure if we need the endpoint protocol but I think the rest of it makes sense! I assume we are adding a get request maybe?

ajaysubra (approved):

ab1470:
> Not sure if we need the endpoint protocol but I think the rest of it makes sense! I assume we are adding a get request maybe?

Removed the protocol. Easy enough to add back in later if it would be useful.

PR #239, opened 2024-11-18
[CHNL-14651] add Klaviyo form preview

ab1470:
# Description

This PR adds some changes that we'll use for testing forms using Xcode previews. Changes include the following:

1. I created a "dummy" `PreviewGridViewController` as a sample of what a gridded app might look like underneath a Klaviyo form. It's an interactive view (you can scroll and change tabs) so we can test interactivity before and after the form is displayed. This is wrapped in an `#if DEBUG` flag so it can't be used in production code.
Screenshot 2024-11-18 at 12 26 49

2. I added a `klaviyo.html` file that will display a form when loaded into a `KlaviyoWebViewController`, and added an Xcode preview for this.
Screenshot 2024-11-18 at 12 27 42

3. I added a `jstest.html` file that can be used for testing the javascript bridge between the native layer and the web layer. This and all relevant files are added under a directory called "Development Assets", following a [convention established by Apple](https://www.avanderlee.com/xcode/development-assets-preview-catalog/) for specifying files that are only used for Xcode previews.

https://github.com/user-attachments/assets/58d0deed-1d26-4c04-abb8-0aee94f4061e

4. I made the background for the `KlaviyoWebViewController` transparent, and made the WebView ignore the safe area so it extends to all edges of the screen

# Check List

### DIFF
--- a/Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
+++ b/Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
@@ -1,0 +1,74 @@
+//
+//  JSTestWebViewModel.swift
+//  klaviyo-swift-sdk
+//
+//  Created by Andrew Balmer on 11/18/24.
+//
+
+#if DEBUG
+import Combine
+import Foundation
+import WebKit
+
+class JSTestWebViewModel: KlaviyoWebViewModeling {
+    let url: URL
+    let loadScripts: [String: WKUserScript]?
+
+    /// Publishes scripts for the `WKWebView` to execute.
+    private var continuation: AsyncStream<(script: String, callback: ((Result<Any?, Error>) -> Void)?)>.Continuation?
+    lazy var scriptStream: AsyncStream<(script: String, callback: ((Result<Any?, Error>) -> Void)?)> = AsyncStream { [weak self] continuation in
+        self?.continuation = continuation
+    }
+
+    init(url: URL) {
+        self.url = url
+        loadScripts = JSTestWebViewModel.initializeLoadScripts()
+    }
+
+    private static func initializeLoadScripts() -> [String: WKUserScript] {
+        var scripts: [String: WKUserScript] = [:]
+
+        if let toggleHandlerScript = try? FileIO.getFileContents(path: "toggleHandler", type: "js") {
+            let script = WKUserScript(source: toggleHandlerScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
+            scripts["toggleMessageHandler"] = script
+        }
+
+        return scripts
+    }
+
+    // MARK: handle WKWebView events
+
+    func handleNavigationEvent(_ event: WKNavigationEvent) {
+        // TODO: handle navigation events
+    }
+
+    func handleScriptMessage(_ message: WKScriptMessage) {
+        if message.name == "toggleMessageHandler" {
+            guard let dict = message.body as? [String: AnyObject] else {
+                return
+            }
+
+            guard let toggleEnabled = dict["toggleEnabled"] as? Bool else {
+                return
+            }
+
+            print("toggle enabled: \(toggleEnabled)")
+
+            let newStatus = toggleEnabled ? "Toggle is on" : "Toggle is off"
+
+            let script = "document.getElementById('toggle-status').innerText = \"\(newStatus)\""
+
+            continuation?.yield((script, { result in
+                switch result {
+                case let .success(content):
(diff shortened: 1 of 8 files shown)


ab1470:
FYI, I tried to use compiler flags (`#if DEBUG`) to ensure that all files within the `Development Assets` directory are only accessible within debug builds and from within the KlaviyoUI package, but I was unable to get it to work. This means that with this PR, the jstest.html and toggleHandler.js files can be accessed by the host app that consumes our SDK. I don't think this is a big deal because the host app would have no use for these files, but if we really want to exclude them we can delete the two `Development Assets` entries from the KlaviyoUI resources array in the package manifest before releasing the app.

The Swift files that are added in this PR are wrapped in `#if DEBUG` flags, so the code cannot be used in production. Additionally, they are internal to the KlaviyoUI package, so the code can't be accessed from the host app.

(cc: @ndurell)

ndurell:
> FYI, I tried to use compiler flags (`#if DEBUG`) to ensure that all files within the `Development Assets` directory are only accessible within debug builds and from within the KlaviyoUI package, but I was unable to get it to work. This means that with this PR, the jstest.html and toggleHandler.js files can be accessed by the host app that consumes our SDK. I don't think this is a big deal because the host app would have no use for these files, but if we really want to exclude them we can delete the two `Development Assets` entries from the KlaviyoUI resources array in the package manifest before releasing the app.
> 
> The Swift files that are added in this PR are wrapped in `#if DEBUG` flags, so the code cannot be used in production. Additionally, they are internal to the KlaviyoUI package, so the code can't be accessed from the host app.
> 
> (cc: @ndurell)

Yah it's probably a drop in the bucket compared to other stuff so no worries. Mostly just wanted to see if it was possible to exclude/remove but still use for testing.

Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
```diff
@@ -0,0 +1,74 @@
...
+            scripts["toggleMessageHandler"] = script
+        }
+
+        return scripts
+    }
+
+    // MARK: handle WKWebView events
+
+    func handleNavigationEvent(_ event: WKNavigationEvent) {
+        // TODO: handle navigation events
```
ajaysubra, line 42: what navigation events?

Sources/KlaviyoUI/KlaviyoWebView/Development Assets/PreviewGridViewController.swift
```diff
@@ -0,0 +1,74 @@
...
+        let itemSize = NSCollectionLayoutSize(widthDimension: .fractionalWidth(0.5), heightDimension: .fractionalHeight(1.0))
+        let item = NSCollectionLayoutItem(layoutSize: itemSize)
+        let groupSize = NSCollectionLayoutSize(widthDimension: .fractionalWidth(1.0), heightDimension: .fractionalWidth(0.4))
+        let group = NSCollectionLayoutGroup.horizontal(layoutSize: groupSize, subitems: [item])
+        group.interItemSpacing = .flexible(12)
+        let section = NSCollectionLayoutSection(group: group)
+        section.interGroupSpacing = 12
+        section.contentInsets = .init(top: 0, leading: 16, bottom: 0, trailing: 16)
+
+        let layout = UICollectionViewCompositionalLayout(section: section)
```
ajaysubra, line 28: it's been a minute since I used this!

ajaysubra (approved):
Thanks for creating this nice little util

Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
ab1470: the ViewController currently passes navigation events to the ViewModel (see [here](https://github.com/klaviyo/klaviyo-swift-sdk/blob/cb8185901e00bb1c2997a91010eee6361bed9b51/Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewController.swift#L108)). These events can be used to let us know when the website loads or if the website fails to load. 

We may not end up using this for the `JSTestWebViewModel`, but I can leave it here for now in case we decide to do some testing based on navigation events.

Sources/KlaviyoUI/KlaviyoWebView/Development Assets/PreviewGridViewController.swift
ab1470: haha me too. I've been in SwiftUI world for so long so UIKit is a throwback for me

ndurell (approved):

PR #242, opened 2024-12-04
added private `fetchForms()` method

ab1470:
# Description

This PR adds an `@_spi(KlaviyoPrivate)` `fetchForms()` method to the `KlaviyoSDK` struct. This method is intended for use in the test app, so we can tap a button to fetch the active forms and display the result in the request log. It's marked private because the host app should not manually call `fetchForms`, as this work will be handled with in the SDK.

Here's a preview of what the test app looks like with a button that's wired up to this new `fetchForms()` method:
Screenshot 2024-12-04 at 15 15 42

# Check List

### DIFF
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -184,3 +184,18 @@ public struct KlaviyoSDK {
         return false
     }
 }
+
+// MARK: - SPI Private methods
+
+// for internal use only
+
+@_spi(KlaviyoPrivate)
+extension KlaviyoSDK {
+    /// Fetch the active forms for the current user.
+    ///
+    /// - warning: For internal use only. The host app should not manually call this method, as
+    /// the logic for fetching and displaying forms will be handled internally within the SDK.
+    public func fetchForms() {
+        dispatchOnMainThread(action: .fetchForms)
+    }
+}


Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -184,3 +184,18 @@ public struct KlaviyoSDK {
         return false
     }
 }
+
+// MARK: - SPI Private methods
+
+// for internal use only
+
+@_spi(KlaviyoPrivate)
```
ajaysubra, line 192: should we add it to `Sources/KlaviyoCore/Networking/PrivateMethods.swift`?

ajaysubra (approved):

Sources/KlaviyoSwift/Klaviyo.swift
ab1470: I thought about that but ended up putting it here because that file is in the `KlaviyoCore/Networking` directory, and this method is neither within the KlaviyoCore library, nor is it networking code.

Personally I think it makes sense to keep it within Klaviyo.swift, as it's easier to find that way, but it's also clearly marked as private and it's separated out from the public API using an extension.

ndurell (approved):

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -184,3 +184,18 @@ public struct KlaviyoSDK {
...
+extension KlaviyoSDK {
+    /// Fetch the active forms for the current user.
+    ///
+    /// - warning: For internal use only. The host app should not manually call this method, as
+    /// the logic for fetching and displaying forms will be handled internally within the SDK.
+    public func fetchForms() {
```
dan-peluso, line 198: Just generally curious - so before was this just not exposed to the test app? I remember you had another pr adding the actual network logic, so that never had logic to call the fetch? Or is this an additional way

ab1470: the previous PR put in the architecture to make the `fetch` call internally within the SDK, but didn't expose a public `fetch` method. The host app that consumes our SDK will never actually manually fetch the forms, so I didn't want to expose a public `fetch` method. For testing purposes, I had a local branch that temporarily exposed the `fetch` method publicly so I could wire it up in the test app, but I never tried to merge that for the reason stated above. 

After thinking some more about it, I figured that the approach in this PR works well; expose the `fetch` method, but hide it behind `@_spi` and document that it's for internal use only. Although the host app can technically call this method if they really wanted to, it's difficult because they need to import our SDK in an unconventional way, and they'll also see the documentation indicating that it's only for internal use. Also, if we decide later that we just don't want to have any possibility of the host app calling this method, we can simply revert this PR in the future, especially after we're done doing dev/testing work on IAF.

belleklaviyo (approved):

PR #243, opened 2024-12-05
Fixed phone number typo

ajaysubra:
# Description

Fixed a typo in the phone number setter method comments

# Check List

# Manual Test Plan

1.

# Supporting Materials


dan-peluso (approved):
I wonder if there's some open-source CI typo checker we could fork for things like this. Probably overkill but might not be a bad idea given the repo is public documentation

ab1470 (approved):

ab1470:
> > I wonder if there's some open-source CI typo checker we could fork for things like this. Probably overkill but might not be a bad idea given the repo is public documentation
> 
> Yeah that's a good idea. We'll probably be working on some automation stuff some time next year and would be a good addition to that work. Also, at some point we should revisit these comments and make sure they work with DocC so that we can generate API reference docs CC @ab1470 @belleklaviyo

Yeah I'd love to get some DocC documentation going. Has that been discussed yet? Not sure how to slot that in with IAF work but I'd be happy to spearhead that task

PR #244, opened 2024-12-13
CHNL-14775: Add badge incrementing logic to extension

belleklaviyo:
# Description
SDK changes: [CHNL-14775](https://klaviyo.atlassian.net/browse/CHNL-14775)

Adding handling of new push payloads to service extension. Modifies the autoclearing set up, so it is on by default. Devs can disable it by adding a key in their Info.plist.

# Check List

# Manual Test Plan

**Pre-reqs:**
1. Set up App Group for Notification Service Extension and main app target with the same name (e.g. `group.com.klaviyo.SDKTestApp.shared`)
2. Add `Klaviyo_App_Group` to both the main app target's and the extension's Info.plists with the name of the App Group

**Default Autoclearing**
1. Send some plain notification from Push Notification Console to populate the badge count (i.e. badge = 3 in payload) with the app in the background and observe badge of 3
```
{
    "aps": {
        "alert": {
            "title": "title",
            "subtitle": "subtitle",
            "body": "body"
        },
        "badge": 3
    }
}
```
3. Open the app
4. Close the app and observe badge cleared to 0

**Disabling Autoclearing**
1. Add `disable_Klaviyo_badge_autoclearing` to the Info.plist of the main app target as a Boolean set to `YES`
2. Send some plain notification from Push Notification Console to populate the badge count (i.e. badge = 3 in payload) with the app in the background and observe badge of 3
3. Open the app
4. Close the app and observe badge of 3 still remains

**Push Notification Console Testing** 
Send payloads as documented [here](https://klaviyo.atlassian.net/wiki/x/yIBjFQE) to cover all possible paths of badge configurations:
- [ ] badge:off
- [ ] badge:on, config:increment_one, value:1
- [ ] badge:on, config:set_count, value:3 (or any valid int value)
- [ ] badge:on, config:set_property, value:5 (or any valid int value)

Note: only other non happy paths worth testing imo are where we are sending the `badge_config` but no `badge_value` due to poor rendering and if there were some new enum case added on the backend and the SDK is not updated:
- [ ] badge:on, config:increment_one, NO VALUE
- [ ] badge:on, config:set_count, NO VALUE
- [ ] badge:on, config:set_property, NO VALUE
- [ ] badge:on, config:ex_new_enum, NO VALUE
- [ ] badge:on, config:ex_new_enum, value:2 (or any valid int value)

**End to End Testing** 

# Supporting Materials

[CHNL-14775]: https://klaviyo.atlassian.net/browse/CHNL-14775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[CHNL-12941]: https://klaviyo.atlassian.net/browse/CHNL-12941?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,7 +297,7 @@ By default, Klaviyo SDK automatically clears all badges on app open. If you want
 
 #### Handling Other Badging Sources
 
-Klaviyo SDK handles Klaviyo pushes, but if you have other sources that change the badge count, use the `KlaviyoSDK().setBadgeCount(:)` method wherever you change the badge count to keep in sync with SDK count.
+Klaviyo SDK will automatically handle the badge count associated with Klaviyo pushes. If you need to manually update the badge count to account for other notification sources, use the `KlaviyoSDK().setBadgeCount(:)` method, which will update the badge count and keep it in sync with the Klaviyo SDK. This method should be used instead of (rather than in addition to) setting the badge count using `UNUserNotificationCenter` and/or `UIApplication` methods.
 
 ### Receiving Push Notifications
 
--- a/README.md
+++ b/README.md
@@ -288,8 +288,8 @@ Klaviyo supports custom badge counts when configuring your push notification in 
 2. Add an App Groups capability and click the plus in the new section to add a new App Group
 3. Pick a name based on the scheme `group.com.[MainTargetBundleId].[descriptor]`
 4. Select your Service Extension target, and add the same App Group with the same name
-5. In your app's `Info.plist`, add a new entry for `Klaviyo_App_Group` as a String with the App Group name
-6. In your Notification Service Extension's `Info.plist`, add a new entry for `Klaviyo_App_Group` as a String with the App Group name
+5. In your app's `Info.plist`, add a new entry for `klaviyo_app_group` as a String with the App Group name
+6. In your Notification Service Extension's `Info.plist`, add a new entry for `klaviyo_app_group` as a String with the App Group name
 
 #### Autoclearing Badge Count
 
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ Klaviyo supports custom badge counts when configuring your push notification in 
 
 #### Autoclearing Badge Count
 
-If you want to automatically clear the badge count of your app to 0 when the app is opened, in your app's `Info.plist`, add a new entry for `Klaviyo_badge_autoclearing` as a Boolean set to `YES`.
+By default, Klaviyo SDK automatically clears all badges on app open. If you want to disable this behavior, in your app's `Info.plist`, add a new entry for `disable_Klaviyo_badge_autoclearing` as a Boolean set to `YES`.
 
 #### Handling Other Badging Sources
 
--- a/README.md
+++ b/README.md
@@ -20,7 +20,6 @@
   - [Collecting Push Tokens](#collecting-push-tokens)
   - [Request Push Notification Permission](#request-push-notification-permission)
   - [Badge Count](#badge-count)
-    - [Set Up](#set-up)
     - [Autoclearing Badge Count](#autoclearing-badge-count)
     - [Handling Other Badging Sources](#handling-other-badging-sources)
   - [Receiving Push Notifications](#receiving-push-notifications)
@@ -47,7 +46,7 @@ Once integrated, your marketing team will be able to better understand your app 
 ## Installation
 
 1. Enable push notification capabilities in your Xcode project. The section "Enable the push notification capability" in this [Apple developer guide](https://developer.apple.com/documentation/usernotifications/registering_your_app_with_apns#2980170) provides detailed instructions.
-2. [Optional but recommended] If you intend to use [rich push notifications](#rich-push) or [custom badge counts](#custom-badge-count) add a [Notification Service Extension](https://developer.apple.com/documentation/usernotifications/unnotificationserviceextension) to your Xcode project. A notification service app extension ships as a separate bundle inside your iOS app. To add this extension to your app:
+2. If you intend to use [rich push notifications](#rich-push) or [custom badge counts](#custom-badge-count) add a [Notification Service Extension](https://developer.apple.com/documentation/usernotifications/unnotificationserviceextension) to your Xcode project. A Notification Service Extension ships as a separate bundle inside your iOS app. To add this extension to your app:
    - Select File > New > Target in Xcode.
    - Select the Notification Service Extension target from the iOS > Application extension section.
    - Click Next.
@@ -57,6 +56,15 @@ Once integrated, your marketing team will be able to better understand your app 
     > ⚠️ The deployment target of your notification service extension defaults to the latest iOS version.
              If this exceeds your app's minimum supported iOS version, push notifications may not display attached media on older devices.
              To avoid this, ensure the extension's minimum deployment target matches that of your app. ⚠️
+
+    Set up an App Group between your main app target and your Notification Service Extension.
+    - Select your main app target > Signing & Capabilities
+    - Select + Capability (make sure it is set to All not Debug or Release) > App Groups
+    - Create a new App Group based on the recommended naming scheme `group.com.[MainTargetBundleId].[descriptor]`
+    - In your app's `Info.plist`, add a new entry for `klaviyo_app_group` as a String with the App Group name
+    - Select your Notification Service Extension target > Signing & Capabilities
+    - Add an App Group with the same name as the main target's App Group
+    - In your Notification Service Extension's `Info.plist`, add a new entry for `klaviyo_app_group` as a String with the App Group name
 
 3. Based on which dependency manager you use, follow the instructions below to install the Klaviyo's dependencies.
 
--- a/README.md
+++ b/README.md
@@ -283,12 +283,13 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 #### Set Up
 Klaviyo supports custom badge counts when configuring your push notification in the editor dashboard. To set up your app, so the Klaviyo SDK properly handles them:
 
-(Pre-requisite: Set up the Push Notification Service Extension)
+(Pre-requisite: Set up the Notification Service Extension)
 1. In XCode, select your main app target, then go to Signing & Capabilities
 2. Add an App Groups capability and click the plus in the new section to add a new App Group
-3. Pick a name based on the scheme `group.[MainTargetBundleId].[descriptor]`
+3. Pick a name based on the scheme `group.com.[MainTargetBundleId].[descriptor]`
 4. Select your Service Extension target, and add the same App Group with the same name
 5. In your app's `Info.plist`, add a new entry for `Klaviyo_App_Group` as a String with the App Group name
+6. In your Notification Service Extension's `Info.plist`, add a new entry for `Klaviyo_App_Group` as a String with the App Group name
 
 #### Autoclearing Badge Count
 
--- a/README.md
+++ b/README.md
@@ -19,6 +19,10 @@
   - [Prerequisites](#prerequisites)
   - [Collecting Push Tokens](#collecting-push-tokens)
   - [Request Push Notification Permission](#request-push-notification-permission)
+  - [Badge Count](#badge-count)
+    - [Set Up](#set-up)
+    - [Autoclearing Badge Count](#autoclearing-badge-count)
+    - [Handling Other Badging Sources](#handling-other-badging-sources)
   - [Receiving Push Notifications](#receiving-push-notifications)
     - [Tracking Open Events](#tracking-open-events)
     - [Deep Linking](#deep-linking)
@@ -43,7 +47,7 @@ Once integrated, your marketing team will be able to better understand your app 
 ## Installation
 
 1. Enable push notification capabilities in your Xcode project. The section "Enable the push notification capability" in this [Apple developer guide](https://developer.apple.com/documentation/usernotifications/registering_your_app_with_apns#2980170) provides detailed instructions.
-2. [Optional but recommended] If you intend to use [rich push notifications](#rich-push) add a [Notification service extension](https://developer.apple.com/documentation/usernotifications/unnotificationserviceextension) to your xcode project. A notification service app extension ships as a separate bundle inside your iOS app. To add this extension to your app:
+2. [Optional but recommended] If you intend to use [rich push notifications](#rich-push) or [custom badge counts](#custom-badge-count) add a [Notification Service Extension](https://developer.apple.com/documentation/usernotifications/unnotificationserviceextension) to your Xcode project. A notification service app extension ships as a separate bundle inside your iOS app. To add this extension to your app:
    - Select File > New > Target in Xcode.
    - Select the Notification Service Extension target from the iOS > Application extension section.
    - Click Next.
@@ -91,7 +95,7 @@ Once integrated, your marketing team will be able to better understand your app 
       </details>
 
 4. Finally, in the `NotificationService.swift` file add the code for the two required delegates from [this](Examples/KlaviyoSwiftExamples/SPMExample/NotificationServiceExtension/NotificationService.swift) file.
-  This sample covers calling into Klaviyo so that we can download and attach the media to the push notification.
+  This sample covers calling into Klaviyo so that we can download and attach the media to the push notification as well as handle custom badge counts.
 
 ## Initialization
 The SDK must be initialized with the short alphanumeric [public API key](https://help.klaviyo.com/hc/en-us/articles/115005062267#difference-between-public-and-private-api-keys1)
@@ -274,6 +278,25 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
     return true
 }
 ```
+### Badge Count
+
+#### Set Up
+Klaviyo supports custom badge counts when configuring your push notification in the editor dashboard. To set up your app, so the Klaviyo SDK properly handles them:
+
+(Pre-requisite: Set up the Push Notification Service Extension)
+1. In XCode, select your main app target, then go to Signing & Capabilities
+2. Add an App Groups capability and click the plus in the new section to add a new App Group
+3. Pick a name based on the scheme `group.[MainTargetBundleId].[descriptor]`
+4. Select your Service Extension target, and add the same App Group with the same name
+5. In your app's `Info.plist`, add a new entry for `Klaviyo_App_Group` as a String with the App Group name
+
+#### Autoclearing Badge Count
+
+If you want to automatically clear the badge count of your app to 0 when the app is opened, in your app's `Info.plist`, add a new entry for `Klaviyo_badge_autoclearing` as a Boolean set to `YES`.
+
+#### Handling Other Badging Sources
+
+Klaviyo SDK handles Klaviyo pushes, but if you have other sources that change the badge count, use the `KlaviyoSDK().setBadgeCount(:)` method wherever you change the badge count to keep in sync with SDK count.
 
 ### Receiving Push Notifications
 
@@ -310,44 +333,6 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
         } else {
             completionHandler([.alert])
         }
-    }
-}
-```
-When tracking opened push notification, you can also decrement the badge count on the app icon by adding the following code to the `userNotificationCenter:didReceive:withCompletionHandler` method:
-
-```swift
-    func userNotificationCenter(
-        _ center: UNUserNotificationCenter,
-        didReceive response: UNNotificationResponse,
-        withCompletionHandler completionHandler: @escaping () -> Void) {
-        // decrement the badge count on the app icon
-        if #available(iOS 16.0, *) {
-            UNUserNotificationCenter.current().setBadgeCount(UIApplication.shared.applicationIconBadgeNumber - 1)
-        } else {
-            UIApplication.shared.applicationIconBadgeNumber -= 1
-        }
-
-        // If this notification is Klaviyo's notification we'll handle it
-        // else pass it on to the next push notification service to which it may belong
-        let handled = KlaviyoSDK().handle(notificationResponse: response, withCompletionHandler: completionHandler)
-        if !handled {
-            completionHandler()
-        }
-    }
-```
-
-Additionally, if you just want to reset the badge count to zero when the app is opened(note that this could be from
-the user just opening the app independent of the push message), you can add the following code to
-the `applicationDidBecomeActive` method in the app delegate:
-
-```swift
-
-func applicationDidBecomeActive(_ application: UIApplication) {
-    // reset the badge count on the app icon
-    if #available(iOS 16.0, *) {
-        UNUserNotificationCenter.current().setBadgeCount(0)
-    } else {
-        UIApplication.shared.applicationIconBadgeNumber = 0
     }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ Klaviyo supports custom badge counts when configuring your push notification in 
 
 #### Autoclearing Badge Count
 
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -274,6 +278,26 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
...
+#### Set Up
+Klaviyo supports custom badge counts when configuring your push notification in the editor dashboard. To set up your app, so the Klaviyo SDK properly handles them:
+
+(Pre-requisite: Set up the Notification Service Extension)
+1. In XCode, select your main app target, then go to Signing & Capabilities
+2. Add an App Groups capability and click the plus in the new section to add a new App Group
+3. Pick a name based on the scheme `group.com.[MainTargetBundleId].[descriptor]`
+4. Select your Service Extension target, and add the same App Group with the same name
+5. In your app's `Info.plist`, add a new entry for `Klaviyo_App_Group` as a String with the App Group name
+6. In your Notification Service Extension's `Info.plist`, add a new entry for `Klaviyo_App_Group` as a String with the App Group name
```
ajaysubra, line 292: why do we need these two?

README.md
```diff
@@ -274,6 +278,26 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
...
+
+#### Autoclearing Badge Count
```
ajaysubra, line 294: Did we say we are going to enable this by default but provide an option to turn it off? May be I'm mis remembering.

README.md
belleklaviyo: the extension is not able to read the main app target's Info.plist since it has its own bundle, so we need to define it in its own Info.plist as well

README.md
```diff
@@ -274,6 +278,26 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
...
+
+By default, Klaviyo SDK automatically clears all badges on app open. If you want to disable this behavior, in your app's `Info.plist`, add a new entry for `disable_Klaviyo_badge_autoclearing` as a Boolean set to `YES`.
```
ajaysubra, line 296:
```suggestion
By default, Klaviyo SDK automatically clears all badges on app open. If you want to disable this behavior, in your app's `Info.plist`, add a new entry for `klaviyo_badge_autoclearing` as a Boolean set to `NO`.
```

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -163,6 +163,7 @@ public struct KlaviyoSDK {
     ///   - deepLinkHandler: a completion handler that will be called when a notification contains a deep link.
     /// - Returns: true if the notificaiton originated from Klaviyo, false otherwise.
     public func handle(notificationResponse: UNNotificationResponse, withCompletionHandler completionHandler: @escaping () -> Void, deepLinkHandler: ((URL) -> Void)? = nil) -> Bool {
+        KlaviyoBadgeCountUtil.syncBadgeCount()
```
ndurell, line 166: Does this need to happen here?

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -585,6 +588,16 @@ extension Store where State == KlaviyoState, Action == KlaviyoAction {
         reducer: KlaviyoReducer())
 }
 
+enum KlaviyoBadgeCountUtil {
+    static func syncBadgeCount() {
+        DispatchQueue.main.async {
```
ndurell, line 593: This feels side effect-y to me.

belleklaviyo: felt safer this way since we are reading `UIApplication.shared.applicationIconBadgeNumber`

ndurell: Safer in what sense?

Sources/KlaviyoSwift/Klaviyo.swift
belleklaviyo: It's an attempt to keep our count in sync with any other (non Klaviyo) notifications that go through this handler

ndurell: Yah I guess I'm asking if it could be done in the environment and as part of the action handling maybe for open event?

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
belleklaviyo: thought calling UIApplication had to be called on main thread or is that an outdated/misremembered best practice

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -513,6 +519,16 @@ struct KlaviyoReducer: ReducerProtocol {
                 _ = klaviyoSwiftEnvironment.setBadgeCount(count)
             }
 
+        case .syncBadgeCount:
+            Task {
+                await MainActor.run {
+                    if let userDefaults = UserDefaults(suiteName: Bundle.main.object(forInfoDictionaryKey: "Klaviyo_App_Group") as? String) {
```
kennyklaviyo, line 525: should this be `klaviyo_badge_autoclearing` like in Sources/KlaviyoCore/KlaviyoEnvironment.swift?

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -24,6 +30,32 @@ public enum KlaviyoExtensionSDK {
         bestAttemptContent: UNMutableNotificationContent,
         contentHandler: @escaping (UNNotificationContent) -> Void,
         fallbackMediaType: String = "jpeg") {
+        // handle badge setting from the push notification payload
```
ajaysubra, line 33: Can this be broken down like, to avoid some duplication. Need to figure out what to do when you can read from user defaults. 


```swift
      if let badgeConfig = bestAttemptContent.userInfo["badge_config"] as? String {
          var newBadgeValue: Int?
          
          guard let userDefaults = getAppGroupUserDefaults else {
              return // not sure what to do here?
          }
          
          switch badgeConfig {
          case KlaviyoBadgeConfig.incrementOne.rawValue:
              let currentBadgeCount = userDefaults.integer(forKey: "badgeCount")
              newBadgeValue = currentBadgeCount + 1
              
          default:
              if let badgeValue = bestAttemptContent.userInfo["badge_value"] as? Int {
                  newBadgeValue = badgeValue
              }
          }
          userDefaults.set(newBadgeValue, forKey: "badgeCount")
          bestAttemptContent.badge = newBadgeValue as? NSNumber
      }


    static var getAppGroupUserDefaults: UserDefaults? {
        guard let appGroup = Bundle.main.object(forInfoDictionaryKey: "Klaviyo_App_Group") as? String else {
            return nil
        }
        return UserDefaults(suiteName: appGroup)
    }
```

belleklaviyo: I'm okay with just returning since if we're unable to read from UserDefaults it likely means the count would be going out of sync at that point. We already plan on having the `set_count` and `set_property` cases have their aps badge set to the `badge_value` so display wise they would still be okay, and we could likewise do this for any other future enum cases so they'd at least be displayed correctly.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
belleklaviyo: casing wise? or are you asking if it should be `klaviyo_badge_autoclearing` and not the app group

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
ajaysubra: That's a good point. I'm aligned.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
ajaysubra: @kennyklaviyo are you saying this should be in SDK state vs. UserDefaults?

kennyklaviyo: ooo sorry didnt realize it was a different string, should this one be lower cased too?

belleklaviyo: yeah it could be lower cased

kennyklaviyo: also..does this mean a developer can explicitly set something in `Klaviyo_App_Group` in their Info.plist just like for `klaviyo_badge_autoclearing` and mess with our values?

kennyklaviyo: ah gotcha that makes sense then, will this be a breaking change btw? as in if someone just bumped the sdk version on their app, it wouldnt be able to build without these new entries?

belleklaviyo: not breaking, if the app group is not defined in the Info.plist it just doesn't load the User Defaults

Sources/KlaviyoSwift/KlaviyoSwiftEnvironment.swift
```diff
@@ -31,7 +31,7 @@ struct KlaviyoSwiftEnvironment {
             stateChangePublisher: StateChangePublisher().publisher,
             setBadgeCount: { count in
                 Task {
-                    if let userDefaults = UserDefaults(suiteName: Bundle.main.object(forInfoDictionaryKey: "Klaviyo_App_Group") as? String) {
+                    if let userDefaults = UserDefaults(suiteName: Bundle.main.object(forInfoDictionaryKey: "klaviyo_app_group") as? String) {
```
kennyklaviyo, line 34: i see below theres a guard,
```
            guard let appGroup = Bundle.main.object(forInfoDictionaryKey: "klaviyo_app_group") as? String,
                  let userDefaults = UserDefaults(suiteName: appGroup) else {
                return
            }
```
should that also be done here in case we end up with `UserDefaults(suiteName: nil)`?

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -513,6 +519,16 @@ struct KlaviyoReducer: ReducerProtocol {
                 _ = klaviyoSwiftEnvironment.setBadgeCount(count)
             }
 
+        case .syncBadgeCount:
+            Task {
+                await MainActor.run {
+                    if let userDefaults = UserDefaults(suiteName: Bundle.main.object(forInfoDictionaryKey: "klaviyo_app_group") as? String) {
```
kennyklaviyo, line 525: same comment here about the guard statement

README.md
```diff
@@ -274,6 +278,26 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
...
+5. In your app's `Info.plist`, add a new entry for `klaviyo_app_group` as a String with the App Group name
+6. In your Notification Service Extension's `Info.plist`, add a new entry for `klaviyo_app_group` as a String with the App Group name
+
+#### Autoclearing Badge Count
+
+By default, Klaviyo SDK automatically clears all badges on app open. If you want to disable this behavior, in your app's `Info.plist`, add a new entry for `klaviyo_badge_autoclearing` as a Boolean set to `NO`. You can turn this on again by setting this to `YES`.
+
+#### Handling Other Badging Sources
+
+Klaviyo SDK handles Klaviyo pushes, but if you have other sources that change the badge count, use the `KlaviyoSDK().setBadgeCount(:)` method wherever you change the badge count to keep in sync with SDK count.
```
kennyklaviyo, line 300: nit: can we run this whole readme thru copilot to correct any grammatical issues/inconsistency in wording?

Sources/KlaviyoSwift/KlaviyoSwiftEnvironment.swift
kennyklaviyo: and if that is the case, is it worth pulling into a util function somewhere?

belleklaviyo: if let is basically the same as a guard, it handles it carefully, just personal preference on readability ig, can switch to the guards if thats preferable

README.md
ab1470, line 300: Does `KlaviyoSDK().setBadgeCount(:)` actually change the badge count, or does it just update the value in UserDefaults?

belleklaviyo: both, but it only changes it if it was able to first update it in UserDefaults

Sources/KlaviyoSwift/KlaviyoSwiftEnvironment.swift
kennyklaviyo: nice thats fine then i think, seems like guard returns in the indent whereas if let executes in the indent so its a matter of where you want nesting vs early returns

kennyklaviyo: since this is a large block in the execute, i would opt for an early return vs the other file where the execution is a single like, the if let reads better there

kennyklaviyo: but i dont feel strongly

README.md
ab1470: Ok. When I read what you wrote I was a little unclear about what exactly the `setBadgeCount` method would do, and whether it should be used in addition to or instead of the native methods for setting the badge count. Maybe something like what I suggest below would make it a little more clear?

```suggestion
Klaviyo SDK will automatically handle the badge count associated with Klaviyo pushes. If you need to manually update the badge count to account for other notification sources, use the `KlaviyoSDK().setBadgeCount(:)` method, which will update the badge count and keep it in sync with the Klaviyo SDK. This method should be used instead of (rather than in addition to) setting the badge count using `UNUserNotificationCenter` and/or `UIApplication` methods.
```

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -24,6 +30,28 @@ public enum KlaviyoExtensionSDK {
...
+            var newBadgeValue: Int?
+            switch badgeConfig {
+            case KlaviyoBadgeConfig.incrementOne.rawValue:
+                let currentBadgeCount = userDefaults.integer(forKey: "badgeCount")
+                newBadgeValue = currentBadgeCount + 1
+            default:
+                if let badgeValue = bestAttemptContent.userInfo["badge_value"] as? Int {
+                    newBadgeValue = badgeValue
+                }
+            }
```
ab1470, line 49:
```suggestion
        if let badgeConfigValue = bestAttemptContent.userInfo["badge_config"] as? String {
            guard let appGroup = Bundle.main.object(forInfoDictionaryKey: "klaviyo_app_group") as? String,
                  let userDefaults = UserDefaults(suiteName: appGroup) else {
                return
            }
            
            guard let badgeConfig = KlaviyoBadgeConfig(rawValue: badgeConfigValue) else { return }

            var newBadgeValue: Int?
            switch badgeConfig {
            case .incrementOne:
                let currentBadgeCount = userDefaults.integer(forKey: "badgeCount")
                newBadgeValue = currentBadgeCount + 1
            case .setCount, .setProperty:
                if let badgeValue = bestAttemptContent.userInfo["badge_value"] as? Int {
                    newBadgeValue = badgeValue
                }
            }
```

I think this is a cleaner way of handling the conversion from a raw value to an enum case. It also allows us to avoid the `default` case, which [in general we should try to avoid](https://www.swiftbysundell.com/tips/avoiding-default-cases-in-switch-statements/)

ajaysubra: Thanks @ab1470 I was looking for ways to avoid the raw value and this is neat. On the `default` case point, I agree in the case of apps but since we are a SDK, I wonder if there is still value in defining what the default behaviour should be in the case we add newer option to this on the backend?

ab1470: ah I see. The question is what behavior do we want in case a new, unknown-to-the-SDK value is supplied by the backend. With my suggested code above, it'll hit line 40, won't be able to convert the raw value to an enum case, and then return without doing anything because it's an unrecognized value. Do we want to do that? Or do we want to try to handle the unknown value anyway?

ab1470: So yeah the question is whether we're comfortable letting the `default` case handle any unknown future values. The advantage of making sure that we're always doing _something_ is also a risk. Suppose someday we add a backend config value like `"do_nothing"`, but the SDK doesn't know how to explicitly handle it so it gets handled by the `default` case and increments the badge anyway

ab1470: ok, I'll leave it up to you then, as you have more knowledge on badges stuff than I do. That being said, we can still add a little safety and readability by adding an `.unknown` case to the enum and creating a custom initializer:
```swift
private enum KlaviyoBadgeConfig {
    case incrementOne
    case setCount
    case setProperty
    case unknown(value: String)
    
    init(rawValue: String) {
        switch rawValue {
        case "increment_one": self = .incrementOne
        case "set_count": self = .setCount
        case "set_property": self = .setProperty
        default: self = .unknown(value: rawValue)
        }
    }
}
```

and then use it like this:
```swift
let badgeConfig = KlaviyoBadgeConfig(rawValue: badgeConfigValue)

var newBadgeValue: Int?

switch badgeConfig {
case .incrementOne:
    let currentBadgeCount = userDefaults.integer(forKey: "badgeCount")
    newBadgeValue = currentBadgeCount + 1
case .setCount, .setProperty, .unknown:
    if let badgeValue = bestAttemptContent.userInfo["badge_value"] as? Int {
        newBadgeValue = badgeValue
    }
```

README.md
```diff
@@ -274,6 +278,25 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
...
+
+#### Set Up
+Klaviyo supports custom badge counts when configuring your push notification in the editor dashboard. To set up your app, so the Klaviyo SDK properly handles them:
+
+(Pre-requisite: Set up the Push Notification Service Extension)
+1. In XCode, select your main app target, then go to Signing & Capabilities
+2. Add an App Groups capability and click the plus in the new section to add a new App Group
+3. Pick a name based on the scheme `group.[MainTargetBundleId].[descriptor]`
+4. Select your Service Extension target, and add the same App Group with the same name
+5. In your app's `Info.plist`, add a new entry for `Klaviyo_App_Group` as a String with the App Group name
```
mark-piana: As discussed in Slack, I think it might make sense to add this portion to the initial Installation section of the Readme since it's probably the most convenient time for a developer to set up App Groups.

README.md
```diff
@@ -274,6 +278,26 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
...
+1. In XCode, select your main app target, then go to Signing & Capabilities
+2. Add an App Groups capability and click the plus in the new section to add a new App Group
+3. Pick a name based on the scheme `group.com.[MainTargetBundleId].[descriptor]`
+4. Select your Service Extension target, and add the same App Group with the same name
+5. In your app's `Info.plist`, add a new entry for `klaviyo_app_group` as a String with the App Group name
+6. In your Notification Service Extension's `Info.plist`, add a new entry for `klaviyo_app_group` as a String with the App Group name
+
+#### Autoclearing Badge Count
+
+By default, Klaviyo SDK automatically clears all badges on app open. If you want to disable this behavior, in your app's `Info.plist`, add a new entry for `klaviyo_badge_autoclearing` as a Boolean set to `NO`. You can turn this on again by setting this to `YES`.
```
mark-piana, line 296:
```suggestion
By default, the Klaviyo SDK automatically clears the badge count on app open. If you want to disable this behavior, add a new entry for `klaviyo_badge_autoclearing` as a Boolean set to `NO` in your app's `Info.plist`. You can re-enable automatically clearing badges by setting this value to `YES`.
```

README.md
```diff
@@ -43,7 +47,7 @@ Once integrated, your marketing team will be able to better understand your app
 ## Installation
```
mark-piana, line 47: I think it would make sense to move the set up steps for App Groups into this section similar to how we have the notification service extension instructions.

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
belleklaviyo: based on [this](https://klaviyo.slack.com/archives/C080HQERMJP/p1734722960414899?thread_ts=1734721301.288619&cid=C080HQERMJP) thread we will do nothing for the .unknown case

README.md
```diff
@@ -274,6 +278,26 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
     return true
 }
 ```
+### Badge Count
+
+#### Set Up
+Klaviyo supports custom badge counts when configuring your push notification in the editor dashboard. To set up your app, so the Klaviyo SDK properly handles them:
```
mark-piana, line 292:
```suggestion
Klaviyo supports setting or incrementing the badge count when you send a push notification. For this functionality to work, you must set up the Notification Service Extension and an App Group as outlined under the Installation section.
```

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -24,6 +40,31 @@ public enum KlaviyoExtensionSDK {
         bestAttemptContent: UNMutableNotificationContent,
         contentHandler: @escaping (UNNotificationContent) -> Void,
         fallbackMediaType: String = "jpeg") {
+        // handle badge setting from the push notification payload
+        if let badgeConfigValue = bestAttemptContent.userInfo["badge_config"] as? String {
+            guard let appGroup = Bundle.main.object(forInfoDictionaryKey: "klaviyo_app_group") as? String,
+                  let userDefaults = UserDefaults(suiteName: appGroup) else {
+                return
```
kennyklaviyo, line 47: when we return here, will the code below for handling rich media execute?

belleklaviyo: hm good point, it won't -- guess this will need to be an if let instead but the rich media guards can stay guards

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -24,6 +40,28 @@ public enum KlaviyoExtensionSDK {
...
+            switch badgeConfig {
+            case .incrementOne:
+                let currentBadgeCount = userDefaults.integer(forKey: "badgeCount")
+                newBadgeValue = currentBadgeCount + 1
+            case .setCount, .setProperty:
+                if let badgeValue = bestAttemptContent.userInfo["badge_value"] as? Int {
+                    newBadgeValue = badgeValue
+                }
+            case .unknown:
+                break
```
kennyklaviyo, line 58: hmm if we break here, this only breaks from the switch statement right? so line 61 and 62 will still execute on `.unknown` i think we'd want to skip entirely, maybe we can just move these into dedicated functions:
```
func handleNotificationServiceDidReceivedRequest(...) {
    handleBadge(...)
    handleMedia(...)
}
```

kennyklaviyo: that way, badge and return however it needs to within its own function and media will still run

belleklaviyo: yeah we can separate them out to be cleaner

kennyklaviyo (commented):
lgtm, gonna let someone with better iOS experience to give the formal approval

ajaysubra (approved):

PR #245, opened 2024-12-18
CHNL-12485: Update V3 API revisions

belleklaviyo:
# Description

Update from 2023-07-15 to 2024-10-15.

# Check List

# Manual Test Plan

Testing will be lumped in with the testing of badge count work [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/244). 

# Supporting Materials
https://developers.klaviyo.com/en/docs/changelog_

### DIFF
--- a/Sources/KlaviyoCore/Networking/NetworkSession.swift
+++ b/Sources/KlaviyoCore/Networking/NetworkSession.swift
@@ -28,7 +28,7 @@ public struct NetworkSession {
         self.data = data
     }
 
-    fileprivate static let currentApiRevision = "2023-07-15"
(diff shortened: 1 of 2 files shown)


ndurell (approved):
Were you able to test this change to ensure this change didn't affect any of our requests?

ajaysubra (approved):

ajaysubra:
> Were you able to test this change to ensure this change didn't affect any of our requests?

@ndurell we met with Paul from DevX and identified that there are no breaking changes to the client APIs we use. However, the reason we are doing this now is so that we can lump it into testing with badge count.

PR #246, opened 2024-12-19
remove `unchecked` from `FileClient`

ab1470:
# Description

I wanted to see if I could remove the `@unchecked` from `FileClient`. Seems like this works, @ndurell what do you think?

### DIFF
--- a/Sources/KlaviyoCore/Utils/FileUtils.swift
+++ b/Sources/KlaviyoCore/Utils/FileUtils.swift
@@ -11,28 +11,34 @@ func write(data: Data, url: URL) throws {
     try data.write(to: url, options: .atomic)
 }
 
-public struct FileClient: @unchecked Sendable {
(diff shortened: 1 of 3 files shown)


Sources/KlaviyoCore/Utils/FileUtils.swift
```diff
@@ -11,28 +11,34 @@ func write(data: Data, url: URL) throws {
     try data.write(to: url, options: .atomic)
 }
 
-public struct FileClient: @unchecked Sendable {
+public struct FileClient: Sendable {
```
ndurell, line 14: Nice! Good catch!

ab1470:
ok well CI is failing now that I've merged this PR. Tests pass locally on Xcode 16.2, so it may be an issue with earlier Xcode versions. I'll look into this tomorrow.

PR #250, opened 2025-01-09
Bump KlaviyoCore 4.0.0 -> 4.1.0

belleklaviyo:
# Description

This release branch was branched off of 4.0.0 which did not have KlaviyoCore updated to 4.0.0 work [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/211) -- but to bring it back into sync this is an effective minor bump from 4.0.0 to 4.1.0 for the additional badge count work + V3 revision update that went into KlaviyoCore

### DIFF
--- a/KlaviyoCore.podspec
+++ b/KlaviyoCore.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "KlaviyoCore"
-  s.version          = "0.1.0"
+  s.version          = "4.1.0"
   s.summary          = "Core functionalities for the Klaviyo SDK"
   s.description      = <<-DESC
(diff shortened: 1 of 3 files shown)


KlaviyoCore.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "KlaviyoCore"
-  s.version          = "0.1.0"
```
dan-peluso, line 3: Oofah

dan-peluso (approved):
Seems simple enough but you might want an ios expert to check this out as well

ab1470:
it looks good but CI is failing. Is this working when you run it locally? Does `pod install` work?

PR #257, opened 2025-01-22
Rel/4.1.1

ajaysubra:
# Description

While releasing RN for badge count noticed that it had a commit that relied on @dan-peluso's PR #236 being in swift so branching from the previous release tag `4.1.0` and cherry picking the above PR. Once released this branch will be deleted. Since this is just a staging branch, no intention to merge after release.

# Check List

# Manual Test Plan

1.

# Supporting Materials


belleklaviyo (approved):
tested the 4.1.2 TF build with this and looks good, thanks for catching this and handling it 🙏

PR #258, opened 2025-01-22
[CHNL-16220] FileIO changes

ab1470:
# Description

This PR makes some changes to the `FileIO` enum:

- it renames FileIO to ResourceLoader, as this object is responsible for reading resources within the app or framework bundle rather than managing files in the os filesystem
- it adds handling for cases when the SDK is imported via Cocoapods. Previously there would be a compiler error because Cocoapods implementations cannot use `Bundle.module`.
- it wraps `ResourceLoader` in `#if DEBUG` tags, because the ResourceLoader is currently only being used for Xcode previews.

# Check List

# Manual Test Plan

1. I've validated that the Xcode previews for `KlaviyoWebViewController` still work as expected


ndurell (approved):
This is all test code correct?

ab1470:
> This is all test code correct?

Yes

PR #259, opened 2025-01-22
[CHNL-16220] present webview from host app

ab1470:
# Description

This PR adds the architecture to a) preload a website, and b) once the site is preloaded, present the website as a modal on top of whatever screen is currently visible.

I have also marked some objects as `@_spi(KlaviyoPrivate)`, because we need to use these objects in the iOS test app without exposing them publicly to consumers of the SKD.

I also added some default implementation for the `preloadWebsite` and `handleNavigationEvent` functions to the `KlaviyoWebViewModeling` protocol.

# Check List

# Manual Test Plan

1. I wired up an example in the iOS test app to validate the changes. 

# Supporting Materials

This video shows the example I wired up in the iOS test app. After tapping the button, the KlavyioWebViewController preloads an "expensive" website (it's loading a very large image). After the preload completes, the image appears as a modal. Notice that I can still navigate through the app while the site is preloading.

https://github.com/user-attachments/assets/f8d2539e-4abd-473f-921c-c10f193a3475

### DIFF
--- a/Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewModel.swift
+++ b/Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewModel.swift
@@ -20,11 +20,6 @@ public protocol KlaviyoWebViewDelegate: AnyObject {
 
 @_spi(KlaviyoPrivate)
 public class KlaviyoWebViewModel: KlaviyoWebViewModeling {
-    enum PreloadError: Error {
-        case timeout
-        case navigationFailed
-    }
-
     public let url: URL
     public let loadScripts: [String: WKUserScript]?
     public weak var delegate: KlaviyoWebViewDelegate?
--- a/Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewModel.swift
+++ b/Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewModel.swift
@@ -42,71 +42,7 @@ public class KlaviyoWebViewModel: KlaviyoWebViewModeling {
         return scripts
     }
 
-    /// Tells the delegate's ``WKWebView`` to preload the URL provided in this ViewModel.
-    ///
-    /// This async method will return after the preload has completed.
-    ///
-    /// By preloading, we can load the URL "headless", so that the ViewController containing
-    /// the ``WKWebView`` will only be presented after the site has successfully loaded.
-    ///
-    /// The caller of this method should `await` completion of this method, then present the ViewController.
-    /// - Parameter timeout: the amount of time, in milliseconds, to wait before throwing a `timeout` error.
-    public func preloadWebsite(timeout: UInt64) async throws {
-        guard let delegate else { return }
-
-        await delegate.preloadUrl()
-
-        do {
-            try await withThrowingTaskGroup(of: Void.self) { group in
-                group.addTask {
-                    try await Task.sleep(nanoseconds: timeout)
-                    throw PreloadError.timeout
-                }
-
-                // Add the navigation event task to the group
-                group.addTask { [weak self] in
-                    guard let self else { return }
-                    for await event in self.navEventStream {
-                        switch event {
-                        case .didFinishNavigation:
-                            return
-                        case .didFailNavigation:
-                            throw PreloadError.navigationFailed
-                        case .didCommitNavigation,
-                             .didStartProvisionalNavigation,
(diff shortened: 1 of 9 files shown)


Sources/KlaviyoUI/KlaviyoWebViewOverlayManager.swift
```diff
@@ -0,0 +1,47 @@
...
+        isLoading = true
+
+        let viewController = KlaviyoWebViewController(viewModel: viewModel)
+        viewController.modalPresentationStyle = modalPresentationStyle
+
+        Task {
+            defer { isLoading = false }
+
+            try await viewModel.preloadWebsite(timeout: 8_000_000_000)
+            topController.present(viewController, animated: true, completion: nil)
```
ndurell, line 44: Could this have changed after 8 seconds?

Tests/KlaviyoUITests/KlaviyoWebViewModelTests.swift
```diff
@@ -5,7 +5,7 @@
 //  Created by Andrew Balmer on 1/21/25.
 //
 
-@testable import KlaviyoUI
+@testable @_spi(KlaviyoPrivate) import KlaviyoUI
```
ndurell, line 8: Do you need this for a test if it's public already ( and 'testable')?

Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewModel.swift
```diff
@@ -9,27 +9,27 @@ import Combine
...
-    let loadScripts: [String: WKUserScript]?
-    weak var delegate: KlaviyoWebViewDelegate?
+@_spi(KlaviyoPrivate)
+public class KlaviyoWebViewModel: KlaviyoWebViewModeling {
+    public let url: URL
+    public let loadScripts: [String: WKUserScript]?
+    public weak var delegate: KlaviyoWebViewDelegate?
 
-    private let (navEventStream, navEventContinuation) = AsyncStream.makeStream(of: WKNavigationEvent.self)
+    public let (navEventStream, navEventContinuation) = AsyncStream.makeStream(of: WKNavigationEvent.self)
```
ndurell, line 30: Seems like a lot of public things in here...

Tests/KlaviyoUITests/KlaviyoWebViewModelTests.swift
ab1470: yeah, it won't compile without both `@testable` and `@_spi`

Sources/KlaviyoUI/KlaviyoWebViewOverlayManager.swift
ab1470: Potentially. I can change the timeout to whatever we think would be appropriate. I just didn't want to wait forever for a website that never loads.

Any suggestions for what I should set the timeout to? 15 seconds? 30?

ndurell: Sorry I mean the top view controller could that change before the site is loaded? The eight seconds is maybe fine...i wish we had a good way to track metrics like this from the sdk.

Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewModel.swift
ab1470: yeah, this was a tradeoff I made in order to make [this change](https://github.com/klaviyo/klaviyo-swift-sdk/pull/259/commits/bd1bc04febfce78e45f57c2d6c5d4556b20acca8) possible. I thought about dropping the `KlaviyoWebViewModeling` protocol and providing default implementation within the `KlaviyoWebViewModel` class, then allowing customization through subclassing, but I didn't like that approach because it doesn't force subclassers to implement their own script handlers

ajaysubra (approved):

Sources/KlaviyoUI/KlaviyoWebViewOverlayManager.swift
ab1470: it's okay if the `topController` changes before the site loads, that's sorta the point of preloading. Notice the video above; I tap the "Show Picsum Overlay" button, which kicks off the preload. While it's preloading, I'm still able to navigate to different tabs (which changes the `topController`). Then, when the preload finishes, it presents the web view on top of whatever the current `topController` is

ndurell (approved):

PR #260, opened 2025-01-24
Removing full forms code 

ajaysubra:
# Description
Since we no longer need to call and parse full forms from the SDK removing that code.


Sources/KlaviyoUI/KlaviyoUI.swift
```diff
@@ -7,8 +7,57 @@
 
 import Foundation
 
-struct KlaviyoUI {
-    func helloWorld() {
-        print("Hello world!")
+struct KlaviyoUIState {
```
ajaysubra, line 10: removing this. committed by mistake.

Sources/KlaviyoUI/KlaviyoUI.swift
```diff
@@ -7,8 +7,10 @@
 
 import Foundation
 
-struct KlaviyoUI {
+public struct KlaviyoUI {
```
ajaysubra, line 10: This needs to be public anyways so keeping it in.

ndurell (approved):

ab1470 (approved):

dan-peluso (approved):

PR #261, opened 2025-01-25
[CHNL-16564] update script and message handling

ab1470:
# Description

This PR makes some updates to the KlaviyoWebViewModeling protocol, all conforming ViewModels, and the KlaviyoWebViewController. We will need these changes in order to handle the javascript messages that the in-app forms emit. 

With this PR, I'm decoupling the javascript that gets injected into the WKWebView from the message handlers that get injected. This will allow us, in a future PR, to tell the WKWebView to listen for messages from  the `KlaviyoNativeBridge`, even though we aren't injecting a script with the same name. 

# Check List

# Manual Test Plan

1. I've tested this using Xcode previews and with the iOS test app

### DIFF
--- a/Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
+++ b/Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
@@ -11,9 +11,9 @@ import Foundation
 import WebKit
 
 class JSTestWebViewModel: KlaviyoWebViewModeling {
-    public enum MessageHandler: String, CaseIterable {
+    private enum MessageHandler: String, CaseIterable {
         case toggleMessageHandler
-        case closeHandler
+        case closeMessageHandler
     }
 
     weak var delegate: KlaviyoWebViewDelegate?
@@ -78,7 +78,7 @@ class JSTestWebViewModel: KlaviyoWebViewModeling {
                     print("Javascript evaluation failed; message: \(error.localizedDescription)")
                 }
             }
-        case .closeHandler:
+        case .closeMessageHandler:
             Task {
                 await delegate?.dismiss()
             }
--- a/Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
+++ b/Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
@@ -11,23 +11,34 @@ import Foundation
 import WebKit
 
 class JSTestWebViewModel: KlaviyoWebViewModeling {
-    let url: URL
-    let loadScripts: [String: WKUserScript]?
+    public enum MessageHandler: String, CaseIterable {
+        case toggleMessageHandler
+        case closeHandler
+    }
+
     weak var delegate: KlaviyoWebViewDelegate?
+
+    let url: URL
+    var loadScripts: Set<WKUserScript>? = JSTestWebViewModel.initializeLoadScripts()
+    var messageHandlers: Set<String>? = Set(MessageHandler.allCases.map(\.rawValue))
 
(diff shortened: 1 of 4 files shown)


Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewModeling.swift
```diff
@@ -14,8 +14,10 @@ public protocol KlaviyoWebViewModeling: AnyObject {
     var url: URL { get }
     var delegate: KlaviyoWebViewDelegate? { get set }
 
-    /// Scripts to be injected into the ``WKWebView`` when the website loads.
-    var loadScripts: [String: WKUserScript]? { get }
+    /// Scripts & message handlers to be injected into the ``WKWebView`` when the website loads.
+    var loadScripts: Set<WKUserScript>? { get }
```
ajaysubra, line 18: Do we need this since @evan-masseau mentioned that he doesn't think we may end up injecting scripts on the native side?

Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
```diff
@@ -11,23 +11,34 @@ import Foundation
 import WebKit
 
 class JSTestWebViewModel: KlaviyoWebViewModeling {
```
ajaysubra, line 13: Is this the view model that we'd use to present a Klaviyo form?

ajaysubra (approved):
Some minor questions but looks good to me

Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewModeling.swift
ab1470: yes, I'd like to leave it for 2 reasons:
1. It gives us flexibility to easily inject scripts if/when we want to. Evan said on Friday that he suspects we may want/need to do that in the future
2. It's optional, so if we decide we're not going to inject any scripts we can simply set that value to `nil`. 

The way I built this, the `KlaviyoWebViewController` is flexible and can be reused when injected with different ViewModels, as long as the VMs conform to the `KlaviyoWebViewModeling` protocol. So we can, for example:

-  instantiate one `KlaviyoWebViewController` using a viewModel `WebsiteWithInjectedScriptsViewModel`
-  instantiate another `KlaviyoWebViewController` using a viewModel `WebsiteWithoutInjectedScriptsViewModel`

Sources/KlaviyoUI/KlaviyoWebView/Development Assets/JSTestWebViewModel.swift
ab1470: No, this is just for dev/testing purposes right now. Whatever we put into production for handling a Klaviyo form will be a different VM that confirms to `KlaviyoWebViewModeling`. That protocol will mandate that we implement all properties and methods that the `KlaviyoWebViewController` depends on

ajaysubra: That sounds great! Do we need a ticket for that work or do you think it'll be captured in existing work?

ab1470: I don't think we need a ticket, I think it'll be captured in existing work. Creating that new VM will be easy

Sources/KlaviyoUI/KlaviyoWebView/KlaviyoWebViewModeling.swift
evan-masseau: yeah its a good pattern to have for the future if we do need injected scripts.

evan-masseau (approved):

PR #262, opened 2025-01-28
[CHNL-16564] move console handler to sdk

ab1470:
# Description

I had previously opened a PR in the iOS test app repo to relay Javascript console log messages from the WKWebView to the native layer, to make dev/debugging easier. I realized that it would be better to have this code live in the Swift SDK, and to add the console logging within the KlaviyoWebViewController, rather than in the viewModels that get injected to the KlaviyoWebViewController. 

With these changes, any KlaviyoWebViewController that's presented on a debug build will automatically relay all Javascript console log messages to the native layer, and the native layer will log those messages to the Xcode console. This logic will not be executed in production builds.

This also adds some more robust logging to the `ResourceLoader`.

I will open a related PR in the iOS test app shortly.

# Check List

# Manual Test Plan

I tested this by running the iOS test app and validating that javascript console log messages are successfully relayed to the native layer, and the native layer outputs those messages to the Xcode console.

### DIFF
--- a/Sources/KlaviyoUI/KlaviyoWebView/Resources/Scripts/consoleHandler.js
+++ b/Sources/KlaviyoUI/KlaviyoWebView/Resources/Scripts/consoleHandler.js
@@ -17,18 +17,6 @@
   //Initialize web wrapper object
   window.WebViewBridge = new Bridge(JSON.parse(strJsonConfig));
 
-  if (/complete|interactive|loaded/.test(document.readyState)) {
-    // In case the document has finished parsing, document's readyState will
-    // be one of "complete", "interactive" or (non-standard) "loaded".
-    WebViewBridge.initialize();
-  } else {
-    // The document is not ready yet, so wait for the DOMContentLoaded event
-    // https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event
-    document.addEventListener('DOMContentLoaded', function () {
-      WebViewBridge.initialize();
-    }, false);
-  }
-
   /**
    * Bridge object to handle communication between JavaScript and Native
    */
@@ -37,40 +25,6 @@
 
     return {
       opts: opts,
-
-      initialize: function () {
-        WebViewBridge.postMessageToNative("documentReady", {});
-
-        var images = document.querySelectorAll("img");
-        var loadedCount = 0;
-        var onImgLoad = function() {
-          loadedCount++;
-
-          if (loadedCount === images.length) {
-            WebViewBridge.postMessageToNative("imagesLoaded", {});
-          }
-        };
-
-        images.forEach(function (img) {
-          if (img.complete) {
-            onImgLoad();
-          } else {
-            img.onload = onImgLoad;
-          }
-        });
-      },
-
-      /**
-       * Default keyword for JS -> Native messaging
-       */
-      defaultAction: opts.defaultAction || "message",
-
-      /**
-       * Bool test if this is Android WebView with registered JavaScript interface
-       */
-      isAndroid: function () {
-        return !!window[opts.bridgeName];
-      },
 
       /**
        * Bool test if this is an Apple WKWebView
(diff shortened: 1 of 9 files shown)


Sources/KlaviyoUI/KlaviyoWebView/Resources/Scripts/consoleHandler.js
```diff
@@ -0,0 +1,84 @@
...
+       */
+      postMessageToNative: function (payload) {
+        try {
+          const serializedPayload = JSON.stringify(payload);
+
+          if (this.isApple()) {
+            window.webkit.messageHandlers[opts.bridgeName].postMessage(serializedPayload);
+          }
+        } catch (e) {
+          console.error("Failed to post message to native layer:", e.message);
```
evan-masseau, line 53: There is a potential to enter an endless loop, I had an updated version of this someplace that would undo the console-linkage below to prevent that, I'll find it

evan-masseau:
```suggestion
          console.error("Failed to post message to native layer:", e.message);
          unlinkConsole()
```

Sources/KlaviyoUI/KlaviyoWebView/Resources/Scripts/consoleHandler.js
```diff
@@ -0,0 +1,84 @@
...
+        }
+      },
+    }
+  }
+
```
evan-masseau, line 58:
```suggestion
function unlinkConsole() {
    ["log", "warn", "error"].forEach(function (method) {
      var bckKey = "_" + method
      console[method] = console[bckKey];
    });
}
```

Sources/KlaviyoUI/KlaviyoWebView/Resources/Scripts/consoleHandler.js
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Send all console output to native layer
+   */
+  if (!!window.WebViewBridge.opts.linkConsole) {
+    ["log", "warn", "error"].forEach(function (method) {
+      var _method = console[method];
+      console[method] = function () {
```
evan-masseau, line 65:
```suggestion
      var bckKey = "_" + method
      console[bckKey] = _method
      console[method] = function () {
```

evan-masseau: Idea is to just keep an instance of the original implementation around, and if there's an issue with the bridge, reset functionality to that. 
Full disclosure I wrote this in github without running it, I couldn't find the first copy i'd written.

ajaysubra (approved):

Sources/KlaviyoUI/KlaviyoWebView/Resources/Scripts/consoleHandler.js
ab1470: thanks for the suggestions. I tested it locally and it looks good

Sources/KlaviyoUI/Utilities/Logger+Ext.swift
```diff
@@ -0,0 +1,33 @@
...
+
+    init(category: String = #file) {
+        self.init(subsystem: Self.subsystem, category: category)
+    }
+}
+
+// MARK: - Loggers
+
+@available(iOS 14.0, *)
+extension Logger {
```
evan-masseau, line 22: love this btw

Sources/KlaviyoUI/KlaviyoWebView/Resources/Scripts/consoleHandler.js
```diff
@@ -0,0 +1,95 @@
+/**
+ * WARNING This JS file should be ES5 compatible to support older versions of Android WebView
+ *
```
evan-masseau, line 3:
```suggestion
 * WARNING This JS file should be ES5 compatible to support older versions of Android WebView
 *
```
This was mostly an android issue, may as well remove

Sources/KlaviyoUI/KlaviyoWebView/Resources/Scripts/consoleHandler.js
```diff
@@ -0,0 +1,95 @@
...
+ * Inject this script file from native code into the head of the HTML document
+ * Format string like this, where strWrapperScript is the contents of this file
+ * and strJsonConfig is the json-encoded handoff parameters detailed below
+ * Always be careful to put quotes around your string arguments! *
+ *
+ *    strWrapperScript + "('" + strJsonConfig + "');"
+ *
+ * @param strJsonConfig - JSON encoded handoff dictionary containing:
+ *    bridgeName: String - name of the native message handler
+ *    defaultAction: String - default action keyword for data posted from JS -> Native
```
evan-masseau, line 13:
```suggestion
```

evan-masseau (approved):

Sources/KlaviyoUI/KlaviyoWebView/Resources/Scripts/consoleHandler.js
ab1470: thanks, good catches

PR #264, opened 2025-01-29
Enable enqueuing from KlaviyoUI

belleklaviyo:
WIP branch for now where KlaviyoUI is able to send a custom event with name "From JS" to the queue. This structure with `@_spi(KlaviyoPrivateQueue)` somewhat protects access to the queue.

Current WIP:
- refactoring CreateEventPayload to support the new `profileEventTracked` type

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
+++ b/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
@@ -12,6 +12,15 @@ import UIKit
 
 typealias DeviceMetadata = PushTokenPayload.PushToken.Attributes.MetaData
 
+public class KlaviyoEventQueue: NSObject {
+    public static let shared = KlaviyoEventQueue()
+
+    @_spi(KlaviyoPrivateQueue)
+    public func enqueue(event: Event) {
+        dispatchOnMainThread(action: .enqueueEvent(event))
+    }
+}
+
 struct KlaviyoState: Equatable, Codable {
     enum InitializationState: Equatable, Codable {
         case uninitialized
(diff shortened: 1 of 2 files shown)


Sources/KlaviyoSwift/Models/Event.swift
```diff
@@ -10,6 +10,10 @@ import Foundation
 import KlaviyoCore
 
 public struct Event: Equatable {
+    public enum IAFProfileEvent: Equatable {
```
ajaysubra, line 13: Curious about the changes in the public `Event` object? Can't we just use the `properties` dictionary?

ajaysubra, line 13: I don't think this needs to be in the `KlaviyoSwift` I would keep it local to `KlaviyoUI`

Sources/KlaviyoCore/Models/APIModels/CreateEventPayload.swift
```diff
@@ -84,6 +84,7 @@ public struct CreateEventPayload: Equatable, Codable {
         public var attributes: Attributes
         public init(name: String,
                     properties: [String: Any]? = nil,
+                    formProperties: [String: Any]? = nil,
```
ajaysubra, line 87: Same question as in the public struct - why this new parameter instead of just sending it in `properties`?

belleklaviyo: the forms properties seemed separate from the metadata properties to me, and I thought if we end up expanding what we're sending in a form event it might be nice to be separate -- but guess we can lump them together

PR #265, opened 2025-02-03
removed unused `AnyCodable`

ab1470:
# Description

This PR removes some unnecessary `AnyCodable` usages.

It's unrelated to IAF changes, but I came across these while working on IAF so I figured I'd put up a PR.

# Check List

# Manual Test Plan

I ran the unit test suite and I validated that events fire correctly in the test app


ndurell (approved):
Nice!

ajaysubra (approved):

PR #266, opened 2025-02-03
[CHNL-16694] move IAF viewmodel to SDK

ab1470:
# Description

This PR, in conjunction with [this test app PR](https://github.com/klaviyo/klaviyo-ios-test-app/pull/53), moves the IAF ViewModel from the test app to the SDK. I had initially added the ViewModel to the test app thinking it would be more useful to have it there during development so we could make changes fully within the test app repo without touching the SDK, but I think it turns out that we can just as easily to the dev work within the SDK. 

The new `IafWebViewModel` will be the production viewModel for managing the In-App Form, though of course we'll continue making changes to it in the next few weeks during the dev process.

# Check List

# Manual Test Plan

1. While using the `ab/CHNL-16694/move-IAF-viewmodel-to-SDK` branch of the test app, I validated that the In-App Form displays correctly when tapping the "Show In-App Form Test" button

### DIFF
--- a/Sources/KlaviyoUI/InAppForms/IafPresentationManager.swift
+++ b/Sources/KlaviyoUI/InAppForms/IafPresentationManager.swift
@@ -1,0 +1,57 @@
+//
+//  IafPresentationManager.swift
+//  klaviyo-swift-sdk
+//
+//  Created by Andrew Balmer on 2/3/25.
+//
+
+import Foundation
+import OSLog
+import UIKit
+
+class IafPresentationManager {
+    static let shared = IafPresentationManager()
+
+    lazy var indexHtmlFileUrl: URL? = {
+        do {
+            return try ResourceLoader.getResourceUrl(path: "InAppFormsTemplate", type: "html")
+        } catch {
+            return nil
+        }
+    }()
+
+    private var isLoading: Bool = false
+
+    @MainActor func presentIaf() {
+        guard !isLoading else {
+            if #available(iOS 14.0, *) {
+                Logger.webViewLogger.log("In-App Form is already loading; ignoring request.")
+            }
+            return
+        }
+
+        guard let fileUrl = indexHtmlFileUrl else {
+            if #available(iOS 14.0, *) {
+                Logger.webViewLogger.warning("URL for local HTML file is nil; unable to present In-App Form.")
+            }
+            return
+        }
+
+        isLoading = true
+
+        let viewModel = IafWebViewModel(url: fileUrl)
+        let viewController = KlaviyoWebViewController(viewModel: viewModel)
+        viewController.modalPresentationStyle = .overCurrentContext
+
+        Task {
+            defer { isLoading = false }
+
+            try await viewModel.preloadWebsite(timeout: 8_000_000_000)
+
+            guard let topController = UIApplication.shared.topMostViewController else {
+                return
+            }
+            topController.present(viewController, animated: true, completion: nil)
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


Sources/KlaviyoUI/InAppForms/IafPresentationManager.swift
```diff
@@ -0,0 +1,60 @@
...
+//
+//  Created by Andrew Balmer on 2/3/25.
+//
+
+import Foundation
+import OSLog
+import UIKit
+
+@_spi(KlaviyoPrivate)
+public class IafPresentationManager {
```
ajaysubra, line 13: nit-pick: should we say IAF in caps?

ajaysubra: also, why is this exposed using private apis?

Sources/KlaviyoUI/InAppForms/IafPresentationManager.swift
```diff
@@ -0,0 +1,60 @@
...
+        isLoading = true
+
+        let viewModel = IafWebViewModel(url: fileUrl)
+        let viewController = KlaviyoWebViewController(viewModel: viewModel)
+        viewController.modalPresentationStyle = .overCurrentContext
+
+        Task {
+            defer { isLoading = false }
+
+            try await viewModel.preloadWebsite(timeout: 8_000_000_000)
```
ajaysubra, line 52: should we remove the timeout?

Sources/KlaviyoUI/InAppForms/IafWebViewModel.swift
```diff
@@ -0,0 +1,45 @@
...
+import WebKit
+
+class IafWebViewModel: KlaviyoWebViewModeling {
+    private enum MessageHandler: String, CaseIterable {
+        case klaviyoNativeBridge = "KlaviyoNativeBridge"
+    }
+
+    weak var delegate: KlaviyoWebViewDelegate?
+
+    let url: URL
```
ajaysubra, line 19: I noted this in my notes in the tech spec, but should we have an initializer that takes in HTML instead of URL so that we are able to string replace company id and other placeholder stuff in the HTML?

Sources/KlaviyoUI/Utilities/ResourceLoader.swift
```diff
@@ -15,6 +15,9 @@ enum ResourceLoaderError: Error {
 }
 
 enum ResourceLoader {
```
ajaysubra, line 17: This was super useful during my PoC for initializer?

Sources/KlaviyoUI/InAppForms/IafPresentationManager.swift
ab1470: > nit-pick: should we say IAF in caps?

Yeah you're right, swift convention is to capitalize acronyms (`URLSession`, `HTTPClient`, etc). I'll update.

Sources/KlaviyoUI/InAppForms/IafWebViewModel.swift
ab1470: Yeah I think we should, but I'd like to handle that in a separate PR. Should be pretty straightforward.

Sources/KlaviyoUI/InAppForms/IafPresentationManager.swift
ajaysubra: So what are the circumstances under which pre loading will fail? @evan-masseau  and I have been talking about this as well and he was suggesting some sort of abort message but we should discuss this more broadly and have some common approach to both platforms.

ab1470: I think it could pretty easily fail if the user's internet connection drops, for example. Or if there's some kind of backend error and the form can't get delivered to the user's device

ajaysubra (approved):

PR #268, opened 2025-02-04
Enqueue profile events

belleklaviyo:
# Description

Builds out the .trackProfileEvent case of IAFMessageBusEvent and enqueues events utilizing the existing `KlaviyoSDK().create(event:..)` call

~~Note: for now I've put this in to be triggered through the closeHandler and added the extra listener in the `InAppFormsTemplate` to help test it~~ Edit: removed

# Check List

# Manual Test Plan

Was able to trigger the form and close it and send an event from that dumps the native bridge payload properties into the properties of the event.

Note: if we are using a Custom Event I assume we need to pull out the specific event name to be able to provide that as the event name

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoUI/InAppForms/Assets/InAppFormsTemplate.html
+++ b/Sources/KlaviyoUI/InAppForms/Assets/InAppFormsTemplate.html
@@ -6,9 +6,57 @@
       data-sdk-version="SDK_VERSION"
 >
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <title>Klaviyo In-App Form Test - 1</title>
-        <script type="text/javascript" src="http://localhost:8080/onsite/js/klaviyo.js?company_id=9BX3wh&env=in-app"></script>
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+            <title>Klaviyo In-App Form Test - 1</title>
+            <script type="text/javascript" src="http://localhost:8080/onsite/js/klaviyo.js?company_id=9BX3wh&env=in-app"></script>
+            <script type="text/javascript">
+                document.addEventListener('DOMContentLoaded', function () {
+                    onElementAvailable("button.klaviyo-close-form", function () {
+                        onElementRemoved("button.klaviyo-close-form", function () {
+                            var name = document.head.dataset.klaviyoMessageHandlerName
+                            window.webkit.messageHandlers.closeHandler.postMessage(JSON.stringify({
+                                "type": "trackProfileEvent",
+                                "version": 1,
+                                "data": {
+                                    "metric": "Form completed by profile",
+                                    "properties": {
+                                        "form_id": "7uSP7t",
+                                        "form_version_id": 8
+                                    }
+                                }
+                            }));
+                        });
+                    });
+                }, false);
+
+                function onElementAvailable(selector, callback) {
+                    var observer = new MutationObserver(function () {
+                        if (document.querySelector(selector)) {
+                            observer.disconnect();
+                            callback();
+                        }
+                    });
+
+                    observer.observe(document.body, {
+                        childList: true,
+                        subtree: true
+                    });
+                }
+
+                function onElementRemoved(selector, callback) {
+                    var observer = new MutationObserver(function () {
+                        if (!document.querySelector(selector)) {
+                            observer.disconnect();
+                            callback();
+                        }
+                    });
+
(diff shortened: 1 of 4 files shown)


ajaysubra (commented):
seeing some code from Andrew PR?

Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
```diff
@@ -59,9 +81,10 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
             ()
         case let .trackAggregateEvent(data):
             KlaviyoSDK().create(aggregateEvent: data)
-        case .trackProfileEvent:
-            // TODO: handle tracktProfileEvent
-            ()
+        case let .trackProfileEvent(data):
+            if let json = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any] {
+                KlaviyoSDK().create(event: Event(name: .customEvent(json["metric"] as! String), properties: json))
```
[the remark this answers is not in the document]
ajaysubra: brittle because if the key in the dictionary changes from metric to something else this breaks? 

Also, I'm guessing the force cast is to test things out and you plan on gracefully handling the case where type cast fails?

belleklaviyo:
> seeing some code from Andrew PR?

yeah this was originally based on his branch before it was merged to master, and then it automatically rebased to master

Sources/KlaviyoUI/InAppForms/Assets/InAppFormsTemplate.html
```diff
@@ -6,9 +6,57 @@
       data-sdk-version="SDK_VERSION"
 >
     <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <title>Klaviyo In-App Form Test - 1</title>
-        <script type="text/javascript" src="http://localhost:8080/onsite/js/klaviyo.js?company_id=9BX3wh&env=in-app"></script>
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
```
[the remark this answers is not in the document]
ajaysubra: why the meta tag?

ajaysubra (approved):
LGTM minor questions

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -13,7 +13,7 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
     case formsDataLoaded
     case formAppeared
     case trackAggregateEvent(Data)
-    case trackProfileEvent
+    case trackProfileEvent(Data)
```
ab1470, line 16: ah, so I think I might have made things a bit confusing with how I implemented the `trackAggregateEvent` case. I made the associated value for that case a `Data` type only because, for that specific case, we don't really have a need to decode the data into a Swift struct. We simply need to pass the raw data on to the onsite API.

I think for the other enum cases, we should actually decode the raw data into a Swift structure. That will make the data easier to work with. [Here's an example](https://gist.github.com/ab1470/ec42c8d4cac818c60768928b9374857a) of what I was thinking the `trackProfileEvent` would look like.

Sorry for creating confusion here

belleklaviyo: I considered that, but considering the properties may contain more fields in the future, I thought sending the entire raw data would be the most flexible, so we wouldn't have to update that struct with every new field

ab1470: ahh okay, that's a good point

ab1470: >  considering the properties may contain more fields in the future
is this speculation or is it based on something specific you heard/read somewhere from someone?

belleklaviyo: in the message bus [doc](https://docs.google.com/document/d/1MVhfEzdoNW6yYC1NBo1O8dUXxNngRZGU79KyOtHI-XM/edit?tab=t.0), here: `Expect properties could contain more fields in the future, so model it as a dict of some kind like string:serializable`

ab1470: nice, good catch

ab1470 (approved):

PR #269, opened 2025-02-05
CHNL-16703: Enable deep links from IAF

belleklaviyo:
# Description
Added support for deep links building out the `IAFNativeBridgeEvent` for `openDeepLink`

# Check List

# Manual Test Plan

For testing I (re)added the close event listener + closeHandler to the template HTML, so on close, 
```
{
    "type": "openDeepLink",
    "version": 1,
    "data": {
        "ios": "klaviyotest://accountInfo",
        "android": "klaviyotest://settings"
    }
}
```
is posted and sent to the handler and opens it

https://github.com/user-attachments/assets/931ea501-4d71-4c09-a367-504e3e31ee18

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
+++ b/Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
@@ -14,7 +14,7 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
     case formAppeared
     case trackAggregateEvent(Data)
     case trackProfileEvent(Data)
-    case openDeepLink
+    case openDeepLink(URL)
 
     private enum CodingKeys: String, CodingKey {
         case type
@@ -47,7 +47,14 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
             let data = try JSONEncoder().encode(decodedData)
             self = .trackProfileEvent(data)
         case .openDeepLink:
-            self = .openDeepLink
+            let url = try container.decode(DeepLinkEventPayload.self, forKey: .data)
+            self = .openDeepLink(url.ios)
         }
+    }
+}
+
+extension IAFNativeBridgeEvent {
+    struct DeepLinkEventPayload: Codable {
+        let ios: URL
     }
 }
--- a/Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
+++ b/Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
@@ -14,7 +14,7 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
     case formAppeared
     case trackAggregateEvent(Data)
     case trackProfileEvent(Data)
-    case openDeepLink(URL)
+    case openDeepLink(String)
 
     private enum CodingKeys: String, CodingKey {
         case type
@@ -55,6 +55,6 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
 
 extension IAFNativeBridgeEvent {
     struct DeepLinkEventPayload: Codable {
-        let ios: URL
+        let ios: String
     }
 }
--- a/Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
+++ b/Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
@@ -14,7 +14,7 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
     case formAppeared
     case trackAggregateEvent(Data)
     case trackProfileEvent(Data)
-    case openDeepLink(String)
+    case openDeepLink(URL)
 
     private enum CodingKeys: String, CodingKey {
         case type
@@ -55,6 +55,6 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
 
 extension IAFNativeBridgeEvent {
     struct DeepLinkEventPayload: Codable {
-        let ios: String
+        let ios: URL
     }
 }
--- a/Tests/KlaviyoUITests/IAFNativeBridgeEventTests.swift
+++ b/Tests/KlaviyoUITests/IAFNativeBridgeEventTests.swift
@@ -24,11 +24,23 @@ struct IAFNativeBridgeEventTests {
         }
         """
 
+        let deepLinkEvent = """
+        {
+            "ios": "klaviyotest://settings",
+            "android": "klaviyotest://settings"
+        }
+        """
+
         let data = json.data(using: .utf8)!
         let event = try JSONDecoder().decode(IAFNativeBridgeEvent.self, from: data)
-        #expect(event == .openDeepLink)
+        guard case let .openDeepLink(url) = event else {
+            Issue.record("event type should be .openDeepLink but was '.\(event)'")
+            return
+        }
+        let deepLinkEventData = try #require(deepLinkEvent.data(using: .utf8))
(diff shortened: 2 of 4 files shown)


Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -47,7 +47,14 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
...
-            self = .openDeepLink
+            let url = try container.decode(DeepLinkEventPayload.self, forKey: .data)
+            self = .openDeepLink(url.ios)
         }
     }
 }
+
+extension IAFNativeBridgeEvent {
+    struct DeepLinkEventPayload: Codable {
+        let ios: URL
```
belleklaviyo, line 58: we are receiving both `ios` and `android` from the payload but imagine we only care about iOS so it'll just contain this

evan-masseau: What'll happen here if the string is not a valid URL?

ajaysubra: @evan-masseau was this going to change based on the SDK name that gets injected into JS?

ajaysubra (approved):
Nice!

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
evan-masseau: I hadn't considered that, but yes i suppose it could. Back to the event spec doc I guess

belleklaviyo: if it's not a valid URL it would throw an error when we try to decode it to this type (line 50) -- I can add safer handling to it so it just silently exits. I also assumed we were doing validation on the message bus side, but if we want to just treat it as a string and safely cast it to a URL on this end we could

evan-masseau: Yeah lets put in safer handling here, as far as I know, right now there's no point earlier in the system that's performing URI validation on this, and regardless its best to validate at each level.

ab1470: @belleklaviyo what were you thinking in terms of safer handling? The way it's currently written, any time we try to decode data into an `IAFNativeBridgeEvent` type, we're forced to handle a potential decoding error (like [here](https://github.com/klaviyo/klaviyo-swift-sdk/blob/d1f0453eeea98cfdbdabd327f79e095a2c648e87/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift#L47), for example). So I think we already do safely handle a decoding error

ajaysubra: but we don't validate URL anywhere no? It just probably won't call the delegate method. There is a method called [canOpenURL](https://developer.apple.com/documentation/uikit/uiapplication/canopenurl(_:)) which we could use.

evan-masseau: > if it's not a valid URL it would throw an error when we try to decode it to this type (line 50)

Forgive my swift ignorance, will that inline `try` on line 50 throw an error or fail silently? I just don't want to introduce a potential crash of the host app. But if that is handled, then ignore what I said!

ajaysubra: Good catch @evan-masseau. @belleklaviyo might want to wrap that in a do-catch block.

ajaysubra: nvm looks like she's already aware of that - https://github.com/klaviyo/klaviyo-swift-sdk/pull/269/files#r1943703046

ab1470: @belleklaviyo sorry, I'm not sure I agree with the change you made. I think you can keep the associated value as a `URL` type. We already safely handle the case where it's not a valid URL (see [my comment above](https://github.com/klaviyo/klaviyo-swift-sdk/pull/269#discussion_r1943712814)). 

>Forgive my swift ignorance, will that inline try on line 50 throw an error or fail silently?

@evan-masseau, it throws an error and we're forced to handle it

belleklaviyo: ahh okay yeah sure, I'll still also add in the `canOpenURL` still though as an extra check cause it seems neat (I didn't know that existed lol)

Tests/KlaviyoUITests/IAFNativeBridgeEventTests.swift
ab1470, line 1: I think you can make this test a bit simpler, like this:

```swift
@Test func testDecodeOpenDeepLink() async throws {
    let json = """
    {
      "type": "openDeepLink",
      "data": {
        "ios": "klaviyotest://settings",
        "android": "klaviyotest://settings"
      }
    }
    """

    let data = try #require(json.data(using: .utf8))
    let event = try JSONDecoder().decode(IAFNativeBridgeEvent.self, from: data)
    guard case let .openDeepLink(url) = event else {
        Issue.record("event type should be .openDeepLink but was '.\(event)'")
        return
    }

    let expectedUrl = try #require(URL(string: "klaviyotest://settings"))
    #expect(url == expectedUrl)
}
```

The reason my [testDecodeAggregateEvent test](testDecodeAggregateEvent) has two separate JSON strings (the "full" JSON and the "inner" JSON) is that I needed to separately encode the "inner" JSON as a `Data` type so I could compare that `Data` to the `IAFNativeBridgeEvent.trackAggregateEvent` case's associatedValue and make sure they match.

In the case of the `testDecodeOpenDeepLink` test, we only care in the end that, after we decode the "full" JSON to an `IAFNativeBridgeEvent` type, 1) it's successfully decoded into the `.openDeepLink` case, and 2) the associatedValue for that case is a URL and that URL's `absoluteString` == "klaviyotest://settings". We can test (1) and (2) directly, without having to create a separate "inner" JSON string and going through that whole process

Tests/KlaviyoUITests/IAFNativeBridgeEventTests.swift
```diff
@@ -24,11 +24,23 @@ struct IAFNativeBridgeEventTests {
...
         """
 
+        let deepLinkEvent = """
+        {
+            "ios": "klaviyotest://settings",
+            "android": "klaviyotest://settings"
+        }
+        """
+
         let data = json.data(using: .utf8)!
```
ab1470, line 34:
```suggestion
        let data = try #require(json.data(using: .utf8))
```
this allows us to avoid force-unwrapping. It'll throw an error and the test will fail if the `json.data()` command fails

Tests/KlaviyoUITests/IAFNativeBridgeEventTests.swift
belleklaviyo: okay yeah while writing the test it felt a bit like I had too many layers, but think I was confusing myself with the layers of nesting and pulling out the `.ios` prop versus just comparing it to the string/url itself, and it felt safer to encode it into the object first then pull it out but guess that's a bit redundant -- makes sense now, thanks for the explanation!

ab1470 (approved):

PR #270, opened 2025-02-06
CHNL-17243: Handle lifecycle events from JS

belleklaviyo:
# Description
Calls `presentIAF()` when receiving `formWillAppear` and created new method `dismissIAF()` for receiving the`formDisappeared` case.

# Check List

# Manual Test Plan
In this mini demo I added some print statements to see when the different messages were being received. You can also tell the webview is dismissed when the form is dismissed as the rest of the app is interactable again/you can navigate away -- as well as reopen the form, etc.

https://github.com/user-attachments/assets/7943a779-7956-4125-a9eb-5cd8f1e732c6

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
@@ -57,4 +57,14 @@ public class IAFPresentationManager {
             topController.present(viewController, animated: true, completion: nil)
         }
     }
+
+    @_spi(KlaviyoPrivate)
+    @MainActor public func dismissIAF() {
+        Task {
+            guard let topController = UIApplication.shared.topMostViewController else {
+                return
+            }
+            topController.dismiss(animated: true)
+        }
+    }
 }
--- a/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
@@ -57,14 +57,4 @@ public class IAFPresentationManager {
             topController.present(viewController, animated: true, completion: nil)
         }
     }
-
-    @_spi(KlaviyoPrivate)
-    @MainActor public func dismissIAF() {
-        Task {
-            guard let topController = UIApplication.shared.topMostViewController else {
-                return
-            }
-            topController.dismiss(animated: true)
-        }
-    }
 }
--- a/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
@@ -55,8 +55,8 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
             // TODO: handle formsDataLoaded
             ()
         case .formWillAppear:
-            Task { @MainActor in
-                IAFPresentationManager.shared.presentIAF()
+            Task {
+                await IAFPresentationManager.shared.presentIAF()
             }
         case let .trackAggregateEvent(data):
             KlaviyoInternal.create(aggregateEvent: data)
@@ -70,8 +70,8 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
                 UIApplication.shared.open(url)
             }
         case .formDisappeared:
-            Task { @MainActor in
-                IAFPresentationManager.shared.dismissIAF()
+            Task {
+                await delegate?.dismiss()
             }
         }
     }
--- a/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
@@ -8,6 +8,7 @@
 import Combine
 import Foundation
 import KlaviyoSwift
+import OSLog
 import WebKit
 
 class IAFWebViewModel: KlaviyoWebViewModeling {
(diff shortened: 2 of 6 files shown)


ajaysubra:
I'm seeing the [PR](https://github.com/klaviyo/fender/pull/37578) for this is merged. Are you able to get the events over the JS bridge?

ajaysubra:
> I'm seeing the [PR](https://github.com/klaviyo/fender/pull/37578) for this is merged. Are you able to get the events over the JS bridge?

nvm I'm seeing the demo now and it does look like you are indeed receiving those! Nice

Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
```diff
@@ -57,4 +57,14 @@ public class IAFPresentationManager {
             topController.present(viewController, animated: true, completion: nil)
         }
     }
+
+    @_spi(KlaviyoPrivate)
```
ajaysubra, line 61: why does this need to be swift private method?

ajaysubra (approved):
Nice.

Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
```diff
@@ -68,6 +69,10 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
             if UIApplication.shared.canOpenURL(url) {
                 UIApplication.shared.open(url)
             }
+        case .formDisappeared:
+            Task { @MainActor in
+                IAFPresentationManager.shared.dismissIAF()
+            }
```
ab1470, line 75: was there a particular reason for doing this instead of just calling `viewModel.dismiss()`? 

I'm a bit concerned about the implementation of your `dismissIAF()` method, as it dismisses whatever the topMostViewController at the time is. The topMostViewController *should* be the KlaviyoWebViewController, but I wonder if the app could get into a state where something else is the topMostViewController (for example, if the host app presents its own view on top of the KlaviyoWebViewController).

I think the existing `viewModel.dismiss()` should work fine. It essentially tells the `KlaviyoWebViewController` that hosts the viewModel to dismiss itself

belleklaviyo: not particularly, I was just trying to have the appear handler and dismiss handler have similar points of call to the `IAFPresentationManager` -- but that's a good point and I could see how that'd be problematic. I'll just use the delegate call instead

ab1470: sounds good. I had thought about putting a reference to the `KlaviyoWebViewController` instance win the `IAFPresentationManager`, in which case the `IAFPresentationManager` could be responsible for dismissing the `KlaviyoWebViewController` (by calling `viewController.dismiss()`). But for now I don't think that's necessary, I think it's fine for the viewController to dismiss itself via the way I describe above

Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
```diff
@@ -54,9 +54,10 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
         case .formsDataLoaded:
             // TODO: handle formsDataLoaded
             ()
-        case .formAppeared:
-            // TODO: handle formAppeared
-            ()
+        case .formWillAppear:
+            Task {
+                await IAFPresentationManager.shared.presentIAF()
+            }
```
ab1470, line 60: My understanding of the `formWillAppear` event is that it gets emitted after all form content has finished loading and the form is ready to appear. If this is correct, then we basically want to tie this `formWillAppear` event into the `preloadWebsite()` architecture; i.e., when calling `preloadWebsite()`, we load the form, then we wait for the `formWillAppear` event, then, when we receive that event, `preloadWebsite()` will return and the `IAFPresentationManager` will present the form. 

It's a little confusing because the `preloadWebsite()` implementation already exists as a default method in the `KlaviyoWebViewModeling` protocol (to which `IAFWebViewModel` conforms). But that default `preloadWebsite()` method was built with a generic website in mind, not an IAF. A generic website emits a `didFinishNavigation` event when it finishes loading, but our IAF is a special case and it emits `formWillAppear` when it finishes loading. So we should override the `preloadWebsite()` method in the IAFWebViewModel to specifically listen for the `formWillAppear` event.

By calling `presentIAF()` from within the ViewModel, as you do above, you're actually creating a new instance of the IAFWebViewModel and KlaviyoWebViewController, and you're waiting for the form to emit a `didFinishNavigation` event rather than a `formWillAppear` event. I'm actually surprised that this implementation worked, because I would think this would create an infinite loop (init the ViewModel and preload the website, when preload finishes then init a new ViewModel and preload the website, when preload finishes init a new ViewModel and preload, etc etc etc).

I think this all sounds confusing, but fortunately it's not too difficult to fix this and wait for `formWillAppear` instead of `didFinishNavigation`. I've put up a [suggestion PR](https://github.com/klaviyo/klaviyo-swift-sdk/pull/272) that shows you how it could work.

belleklaviyo:
At this moment, while testing, the SDK is still receiving the outdated `formAppeared` and not the updated `formWillAppear` -- so it does not actually successfully decode into the event (unless you change back all instances to `formAppeared`) but going ahead and keeping this in as I imagine the naming on the sending side will be updated soon

ajaysubra:
> At this moment, while testing, the SDK is still receiving the outdated `formAppeared` and not the updated `formWillAppear` -- so it does not actually successfully decode into the event (unless you change back all instances to `formAppeared`) but going ahead and keeping this in as I imagine the naming on the sending side will be updated soon

@timstephens-k is there a branch with the changed event names that Belle can use to test locally? Is it [this](https://github.com/klaviyo/fender/pull/37938)?

belleklaviyo:
found [this](https://github.com/klaviyo/fender/pull/37923) branch to test on and confirmed it is working with `formWillAppear`

ab1470 (approved):
Nice!

PR #273, opened 2025-02-12
fix preload timeout

ab1470:
# Description

This PR fixes an issue where the preloading timeout wasn't working on the IAFWebViewModel

# Check List

# Manual Test Plan

1. I manually simulated both a timeout and a successful `formWillAppear` event and validated that the code works as expected in both cases


belleklaviyo (approved):
nice thanks for fixing this!

PR #274, opened 2025-02-13
CHNL-16216: Implement abort and timeout

belleklaviyo:
# Description

~~Reused the existing timeout of 8.0 seconds~~ Defined new timeout const of 10 seconds (or 10_000_000_000 nanoseconds) and used where the viewModel preloads the website to catch any errors that occur

# Check List

# Manual Test Plan
1. In the case of the webview being initialized to preload the form and it times out/is unable to load it, the webview just never loads but logs the error. This was tested by giving an invalid src address. No webview is shown and app is usable as normal.

https://github.com/user-attachments/assets/285ee789-3429-43d0-a761-016d4f891f4b

2. In the case of the webview receiving an abort event at any other point (ex. after successfully presenting), the webview is destroyed and reason is logged, and the app is usable as normal with the webview gone.

https://github.com/user-attachments/assets/18e947bc-a192-4171-89d0-7ae6ea536c1b

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
@@ -52,7 +52,6 @@ public class IAFPresentationManager {
             do {
                 try await viewModel.preloadWebsite(timeout: 8_000_000_000)
             } catch {
-                isLoading = false
                 if #available(iOS 14.0, *) {
                     Logger.webViewLogger.warning("Error preloading In-App Form: \(error).")
                 }
--- a/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
@@ -49,7 +49,15 @@ public class IAFPresentationManager {
         Task {
             defer { isLoading = false }
 
-            try await viewModel.preloadWebsite(timeout: 8_000_000_000)
+            do {
+                try await viewModel.preloadWebsite(timeout: 8_000_000_000)
+            } catch {
+                isLoading = false
+                if #available(iOS 14.0, *) {
+                    Logger.webViewLogger.warning("Error preloading In-App Form: \(error).")
+                }
+                return
+            }
 
             guard let topController = UIApplication.shared.topMostViewController else {
                 return
--- a/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+import KlaviyoCore
 import OSLog
 import UIKit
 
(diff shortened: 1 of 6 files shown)


Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
```diff
@@ -49,7 +49,14 @@ public class IAFPresentationManager {
...
             defer { isLoading = false }
 
-            try await viewModel.preloadWebsite(timeout: 8_000_000_000)
+            do {
+                try await viewModel.preloadWebsite(timeout: 8_000_000_000)
+            } catch {
+                if #available(iOS 14.0, *) {
+                    Logger.webViewLogger.warning("Error preloading In-App Form: \(error).")
+                }
+                return
```
belleklaviyo, line 58: pretty sure any error would be caught here and no need to destroy the webview, just early return and never present it

Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
```diff
@@ -49,7 +49,14 @@ public class IAFPresentationManager {
         Task {
             defer { isLoading = false }
 
-            try await viewModel.preloadWebsite(timeout: 8_000_000_000)
+            do {
+                try await viewModel.preloadWebsite(timeout: 8_000_000_000)
```
kennyklaviyo, line 53: is this timeout the right value? what unit is this measured in?

dan-peluso: I'm still leaning using the Klaviyo 'networkTimeout' value for this (or maybe making a new one if we feel it's necessary)

belleklaviyo: @kennyklaviyo it's in nanoseconds so that's 8 seconds and @dan-peluso yeah sure, iOS doesn't have that timeout value already existing (everywhere that uses some value is just hardcoded) but I can fix that if we're good with 10 seconds

kennyklaviyo: https://github.com/klaviyo/klaviyo-android-sdk/blob/master/sdk/core/src/main/java/com/klaviyo/core/config/KlaviyoConfig.kt#L60
that makes sense then! if you make a constant, be sure to comment or renaming the var to indicate nanoseconds

Sources/KlaviyoUI/KlaviyoWebViewOverlayManager.swift
```diff
@@ -36,7 +36,7 @@ public class KlaviyoWebViewOverlayManager {
         Task {
             defer { isLoading = false }
 
-            try await viewModel.preloadWebsite(timeout: 8_000_000_000)
+            try await viewModel.preloadWebsite(timeout: 10_000_000_000) // 10 seconds
```
belleklaviyo, line 39: this presenter isn't actually used in practice, just for internal testing but just changed it too to keep it all the same

ajaysubra (approved):

ab1470 (approved):
one nit comment but otherwise looks great!

PR #275, opened 2025-02-14
CHNL-17239: Implement handshake with JS

belleklaviyo:
# Description
- injects the handshake array of events with its versions it supports
- builds out the IAFNativeBridgeEvents to support future versioning differences
- adds the `handShook` event (nothing to really handle just avoids a warning when receiving it)
- reordered events for tidiness
- fixes naming of `data-klaviyo-message-handler-name` to `data-native-bridge-name`

# Check List

# Manual Test Plan

Some things needed to test on this branch:
- change `company_id` in InAppFormsTemplate.html to local test account
- change where we are injecting `data-sdk-name` to `data-klaviyo-sdk-name` (to be reverted on fender side at some point)
- change where we are injecting `data-sdk-version` to `data-klaviyo-sdk-version` (to be reverted on fender side at some point)

Currently, the `handShook` event is emitted so long as there is at least one event included, doesn't matter which one. It just restricts what JS sends. Changing what cases are included in `allCases`, removing `formWillAppear` prevents the form from being presented. 

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoUI/InAppForms/Assets/InAppFormsTemplate.html
+++ b/Sources/KlaviyoUI/InAppForms/Assets/InAppFormsTemplate.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
-<head data-klaviyo-message-handler-name="KlaviyoNativeBridge"
+<head data-native-bridge-name="KlaviyoNativeBridge"
       data-support-event-types='test1,test2'
 >
     <meta charset="UTF-8">
--- a/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
@@ -28,13 +28,13 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
 
     private var sdkNameWKScript: WKUserScript {
         let sdkName = environment.sdkName()
-        let sdkNameScript = "document.head.setAttribute('data-klaviyo-sdk-name', '\(sdkName)');"
+        let sdkNameScript = "document.head.setAttribute('data-sdk-name', '\(sdkName)');"
         return WKUserScript(source: sdkNameScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
     }
 
     private var sdkVersionWKScript: WKUserScript {
         let sdkVersion = environment.sdkVersion()
-        let sdkVersionScript = "document.head.setAttribute('data-klaviyo-sdk-version', '\(sdkVersion)');"
+        let sdkVersionScript = "document.head.setAttribute('data-sdk-version', '\(sdkVersion)');"
         return WKUserScript(source: sdkVersionScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
     }
 
--- a/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
@@ -26,25 +26,32 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
     public let (navEventStream, navEventContinuation) = AsyncStream.makeStream(of: WKNavigationEvent.self)
     private let (formWillAppearStream, formWillAppearContinuation) = AsyncStream.makeStream(of: Void.self)
 
+    private var sdkNameWKScript: WKUserScript {
+        let sdkName = environment.sdkName()
+        let sdkNameScript = "document.head.setAttribute('data-klaviyo-sdk-name', '\(sdkName)');"
+        return WKUserScript(source: sdkNameScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
+    }
+
+    private var sdkVersionWKScript: WKUserScript {
+        let sdkVersion = environment.sdkVersion()
+        let sdkVersionScript = "document.head.setAttribute('data-klaviyo-sdk-version', '\(sdkVersion)');"
+        return WKUserScript(source: sdkVersionScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
+    }
+
+    private var handshakeWKScript: WKUserScript {
+        let handshakeStringified = IAFNativeBridgeEvent.handshake
+        let handshakeScript = "document.head.setAttribute('data-native-bridge-handshake', '\(handshakeStringified)');"
+        return WKUserScript(source: handshakeScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
+    }
+
     init(url: URL) {
         self.url = url
         initializeLoadScripts()
     }
 
     func initializeLoadScripts() {
-        let sdkName = environment.sdkName()
-        let sdkNameScript = "document.head.setAttribute('data-sdk-name', '\(sdkName)');"
-        let sdkNameWKScript = WKUserScript(source: sdkNameScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
         loadScripts?.insert(sdkNameWKScript)
-
-        let sdkVersion = environment.sdkVersion()
-        let sdkVersionScript = "document.head.setAttribute('data-sdk-version', '\(sdkVersion)');"
-        let sdkVersionWKScript = WKUserScript(source: sdkVersionScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
         loadScripts?.insert(sdkVersionWKScript)
-
-        let handshakeStringified = IAFNativeBridgeEvent.handshake
-        let handshakeScript = "document.head.setAttribute('data-native-bridge-handshake', '\(handshakeStringified)');"
-        let handshakeWKScript = WKUserScript(source: handshakeScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
         loadScripts?.insert(handshakeWKScript)
     }
 
--- a/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
@@ -41,6 +41,11 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
         let sdkVersionScript = "document.head.setAttribute('data-sdk-version', '\(sdkVersion)');"
         let sdkVersionWKScript = WKUserScript(source: sdkVersionScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
         loadScripts?.insert(sdkVersionWKScript)
+
+        let handshakeStringified = IAFNativeBridgeEvent.handshake
+        let handshakeScript = "document.head.setAttribute('data-native-bridge-handshake', '\(handshakeStringified)');"
+        let handshakeWKScript = WKUserScript(source: handshakeScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
+        loadScripts?.insert(handshakeWKScript)
     }
 
     func preloadWebsite(timeout: UInt64) async throws {
@@ -131,20 +136,20 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
         case .formWillAppear:
             formWillAppearContinuation.yield()
             formWillAppearContinuation.finish()
-        case let .trackAggregateEvent(data):
-            KlaviyoInternal.create(aggregateEvent: data)
+        case .formDisappeared:
+            Task {
+                await delegate?.dismiss()
+            }
         case let .trackProfileEvent(data):
             if let jsonEventData = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any],
                let metricName = jsonEventData["metric"] as? String {
                 KlaviyoSDK().create(event: Event(name: .customEvent(metricName), properties: jsonEventData))
             }
+        case let .trackAggregateEvent(data):
+            KlaviyoInternal.create(aggregateEvent: data)
         case let .openDeepLink(url):
             if UIApplication.shared.canOpenURL(url) {
                 UIApplication.shared.open(url)
-            }
-        case .formDisappeared:
-            Task {
-                await delegate?.dismiss()
             }
         case let .abort(reason):
             if #available(iOS 14.0, *) {
@@ -152,6 +157,10 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
             }
             Task {
                 await delegate?.dismiss()
+            }
+        case .handShook:
+            if #available(iOS 14.0, *) {
+                Logger.webViewLogger.info("Successful handshake with JS")
             }
         }
     }
--- a/Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
+++ b/Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
@@ -7,6 +7,7 @@
 
 import AnyCodable
 import Foundation
+import OSLog
 
 enum IAFNativeBridgeEvent: Decodable, Equatable {
     case formsDataLoaded
@@ -75,14 +76,14 @@ extension IAFNativeBridgeEvent {
     }
 }
 
-extension IAFNativeBridgeEvent: CaseIterable {
+extension IAFNativeBridgeEvent {
     public static var handshake: String {
         struct HandshakeData: Codable {
             var type: String
             var version: Int
         }
 
-        let handshakeArray = allCases.map { event -> HandshakeData in
+        let handshakeArray = handshakeEvents.map { event -> HandshakeData in
             HandshakeData(type: event.name, version: event.version)
         }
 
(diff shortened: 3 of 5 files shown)


Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -71,3 +74,65 @@ extension IAFNativeBridgeEvent {
...
+
+    public static var allCases: [IAFNativeBridgeEvent] {
+        // events that JS is permitted to sending
+        [
+            .formWillAppear,
+            .formDisappeared,
+            .trackProfileEvent(Data()),
+            .trackAggregateEvent(Data()),
+            .openDeepLink(URL(string: "https://example.com")!),
+            .abort("")
```
belleklaviyo, line 109: ordered this to be the same as Android/as documented in the message bus doc and reordered the cases in this file to be the same way

Sources/KlaviyoUI/InAppForms/Assets/InAppFormsTemplate.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
-<head data-klaviyo-message-handler-name="KlaviyoNativeBridge"
+<head data-native-bridge-name="KlaviyoNativeBridge"
```
belleklaviyo, line 3: got this from @dan-peluso -- don't personally know when this was changed? but testing it it does need to be `data-native-bridge-name`

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -71,3 +74,65 @@ extension IAFNativeBridgeEvent {
...
+        switch self {
+        case .formsDataLoaded: return 1
+        case .formWillAppear: return 1
+        case .formDisappeared: return 1
+        case .trackProfileEvent: return 1
+        case .trackAggregateEvent: return 1
+        case .openDeepLink: return 1
+        case .abort: return 1
+        case .handShook: return 1
+        }
```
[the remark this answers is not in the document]
ajaysubra: thoughts on maintaing a constant for the version like [this](https://github.com/klaviyo/klaviyo-swift-sdk/blob/master/Sources/KlaviyoCore/Utils/Version.swift) locally in KlaviyoForms

Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
```diff
@@ -41,6 +41,11 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
         let sdkVersionScript = "document.head.setAttribute('data-sdk-version', '\(sdkVersion)');"
         let sdkVersionWKScript = WKUserScript(source: sdkVersionScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
         loadScripts?.insert(sdkVersionWKScript)
+
```
ajaysubra, line 44: nit suggestion: can we make this a little modular like, 
```swift
func initializeLoadScripts() {
      loadScripts?.insert(sdkNameWKScript)
      loadScripts?.insert(sdkVersionWKScript)
      loadScripts?.insert(handshakeWKScript)
}
```
and each script is a computed var or a method to make reading this a little easier? Not a blocker by any means.

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -71,3 +74,65 @@ extension IAFNativeBridgeEvent {
...
+    }
+
+    public static var allCases: [IAFNativeBridgeEvent] {
+        // events that JS is permitted to sending
+        [
+            .formWillAppear,
+            .formDisappeared,
+            .trackProfileEvent(Data()),
+            .trackAggregateEvent(Data()),
+            .openDeepLink(URL(string: "https://example.com")!),
```
ajaysubra, line 108: does this string match with the url when comparing? Curious why the url?

ajaysubra (approved):

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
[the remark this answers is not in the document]
belleklaviyo: do you think different event cases could have different versions or just one version across the board

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
belleklaviyo: this string doesn't actually matter at all it's just the get the cases we care about listed but this value is never used. it needs to be passed in since the openDeepLink event expects a URL element

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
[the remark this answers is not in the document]
ajaysubra: I think just one. @evan-masseau can you confirm?

ajaysubra (approved):

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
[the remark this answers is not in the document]
evan-masseau: Individual event types could have different versions from each other in the future.

Sources/KlaviyoUI/InAppForms/Assets/InAppFormsTemplate.html
evan-masseau: It was when I wrote the [handshake spec](https://klaviyo.slack.com/archives/C064QGU3BCG/p1738684647925689?thread_ts=1738608755.736679&cid=C064QGU3BCG). I wanted to consolidate terminology, `Message Handler` and `Native Bridge` were being used interchangeably and its easier to have a single name for it in the long run

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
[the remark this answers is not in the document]
belleklaviyo: ^ if that's the case I'm personally leaning towards just keeping the versions to how I have them now to avoid duplicating all the cases again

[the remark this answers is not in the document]
evan-masseau: Yeah I didn't feel like extracting a constant for version number was appropriate when I wrote the android analogue to this.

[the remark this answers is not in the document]
ajaysubra: This make sense to me. Ignore the suggestion.

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -71,3 +74,65 @@ extension IAFNativeBridgeEvent {
         let reason: String
     }
 }
+
+extension IAFNativeBridgeEvent: CaseIterable {
```
ab1470, line 78: I think this isn't a good use-case for `CaseIterable`. From what I understand by looking at this code, it seems like there's a subset of events that get sent to the IAF HTML for the handshake. When using `CaseIterable`'s `allCases`, it usually should be used to reference "all cases" of the enum, not "a certain subset of cases that I need for a particular situation". 

There's a risk here by making `IAFNativeBridgeEvent` `CaseIterable`, because if somewhere down the line in another part of the code we want to actually access _all_ cases of the enum, we may use the `allCases` that you define here, without realizing that it's actually a subset and not _all_ cases.

There's another wrinkle with `CaseIterable` in this situation. Because some of the enum cases have associated values, to conform to `CaseIterable` and define the `allCases` array, you'll need to set the associated values for the cases that require them. You're doing this with the `allCases` var you set below. This means that, when someone asks for `allCases` (potentially in a different part of the code, as in the scenario I describe in the paragraph above), they're going to get cases with a hard-coded associated value. As you note in your [comment below](https://github.com/klaviyo/klaviyo-swift-sdk/pull/275/files#r1959949600), it may not matter in this particular PR that we're hard-coding "example.com" as an associated value, this could be a problem elsewhere in the code where we're asking for `allCases` and inexplicably getting "example.com" as an associated value.

More generally, using `CaseIterable` for enums with associated values can be tricky for the reason I describe above. If an enum case has a string or numeric associated value, then there are literally infinite cases for the enum (`SomeEnum.count(0)`, `SomeEnum.count(1)`, `SomeEnum.count(2)`, etc). So strictly speaking, it's not actually possible to name "all cases" of the enum. There can still potentially be reasons to conform an enum with associated values to `CaseIterable`, but there should be a good reason for it.

After that very long comment, the good news is that the solution in this case is simple. You can remove the `CaseIterable` conformance, then simply rename your `allCases` to something more descriptive for this particular situation (maybe `casesForHandshake`?). I think you can also make that `var` `private`.

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -71,3 +74,65 @@ extension IAFNativeBridgeEvent {
...
+        }
+
+        do {
+            let encoder = JSONEncoder()
+            let jsonData = try encoder.encode(handshakeArray)
+            if let jsonString = String(data: jsonData, encoding: .utf8) {
+                return jsonString
+            }
+        } catch {
+            print("Error encoding handshake data: \(error)")
```
ab1470, line 96:
```suggestion
            if #available(iOS 14.0, *) {
                Logger.webViewLogger.warning("Error encoding handshake data: \(error)")
            }
```

ab1470: you'll also need to `import OSLog` to make this work

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -71,3 +74,65 @@ extension IAFNativeBridgeEvent {
         let reason: String
     }
 }
+
+extension IAFNativeBridgeEvent: CaseIterable {
+    public static var handshake: String {
```
ab1470, line 79: I think there should probably be a unit test to validate that this outputs what you expect it to output

Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
ajaysubra: I will accept the compliment for Belle's work!

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
ab1470: love that consistency

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
belleklaviyo: got it, that all makes sense -- thanks for the explanation! will change

Sources/KlaviyoUI/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -71,3 +75,67 @@ extension IAFNativeBridgeEvent {
...
+            }
+        } catch {
+            if #available(iOS 14.0, *) {
+                Logger.webViewLogger.warning("Error encoding handshake data: \(error)")
+            }
+        }
+        return ""
+    }
+
+    private static var handshakeEvents: [IAFNativeBridgeEvent] {
```
ab1470, line 104: nice, better than the name I suggested in my comment haha

Tests/KlaviyoUITests/IAFNativeBridgeEventTests.swift
```diff
@@ -13,6 +13,26 @@ import Foundation
 import Testing
 
 struct IAFNativeBridgeEventTests {
+    @Test func testHandshakeCreated() async throws {
```
ab1470, line 16: 👍

ab1470 (commented):
thanks for those changes, looking good. One last thing: can you add another unit test to validate that the HTML attribute is successfully getting injected into the IAF HTML? It should be mostly a copy-paste of [this test](https://github.com/klaviyo/klaviyo-swift-sdk/blob/c9bdf25e65a9c575d1e30216dcfcaa156c2ac60b/Tests/KlaviyoUITests/IAFWebViewModelTests.swift#L45)

Tests/KlaviyoUITests/IAFNativeBridgeEventTests.swift
```diff
@@ -13,6 +13,26 @@ import Foundation
...
+        let expectedHandshake = """
+        [{"type":"formWillAppear","version":1},{"type":"formDisappeared","version":1},{"type":"trackProfileEvent","version":1},{"type":"trackAggregateEvent","version":1},{"type":"openDeepLink","version":1},{"type":"abort","version":1}]
+        """
+        #expect(IAFNativeBridgeEvent.handshake == expectedHandshake)
+    }
```
ab1470, line 21: so [this test is failing](https://github.com/klaviyo/klaviyo-swift-sdk/actions/runs/13397611557/attempts/1#summary-37420449247) because we're not guaranteed that when the JSONEncoder encodes the handshakeArray, the order of the keys matches what you hard-coded as the expected test result. For example, in the failing test, the JSONEncoder encoded one of the handshakeArray items as `{"version":1,"type":"formDisappeared"}`, but the expected result in your test is `{"type":"formDisappeared","version":1}`.

The solution I coded up is clunky but it works. If we decode the hard-coded test string and the `IAFNativeBridgeEvent.handshake` var into `HandshakeData` types, we can compare them and validate that they're equal. I had to create a new `HandshakeData` type here in this test, because the one you created in `IAFNativeBridgeEvent` isn't publicly accessible and I don't want to change that. 

```suggestion
    @Test func testHandshakeCreated() async throws {
        struct HandshakeData: Codable, Equatable {
            var type: String
            var version: Int
        }
        
        let expectedHandshake = """
        [
            { "type": "formWillAppear", "version": 1 },
            { "type": "formDisappeared", "version": 1 },
            { "type": "trackProfileEvent", "version": 1 },
            { "type": "trackAggregateEvent", "version": 1 },
            { "type": "openDeepLink", "version": 1 },
            { "type": "abort", "version": 1 }
        ]
        """
        
        let expectedData = try #require(expectedHandshake.data(using: .utf8))
        let expectedDataDecoded = try JSONDecoder().decode([HandshakeData].self, from: expectedData)
        
        let handshakeData = try #require(IAFNativeBridgeEvent.handshake.data(using: .utf8))
        let handshakeDataDecoded = try JSONDecoder().decode([HandshakeData].self, from: handshakeData)
        
        #expect(expectedDataDecoded == handshakeDataDecoded)
    }
```

belleklaviyo: lol jinx yeah was just fixing that and pushed something I think works that addresses that ^

ab1470: lol [good thinking](https://github.com/klaviyo/klaviyo-swift-sdk/pull/275/commits/7f2630892dd8915350e5f2207afe7927437f4b34). I should have asked you if you were working on a solution before I coded one up myself.

Tests/KlaviyoUITests/IAFWebViewModelTests.swift
```diff
@@ -79,4 +79,27 @@ final class IAFWebViewModelTests: XCTestCase {
...
+        // When
+        let script = "document.head.getAttribute('data-native-bridge-handshake');"
+        let delegate = try XCTUnwrap(viewModel.delegate)
+        let result = try await delegate.evaluateJavaScript(script)
+        let resultString = try XCTUnwrap(result as? String)
+
+        // Then
+        let handshakeString =
+            """
+            [{"type":"formWillAppear","version":1},{"type":"formDisappeared","version":1},{"type":"trackProfileEvent","version":1},{"type":"trackAggregateEvent","version":1},{"type":"openDeepLink","version":1},{"type":"abort","version":1}]
```
ab1470, line 101: ugh, unfortunately you're going to have the same issue here as you already had [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/275#discussion_r1960403830). Sorry, these tests are annoying 🤷‍♂️

belleklaviyo: ahhh good catch, my bad -- since this test is being skipped by CI I didn't think about fixing it here too

ab1470: nice, thanks for fixing

ab1470 (approved):
🪩🪩🪩

PR #278, opened 2025-02-18
[CHNL-17241] make IAF HTML company agnostic

ab1470:
# Description

This PR dynamically injects the KlaviyoJS load script into the IAF HTML when it's loaded into the WKWebView. Before injecting the script, this code will assemble the script's `src` URL using the URL host and Company ID from the SDK's environment.

# Check List

# Manual Test Plan

As validation that the script is successfully injected when the IAF WebView loads, I used Safari to inspect the form. As you can see, the KlaviyoJS script appears in the DOM:

Screenshot 2025-02-18 at 14 42 29

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoInternal.swift
+++ b/Sources/KlaviyoSwift/KlaviyoInternal.swift
@@ -11,6 +11,9 @@ import KlaviyoCore
 ///
 /// - Note: Can only be accessed from other modules within the Klaviyo-Swift-SDK package; cannot be accessed from the host app.
 package struct KlaviyoInternal {
+    /// the apiKey (a.k.a. CompanyID) for the current SDK instance.
+    package static var apiKey: String? { klaviyoSwiftEnvironment.state().apiKey }
+
     /// Create and send an aggregate event.
     /// - Parameter event: the event to be tracked in Klaviyo
     package static func create(aggregateEvent: AggregateEventPayload) {
--- a/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
@@ -28,6 +28,7 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
 
     private var klaviyoJsWKScript: WKUserScript? {
         guard let companyId = KlaviyoInternal.apiKey else {
+            environment.emitDeveloperWarning("SDK must be initialized before usage.")
             if #available(iOS 14.0, *) {
                 Logger.webViewLogger.warning("Unable to initialize KlaviyoJS script on In-App Form HTML due to missing API key.")
             }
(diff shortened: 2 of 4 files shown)


belleklaviyo (approved):

Sources/KlaviyoSwift/KlaviyoInternal.swift
```diff
@@ -11,6 +11,9 @@ import KlaviyoCore
 ///
 /// - Note: Can only be accessed from other modules within the Klaviyo-Swift-SDK package; cannot be accessed from the host app.
 package struct KlaviyoInternal {
+    /// the apiKey (a.k.a. CompanyID) for the current SDK instance.
```
ajaysubra, line 14: Given the async nature of [initialize](https://github.com/klaviyo/klaviyo-swift-sdk/blob/e12f7c645db48dbf14264f5963d8d196631372a6/Sources/KlaviyoSwift/StateManagement/StateManagement.swift#L135) are there any risks with this value not being in state when accessed from `KlaviyoForms`? Probably not but just confirming? 

Looked at the code and looks like the api key is set to state before the async operation starts so maybe we should be fine?


https://github.com/klaviyo/klaviyo-swift-sdk/blob/e12f7c645db48dbf14264f5963d8d196631372a6/Sources/KlaviyoSwift/StateManagement/StateManagement.swift#L135

Sources/KlaviyoUI/InAppForms/IAFWebViewModel.swift
```diff
@@ -26,6 +26,32 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
     public let (navEventStream, navEventContinuation) = AsyncStream.makeStream(of: WKNavigationEvent.self)
     private let (formWillAppearStream, formWillAppearContinuation) = AsyncStream.makeStream(of: Void.self)
 
+    private var klaviyoJsWKScript: WKUserScript? {
+        guard let companyId = KlaviyoInternal.apiKey else {
+            if #available(iOS 14.0, *) {
+                Logger.webViewLogger.warning("Unable to initialize KlaviyoJS script on In-App Form HTML due to missing API key.")
```
ajaysubra, line 32: Does it make sense to emit a developer warning here like - https://github.com/klaviyo/klaviyo-swift-sdk/blob/e12f7c645db48dbf14264f5963d8d196631372a6/Sources/KlaviyoSwift/StateManagement/StateManagement.swift#L130

ajaysubra (approved):
LGTM 
* 1 question that I just wanted your thoughts on 
* one minor developer experience comment

Sources/KlaviyoSwift/KlaviyoInternal.swift
ab1470: > Looked at the code and looks like the api key is set to state before the async operation starts so maybe we should be fine?

thanks for looking into that. Yeah, looks like the API key is set immediately when the action is called, so we should be good.

evan-masseau: oh nice, that's one less thing to worry about

PR #279, opened 2025-02-19
Fix retain cycle on consoleMessageHandler

belleklaviyo:
# Description
When testing, tapping "Show In-App Form Test" button multiple times created a new ViewModel and ViewController each time. Seems to be this consoleMessageHandler that's causing the strong retain cycle

Before:
image

After:

image
 

# Check List

# Manual Test Plan

1.

# Supporting Materials


ab1470 (approved):
nice find!

ajaysubra (approved):

PR #280, opened 2025-02-20
[CHNL-16714] add registerForInAppForms

ab1470:
# Description

This PR adds a `registerForInAppForms` method that, when called, will present the IAF web view and display a form, if one is available.

# Check List

# Manual Test Plan

I wired up the "Show In-App Form Test" button in the iOS test app to the new `registerForInAppForms` method and validated that it works as expected


ajaysubra (approved):
nice!!!

belleklaviyo (approved):

PR #283, opened 2025-02-20
CHNL-16714: Prevent multiple native view from being shown simultaneously

belleklaviyo:
# Description
Part of https://klaviyo.atlassian.net/browse/CHNL-17524

Added guard to check there isn't already a KlaviyoWebViewController being presented

# Check List

# Manual Test Plan
When trying to call `presentIAF()` multiple times, the `!isLoading` guard already prevented multiple presentation a lot of the times, so this is just an extra level of protection. To test this new guard, I commented out the `!isLoading` guard and rapidly clicked the show in app form button. Before it was possible to trigger multiple native views one after another. Now this is prevented.

Before:

https://github.com/user-attachments/assets/b5659741-eccb-4cdd-9161-c1d2fdee1a84

After:

https://github.com/user-attachments/assets/85c51c62-1fac-439e-85ea-2442c9f45f49

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
@@ -61,11 +61,9 @@ class IAFPresentationManager {
                 return
             }
 
-            let controllersInStack = topController.navigationController?.viewControllers ?? []
-
-            if topController.isKlaviyoVC || controllersInStack.contains(where: \.isKlaviyoVC) {
+            if topController.isKlaviyoVC || topController.hasKlaviyoVCInStack {
                 if #available(iOS 14.0, *) {
-                    Logger.webViewLogger.warning("In-App Form is already presenting; ignoring request")
+                    Logger.webViewLogger.warning("In-App Form is already being presented; ignoring request")
                 }
             } else {
                 topController.present(viewController, animated: true, completion: nil)
@@ -77,5 +75,12 @@ class IAFPresentationManager {
 extension UIViewController {
     fileprivate var isKlaviyoVC: Bool {
         self is KlaviyoWebViewController
+    }
+
+    fileprivate var hasKlaviyoVCInStack: Bool {
+        guard let navigationController = navigationController else {
+            return false
+        }
+        return navigationController.viewControllers.contains(where: \.isKlaviyoVC)
     }
 }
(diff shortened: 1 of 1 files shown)


Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
```diff
@@ -60,7 +60,17 @@ class IAFPresentationManager {
             guard let topController = UIApplication.shared.topMostViewController else {
                 return
             }
-            topController.present(viewController, animated: true, completion: nil)
+
+            let controllersInStack = topController.navigationController?.viewControllers ?? []
+
+            if topController is KlaviyoWebViewController || controllersInStack.contains(where: { $0 is KlaviyoWebViewController }) {
```
belleklaviyo, line 66: this felt safe to do as its checking both the current highest VC as well as any other ones in the scenario an app has another thing presented in front (seems edge casey)

ajaysubra: Nit: can we create a private extension on `UIViewController` and call it something like `isKlaviyoVC` or some better name and then this logic can be simplified to `topController .isKlaviyoVC`?

Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
```diff
@@ -60,7 +60,17 @@ class IAFPresentationManager {
...
+                if #available(iOS 14.0, *) {
+                    Logger.webViewLogger.warning("In-App Form is already presenting; ignoring request")
+                }
+                return
```
ajaysubra, line 70: do we need this return if we have the else?

ajaysubra (approved):

Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
belleklaviyo: ah my bad originally had this as a guard, thanks for catching that

Sources/KlaviyoUI/InAppForms/IAFPresentationManager.swift
```diff
@@ -60,7 +60,22 @@ class IAFPresentationManager {
             guard let topController = UIApplication.shared.topMostViewController else {
                 return
             }
-            topController.present(viewController, animated: true, completion: nil)
+
+            let controllersInStack = topController.navigationController?.viewControllers ?? []
+
+            if topController.isKlaviyoVC || controllersInStack.contains(where: \.isKlaviyoVC) {
```
ajaysubra, line 66: I think thinking `controllersInStack` can also live in the extension

ab1470 (approved):
nice!

PR #286, opened 2025-02-21
Swap `a.klaviyo` to `static.klaviyo`

belleklaviyo:
# Description

Per discussion [here](https://klaviyo.slack.com/archives/C064QGU3BCG/p1740158034111069)

A regression led to `a.klaviyo` being able to evade the VPN for using `assetSource` but to be safe we should use `static.klaviyo` as that is what android is using too (rip life hack)

Added logic so for local envs it still uses both `localhost:8080` to load the IAF as well as route API requests but for prod envs it uses `static.klaviyo` to load the IAF and `a.klaviyo` to route API requests

https://github.com/user-attachments/assets/f3dc24ca-1aeb-4add-94d7-04d03ab624d5

# Check List

# Manual Test Plan

1.

# Supporting Materials


ab1470 (approved):

Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -72,7 +72,7 @@ public struct KlaviyoEnvironment {
     static let productionHost: URLComponents = {
```
evan-masseau, line 72: Is this also used by API traffic?

evan-masseau:
> (rip life hack)

I feel like such a party pooper

evan-masseau (changes requested):
[FYI](https://klaviyo.slack.com/archives/C064QGU3BCG/p1740169779152309?thread_ts=1740158034.111069&cid=C064QGU3BCG)

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -41,7 +41,7 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
             return nil
         }
 
-        var apiURL = environment.apiURL()
+        var apiURL = environment.cdnURL()
```
ajaysubra, line 44: what happens when you want to point to local from the test app?

belleklaviyo: yeah realized this doesn't work, fixing this

Sources/KlaviyoCore/Networking/PrivateMethods.swift
```diff
@@ -14,5 +14,12 @@ import Foundation
 public func overrideSDKDefaults(urlComponents: URLComponents? = nil) {
     if let urlComponents {
         environment.apiURL = { urlComponents }
+
+        var cdnURLComponents = urlComponents
```
ajaysubra, line 18: this doesn't need to be called `cdnURLComponents` no for like client api calls.

ajaysubra (approved):
minor nit pick but looks good to me

Tests/KlaviyoCoreTests/TestUtils.swift
```diff
@@ -95,6 +95,7 @@ extension KlaviyoEnvironment {
             emitDeveloperWarning: { _ in },
             networkSession: { NetworkSession.test() },
             apiURL: { URLComponents(string: "https://dead_beef")! },
+            cdnURL: { URLComponents(string: "https://dead_beef")! },
```
evan-masseau, line 98: nit: I'd avoid using the same value for test stubs, else you could mix these properties up and a test won't catch it.

Tests/KlaviyoSwiftTests/KlaviyoTestUtils.swift
```diff
@@ -44,6 +44,7 @@ extension KlaviyoEnvironment {
             emitDeveloperWarning: { _ in },
             networkSession: { NetworkSession.test() },
             apiURL: { URLComponents(string: "https://dead_beef")! },
+            cdnURL: { URLComponents(string: "https://dead_beef")! },
```
evan-masseau, line 47: same as above

evan-masseau (approved):
LGTM, but could you just use distinct values in test environment to avoid a mixup down the line going undetected

PR #293, opened 2025-02-24
(Do Not Merge) rel/4.2.0

belleklaviyo:
# Description

4.2.0 version bump


ajaysubra:
> Do we not have a podspec for `KlaviyoForms`?

May be it doesn't need to be updated! Ignore if that's the case

evan-masseau (approved):

KlaviyoForms.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = "KlaviyoForms"
-  s.version          = "0.1.0"
+  s.version          = "4.2.0"
```
dan-peluso, line 3: This is certainly a discussion outside the context of this pr, but I'm personally not a fan of correlating the version of this product to that of the rest of the SDK (mainly bc it's a different product, it could also maybe give the wrong impression that we've added functionality in other aspects of the app).

dan-peluso (approved):

KlaviyoForms.podspec
ajaysubra: Yeah that's a good point and also in the future when we just make updates to forms we'd have to update the other modules as well. it gets worse when it's a breaking change but I think in terms of looking from out (dev point of view) inwards there is an argument to be made that devs only care about our SDK as a whole vs individual parts. For now keeping it in lock step but let's talk internally at some point.

PR #297, opened 2025-02-27
Minor refactor

ajaysubra:
# Description

Was reading the code and tried to make some suggestion to make readability better. Logging was hindering code readability so made it one line to skim over it easily. Also, removed some unused code for web view delegates that we don't need. 

Planning to not get this in since we've already done a good amount of testing but plan to get in post release. 

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -26,16 +26,12 @@ class IAFPresentationManager {
     @MainActor
     func presentIAF(assetSource: String? = nil) {
         guard !isLoading else {
-            if #available(iOS 14.0, *) {
-                Logger.webViewLogger.log("In-App Form is already loading; ignoring request.")
-            }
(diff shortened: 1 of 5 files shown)


Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -51,33 +47,30 @@ class IAFPresentationManager {
...
+    fileprivate var containsKlaviyoForm: Bool {
+        isKlaviyoVC || hasKlaviyoVCInStack
+    }
+
+    private var isKlaviyoVC: Bool {
         self is KlaviyoWebViewController
     }
 
-    fileprivate var hasKlaviyoVCInStack: Bool {
+    private var hasKlaviyoVCInStack: Bool {
```
ndurell, line 73: was this fileprivate for a reason?

ajaysubra: swiftlint did this 🤣

PR #299, opened 2025-02-28
WIP

belleklaviyo:
# Description

This works and always ensures we have the companyId when we call registerForInAppForms but it is currently getting called way too many times WIP

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoInternal.swift
+++ b/Sources/KlaviyoSwift/KlaviyoInternal.swift
@@ -21,10 +21,9 @@ package struct KlaviyoInternal {
         if cancellable == nil {
             cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
                 .receive(on: DispatchQueue.main)
-                .eraseToAnyPublisher()
-                .filter { $0.initalizationState == .initialized }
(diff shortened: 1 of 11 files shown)


Sources/KlaviyoSwift/KlaviyoInternal.swift
```diff
@@ -5,14 +5,38 @@
...
+    package static func apiKey(completion: ((String?) -> Void)? = nil) {
+        print("apiKey call")
+        if cancellable == nil {
+            cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
+                .receive(on: DispatchQueue.main)
+                .eraseToAnyPublisher()
+                .filter { $0.initalizationState == .initialized }
+                .filter { $0.apiKey != nil }
+                .map(\.apiKey!)
+                .sink(receiveCompletion: { _ in
```
ndurell, line 28: Add distinctuntilchanged before this.

ajaysubra: looks like it's `removeDuplicates()`

PR #304, opened 2025-03-03
Fix memory leak + cancellable issue

belleklaviyo:
# Description

Changing where we are creating the KlaviyoWebViewController to only when we absolutely know we are presenting something fixes a memory leak. And using `.prefix(1)` helps us avoid having to manually cancel or check for nil of the cancellable.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoInternal.swift
+++ b/Sources/KlaviyoSwift/KlaviyoInternal.swift
@@ -17,15 +17,18 @@ package struct KlaviyoInternal {
     /// the apiKey (a.k.a. CompanyID) for the current SDK instance.
     /// - Parameter completion: completion hanlder that will be called when apiKey is avaialble after SDK is initilized
     package static func apiKey(completion: @escaping ((String?) -> Void)) {
-        cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
-            .receive(on: DispatchQueue.main)
-            .filter { $0.initalizationState == .initialized }
-            .compactMap(\.apiKey)
-            .removeDuplicates()
-            .prefix(1)
-            .sink(receiveValue: {
-                completion($0)
-            })
+        if cancellable == nil {
+            cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
+                .receive(on: DispatchQueue.main)
+                .filter { $0.initalizationState == .initialized }
+                .compactMap(\.apiKey)
+                .removeDuplicates()
+                .prefix(1)
+                .sink(receiveValue: {
+                    completion($0)
+                    cancellable = nil
+                })
+        }
     }
 
     /// Create and send an aggregate event.
--- a/Sources/KlaviyoSwift/KlaviyoInternal.swift
+++ b/Sources/KlaviyoSwift/KlaviyoInternal.swift
@@ -17,16 +17,16 @@ package struct KlaviyoInternal {
     /// the apiKey (a.k.a. CompanyID) for the current SDK instance.
     /// - Parameter completion: completion hanlder that will be called when apiKey is avaialble after SDK is initilized
     package static func apiKey(completion: @escaping ((String?) -> Void)) {
-        if cancellable == nil {
-            cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
-                .receive(on: DispatchQueue.main)
-                .filter { $0.initalizationState == .initialized }
-                .compactMap(\.apiKey)
-                .removeDuplicates()
-                .sink(receiveValue: {
+        cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
+            .receive(on: DispatchQueue.main)
+            .filter { $0.initalizationState == .initialized }
+            .compactMap(\.apiKey)
+            .removeDuplicates()
+            .prefix(1)
+            .sink(
+                receiveValue: {
                     completion($0)
                 })
-        }
     }
 
     /// Create and send an aggregate event.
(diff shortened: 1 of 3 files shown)


Sources/KlaviyoSwift/KlaviyoInternal.swift
```diff
@@ -17,16 +17,16 @@ package struct KlaviyoInternal {
...
-                .removeDuplicates()
-                .sink(receiveValue: {
+        cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
+            .receive(on: DispatchQueue.main)
+            .filter { $0.initalizationState == .initialized }
+            .compactMap(\.apiKey)
+            .removeDuplicates()
+            .prefix(1)
+            .sink(
+                receiveValue: {
```
ajaysubra, line 27: is this formatted correctly?

ajaysubra: Nit:  can we move `receiveValue` in line with sink and have the closing line up with sink?

Sources/KlaviyoSwift/KlaviyoInternal.swift
```diff
@@ -17,16 +17,16 @@ package struct KlaviyoInternal {
...
     /// - Parameter completion: completion hanlder that will be called when apiKey is avaialble after SDK is initilized
     package static func apiKey(completion: @escaping ((String?) -> Void)) {
-        if cancellable == nil {
-            cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
-                .receive(on: DispatchQueue.main)
-                .filter { $0.initalizationState == .initialized }
-                .compactMap(\.apiKey)
-                .removeDuplicates()
-                .sink(receiveValue: {
+        cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
```
ndurell, line 20: wait do you still want the cancel first?

Sources/KlaviyoSwift/KlaviyoInternal.swift
```diff
@@ -17,16 +17,16 @@ package struct KlaviyoInternal {
...
+            .receive(on: DispatchQueue.main)
+            .filter { $0.initalizationState == .initialized }
+            .compactMap(\.apiKey)
+            .removeDuplicates()
+            .prefix(1)
```
ndurell, line 25: I think there is first too...not sure if that completes.

Sources/KlaviyoSwift/KlaviyoInternal.swift
ndurell: Uhh can you?

ndurell: Does completion give you the value?

belleklaviyo: swift lint 😞

Sources/KlaviyoSwift/KlaviyoInternal.swift
belleklaviyo: I don't think its necessary with the `.prefix(1)` that effectively handles it

ndurell: Ok if this was called may times in a row would you them all to run?

Sources/KlaviyoSwift/KlaviyoInternal.swift
ajaysubra: image

ndurell: Yah was just more curious what the difference was. I believe you it completes lol.

Sources/KlaviyoSwift/KlaviyoInternal.swift
ajaysubra: yeah there is a chance of that may be no harm in adding in the check tbh

Sources/KlaviyoSwift/KlaviyoInternal.swift
ajaysubra: lol I bring receipts always!

Sources/KlaviyoSwift/KlaviyoInternal.swift
ajaysubra: added the check back and also assigned cancellable to nil mainly because, 
1. when called multiple times it is possible that new subscriptions are created and this is not ideal
2. at any given point we only want to have one subscription in `KlaviyoInternal`

This solution is still not perfect since, 
1. In the case where subscription isn't nil and this method gets called, we don't call the completion handler which will lead to - `SWIFT TASK CONTINUATION MISUSE: presentIAF(assetSource:) leaked its continuation!`
2. We want to call the completion handler when the SDK isn't initialized because API key isn't set. You can add a timeout and do this but decided not to add more code to this since it's very close to release.


```swift
    package static func apiKey(completion: @escaping ((String?) -> Void)) {
        if cancellable == nil {
            cancellable = KlaviyoSwiftEnvironment.production.statePublisher()
                .receive(on: DispatchQueue.main)
                .filter { $0.initalizationState == .initialized }
                .compactMap(\.apiKey)
                .removeDuplicates()
                .prefix(1)
                .timeout(.seconds(1), scheduler: DispatchQueue.main)
                .sink(receiveCompletion: { _ in
                    completion(nil)
                    cancellable = nil
                }, receiveValue: { apiKey in
                    completion(apiKey)
                    cancellable = nil
                })
        }
    }
```

ndurell (approved):

PR #305, opened 2025-03-05
Fixing timeout

ajaysubra:
# Description

In #304 we moved the creation of the native view (View controller aka VC) containing the web view closer to the point when we are presenting the VC. This was done to avoid creation of the VC until needed and avoid a memory leak when IAFs is registered multiple times. What this did was, the web view wasn't preloading since `viewModel.preloadWebsite(timeout: NetworkSession.networkTimeout)` essentially uses the delegation pattern and calls into the VC where the web view is loaded headless. And since we didn't have a VC instance that code just did nothing and hence the timeout logic never got executed.

In this PR, I'm moving the VC before preload and hence avoid the timeout issue. There is still the VC which needs a larger refactor to fix but will be patched soon.

# Check List

# Manual Test Plan

1.

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -50,6 +50,8 @@ class IAFPresentationManager {
             }
 
             let viewModel = IAFWebViewModel(url: fileUrl, companyId: companyId, assetSource: assetSource)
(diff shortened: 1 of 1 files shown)


evan-masseau (approved):

dan-peluso (approved):

ajaysubra:
Fixed in #306

PR #317, opened 2025-03-24
Trim whitespace for create profile path

belleklaviyo:
# Description

#314 only fixed if we were making a token request w/ an identifier, but it didn't actually trim the whitespace for the create profile path. This ensures the identifiers are always trimmed.

# Check List

# Manual Test Plan

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
+++ b/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
@@ -22,9 +22,9 @@ extension Profile {
         externalId: String? = nil,
         anonymousId: String) -> ProfilePayload {
         ProfilePayload(
-            email: self.email?.trimWhiteSpaceOrReturnNilIfEmpty(),
-            phoneNumber: self.phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty(),
-            externalId: self.externalId?.trimWhiteSpaceOrReturnNilIfEmpty(),
+            email: email?.trimWhiteSpaceOrReturnNilIfEmpty() ?? self.email?.trimWhiteSpaceOrReturnNilIfEmpty(),
+            phoneNumber: phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty() ?? self.phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty(),
+            externalId: externalId?.trimWhiteSpaceOrReturnNilIfEmpty() ?? self.externalId?.trimWhiteSpaceOrReturnNilIfEmpty(),
             firstName: firstName,
             lastName: lastName,
             organization: organization,
--- a/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
+++ b/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
@@ -9,7 +9,7 @@ import Foundation
 import KlaviyoCore
 
 extension String {
-    fileprivate func returnNilIfEmpty() -> String? {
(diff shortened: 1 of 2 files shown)


Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -22,9 +22,9 @@ extension Profile {
         externalId: String? = nil,
         anonymousId: String) -> ProfilePayload {
         ProfilePayload(
-            email: email ?? self.email?.returnNilIfEmpty(),
-            phoneNumber: phoneNumber ?? self.phoneNumber?.returnNilIfEmpty(),
-            externalId: externalId ?? self.externalId?.returnNilIfEmpty(),
+            email: self.email?.trimWhiteSpaceOrReturnNilIfEmpty(),
```
ndurell, line 25: Are these functionally these or did we change the behavior?

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -22,9 +22,9 @@ extension Profile {
...
+            phoneNumber: self.phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty(),
+            externalId: self.externalId?.trimWhiteSpaceOrReturnNilIfEmpty(),
```
ab1470, line 27: I think you're changing the behavior. With this update, looks like you're ignoring the parameters in the function's input

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
belleklaviyo: its what it should've been functionally, but it's just always been broken (I think?) -- before, when an identifier was passed in, like `foo@blob.com    ` because its not nil, it wouldn't fall to the `returnNilIfEmpty()` trimmer -- so now we are just making sure it's always getting trimmed

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
belleklaviyo: ah you're right ok its not this I'll return this to draft and fix

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -22,9 +22,9 @@ extension Profile {
         externalId: String? = nil,
         anonymousId: String) -> ProfilePayload {
         ProfilePayload(
-            email: email ?? self.email?.returnNilIfEmpty(),
-            phoneNumber: phoneNumber ?? self.phoneNumber?.returnNilIfEmpty(),
-            externalId: externalId ?? self.externalId?.returnNilIfEmpty(),
+            email: email?.trimWhiteSpaceOrReturnNilIfEmpty() ?? self.email?.trimWhiteSpaceOrReturnNilIfEmpty(),
```
ndurell, line 25: ahh that makes more sense!

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -22,9 +22,9 @@ extension Profile {
...
+            phoneNumber: phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty() ?? self.phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty(),
```
ndurell, line 26: If we can update a test so it captures this it would be great!

ab1470 (approved):

PR #318, opened 2025-03-25
Trim whitespace upstream

belleklaviyo:
# Description

This should be the last whitespace change (I hope) -- @ab1470 ran into a path where if you create a profile with a identifier with trailling whitespace, that request itself has the identifier trimmed, but then when firing an event, that still holds the old identifier with trailing whitespace. There are several places where these identifiers are being pulled from state and then trimmed, but to avoid duplicating it everywhere we should just trim it upstream wherever it is being set and stored in state.

`updateStateWithProfile` is called when create profile is called
`updateEmail/PhoneNumber/ExternalId` is called when setEmail/PhoneNumber/ExternalId is called

Removed some tests that were no longer relevant as they tested things that would never happen.

# Check List

# Manual Test Plan

https://github.com/user-attachments/assets/8f4ff697-82bd-4257-b588-58842bfa1fc3

# Supporting Materials

### DIFF
--- a/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
+++ b/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
@@ -9,9 +9,8 @@ import Foundation
 import KlaviyoCore
 
 extension String {
-    fileprivate func trimWhiteSpaceOrReturnNilIfEmpty() -> String? {
-        let trimmedString = trimmingCharacters(in: .whitespacesAndNewlines)
-        return trimmedString.isEmpty ? nil : trimmedString
+    fileprivate func returnNilIfEmpty() -> String? {
+        trimmingCharacters(in: .whitespacesAndNewlines).isEmpty ? nil : self
     }
 }
 
@@ -22,9 +21,9 @@ extension Profile {
         externalId: String? = nil,
         anonymousId: String) -> ProfilePayload {
         ProfilePayload(
-            email: email?.trimWhiteSpaceOrReturnNilIfEmpty() ?? self.email?.trimWhiteSpaceOrReturnNilIfEmpty(),
-            phoneNumber: phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty() ?? self.phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty(),
-            externalId: externalId?.trimWhiteSpaceOrReturnNilIfEmpty() ?? self.externalId?.trimWhiteSpaceOrReturnNilIfEmpty(),
+            email: email ?? self.email?.returnNilIfEmpty(),
+            phoneNumber: phoneNumber ?? self.phoneNumber?.returnNilIfEmpty(),
+            externalId: externalId ?? self.externalId?.returnNilIfEmpty(),
             firstName: firstName,
             lastName: lastName,
             organization: organization,
--- a/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
+++ b/Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
@@ -9,8 +9,9 @@ import Foundation
 import KlaviyoCore
 
 extension String {
-    fileprivate func returnNilIfEmpty() -> String? {
-        trimmingCharacters(in: .whitespacesAndNewlines).isEmpty ? nil : self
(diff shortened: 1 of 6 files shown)


Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
```diff
@@ -9,9 +9,8 @@ import Foundation
 import KlaviyoCore
 
 extension String {
-    fileprivate func trimWhiteSpaceOrReturnNilIfEmpty() -> String? {
-        let trimmedString = trimmingCharacters(in: .whitespacesAndNewlines)
-        return trimmedString.isEmpty ? nil : trimmedString
+    fileprivate func returnNilIfEmpty() -> String? {
+        trimmingCharacters(in: .whitespacesAndNewlines).isEmpty ? nil : self
```
belleklaviyo, line 13: this is returning this to what it originally was some 2 PRs ago

Tests/KlaviyoSwiftTests/APIRequestErrorHandlingTests.swift
```diff
@@ -106,28 +106,6 @@ class APIRequestErrorHandlingTests: XCTestCase {
...
-        await store.receive(.deQueueCompletedResults(request), timeout: TIMEOUT_NANOSECONDS) {
-            $0.requestsInFlight = []
-            $0.flushing = false
-            $0.email = "foo@blob.com      "
-            if case let .registerPushToken(payload) = request.endpoint {
-                XCTAssertEqual(payload.data.attributes.profile.data.attributes.email, "foo@blob.com", "Email should be trimmed of whitespace before being sent.")
-            }
-        }
-    }
-
```
belleklaviyo, line 130: this test is no longer relevant as it injects an email into the request payload -- this is not an actual, possible path from the actual setters

Tests/KlaviyoSwiftTests/KlaviyoModelsTest.swift
```diff
@@ -90,26 +90,4 @@ class KlaviyoModelsTest: XCTestCase {
...
-            phoneNumber: "+12345678901    ",
-            externalId: "a       ",
-            anonymousId: anonymousId)
-
-        XCTAssertEqual(apiProfile.attributes.email, "test@blob.com")
-        XCTAssertEqual(apiProfile.attributes.phoneNumber, "+12345678901")
-        XCTAssertEqual(apiProfile.attributes.externalId, "a")
-        XCTAssertEqual(apiProfile.attributes.firstName, profile.firstName)
-        XCTAssertEqual(apiProfile.attributes.lastName, profile.lastName)
-        XCTAssertEqual(apiProfile.attributes.anonymousId, anonymousId)
```
belleklaviyo, line 113: same thing here, the props going to `toAPIModel` would come from state which from the setters would already have the trimmed value

njparadis (approved):

Tests/KlaviyoSwiftTests/APIRequestErrorHandlingTests.swift
ndurell: It might be possible if state were loaded from an old sdk right?

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
ab1470: https://github.com/klaviyo/klaviyo-swift-sdk/pull/318/files#r2012283619 good point, @ndurell 

I'm thinking that, to be safe, we leave this as it was and we trim whitespaces when creating the `ProfilePayload`. @ndurell, what do you think?

ndurell: yah might be better...sorry I haven't been more help here...I think I'm still a little lost on what we are trying to fix.

ab1470: cmd-shift-z cmd-shift-z

ab1470 (approved):

Sources/KlaviyoSwift/Models/ProfileAPIExtension.swift
ab1470: @belleklaviyo sorry for leading you astray on this one!

PR #320, opened 2025-03-28
Updated README for Silent Push

jasper-hsu-kyvo:
# Description

Updated the README for Silent Push. This will not be merged in until we are ready - just throwing it up.

# Check List

# Manual Test Plan

1.

# Supporting Materials


README.md
```diff
@@ -514,6 +515,11 @@ By default, the Klaviyo SDK automatically clears the badge count on app open. If
 
 Klaviyo SDK will automatically handle the badge count associated with Klaviyo pushes. If you need to manually update the badge count to account for other notification sources, use the `KlaviyoSDK().setBadgeCount(:)` method, which will update the badge count and keep it in sync with the Klaviyo SDK. This method should be used instead of (rather than in addition to) setting the badge count using `UNUserNotificationCenter` and/or `UIApplication` methods.
 
+#### Silent Push Notifications
+
+Silent push notifications (also known as background pushes) allow your app to receive payloads from Klaviyo without displaying a visible alert to the user. These are typically used to trigger background behavior, such as displaying content, personalizing the app interface, or downloading new information from a server.
+>  ℹ️ Silent push support is available by default. No special SDK methods are required to receive them, but your app must enable `Remote Notifications` under `Background Modes`.
```
ajaysubra, line 521: I would add the delegate method that they need to implement with code samples.

ajaysubra: also not a blocker but if you can update the [example app](https://github.com/klaviyo/klaviyo-swift-sdk/blob/master/Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift) with that code that would be good for devs to look at.

jasper-hsu-kyvo: Look good now? Sorry still don't really know Swift well so lmk...

ajaysubra: Thanks for making the update. This looks good. Only last change I would suggest is instead of saying " but your app must enable `Remote Notifications` under `Background Modes`" I recommend linking to official apple docs for two reasons, 
1. If something changes on their side for setup, we don't need to update the docs
2. It's a little more clear with screenshots etc.

Point to the [Enable the remote notifications capability](https://developer.apple.com/documentation/usernotifications/pushing-background-updates-to-your-app#Enable-the-remote-notifications-capability) and [Enable the remote notifications capability](https://developer.apple.com/documentation/usernotifications/pushing-background-updates-to-your-app#Enable-the-remote-notifications-capability) section more specifically. The middle section [Enable the remote notifications capability](https://developer.apple.com/documentation/usernotifications/pushing-background-updates-to-your-app#Enable-the-remote-notifications-capability) is what we take care in our sending pipeline so suggesting to point to the first and last section specifically.

README.md
```diff
@@ -514,6 +515,29 @@ By default, the Klaviyo SDK automatically clears the badge count on app open. If
...
+>  ℹ️ Silent push support is available by default. No special SDK methods are required to receive them. See [enable the remote notifications capability](https://developer.apple.com/documentation/usernotifications/pushing-background-updates-to-your-app#Enable-the-remote-notifications-capability) and [receive background notifications](https://developer.apple.com/documentation/usernotifications/pushing-background-updates-to-your-app#Enable-the-remote-notifications-capability) for more details.
+
+To handle silent push notifications, implement the following delegate method in your `AppDelegate`:
+
+```
+func application(_ application: UIApplication, didReceiveRemoteNotification userInfo: [AnyHashable : Any], fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+
+    // Convert payload to Data
+    do {
+        let serializedPayload = try JSONSerialization.data(withJSONObject: userInfo, options: [])
```
ndurell, line 530: why are we serializing this here? Should we describe the structure and where the kv pairs are?

Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift
```diff
@@ -96,6 +96,22 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    func application(
+        _ application: UIApplication,
+        didReceiveRemoteNotification userInfo: [AnyHashable : Any],
+        fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void
+    ) {
+        // Convert payload to Data
+        do {
+            let serializedPayload = try JSONSerialization.data(withJSONObject: userInfo, options: [])
+
+            // Process or forward the data
```
jordan-griffin, line 110: Do we need to add something to parse out key/value information?

README.md
```diff
@@ -514,6 +515,29 @@ By default, the Klaviyo SDK automatically clears the badge count on app open. If
 
 Klaviyo SDK will automatically handle the badge count associated with Klaviyo pushes. If you need to manually update the badge count to account for other notification sources, use the `KlaviyoSDK().setBadgeCount(:)` method, which will update the badge count and keep it in sync with the Klaviyo SDK. This method should be used instead of (rather than in addition to) setting the badge count using `UNUserNotificationCenter` and/or `UIApplication` methods.
 
+#### Silent Push Notifications
```
jordan-griffin, line 518: We have a "custom data" section proposed for the Android SDK that digs into key/value pairs and I'm kinda wondering if we want something similar for iOS here

README.md
```diff
@@ -514,6 +515,29 @@ By default, the Klaviyo SDK automatically clears the badge count on app open. If
...
+
+Silent push notifications (also known as background pushes) allow your app to receive payloads from Klaviyo without displaying a visible alert to the user. These are typically used to trigger background behavior, such as displaying content, personalizing the app interface, or downloading new information from a server.
+>  ℹ️ Silent push support is available by default. No special SDK methods are required to receive them. See [enable the remote notifications capability](https://developer.apple.com/documentation/usernotifications/pushing-background-updates-to-your-app#Enable-the-remote-notifications-capability) and [receive background notifications](https://developer.apple.com/documentation/usernotifications/pushing-background-updates-to-your-app#Enable-the-remote-notifications-capability) for more details.
```
jordan-griffin, line 521: nit: I guess what we are trying to say here is that there's nothing the Klaviyo SDK needs to do to handle silent push and its own the user to implement their silent push handlers themselves via the provided code snippet below, but it read to me as "you dont need to do anything for silent push to work" followed by a code snippet to get set up which was a little confusing.

README.md
jasper-hsu-kyvo: Oops that was left in from the test app code I was modeling off of - I guess we don't need to serialize to JSON at all and just leave the data/`userInfo` as that dictionary of type `[AnyHashable : Any]`

Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift
jasper-hsu-kyvo: True can add that in

README.md
jasper-hsu-kyvo: Ahh good callout we should have parity

README.md
jasper-hsu-kyvo: I can add a line that says something along the lines of "you still need to do some work - here is what you do"

Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift
```diff
@@ -97,6 +97,23 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
     }
 
+    // MARK: Silent Push Notification implementation
+
+    func application(
+        _ application: UIApplication,
+        didReceiveRemoteNotification userInfo: [AnyHashable: Any],
+        fetchCompletionHandler completionHandler: @escaping (UIBackgroundFetchResult) -> Void) {
+        // Access custom key-value pairs from the top level
+        if let customData = userInfo["key_value_pairs"] as? [String: String] {
```
ndurell, line 107: hmm this is where store key_value_pairs?

jordan-griffin (approved):

belleklaviyo:
small nit: in Step 4 of the Installation steps, where we call out the sample `NotificationService.file` also mention custom data 

> 4. Finally, in the `NotificationService.swift` file add the code for the two required delegates from [this](Examples/KlaviyoSwiftExamples/SPMExample/NotificationServiceExtension/NotificationService.swift) file. This sample covers calling into Klaviyo so that we can download and attach the media to the push notification as well as handle custom badge counts. **It also demonstrates how to access custom data (key-value pairs) sent from Klaviyo..**

or something like that

Examples/KlaviyoSwiftExamples/Shared/AppDelegate.swift
jasper-hsu-kyvo: Was able to confirm with sim btw*

ndurell (approved):

PR #325, opened 2025-04-09
changed SwiftFormat `closingparen` setting to `balanced`

ab1470:
# Description

This PR updates our .swiftformat settings to adjust the closing parentheses behavior. Prior to this PR, SwiftFormat would automatically change 

```
func foo(
    parameter1: String,
    parameter2: String
) {
    print("bar")
}
```
to
```
func foo(
    parameter1: String,
    parameter2: String) {
    print("bar")
}
```

Personally, I find that more difficult to read. It's also contrary to the default Xcode behavior for handling the closing parenthesis on multiline function signature; if you were to highlight the 2nd code block above in Xcode and hit control-m, Xcode would automatically reformat it to look like the 1st code block. In general, I prefer to adhere to Xcode's default recommended code style because I think it does a pretty good job and because I don't think we should be "fighting the system".

This PR changes the SwiftFormat rule so that SwiftFormat will format a multiline function signature to look like the 1st code block above, not the 2nd.

### DIFF
--- a/.swiftformat
+++ b/.swiftformat
@@ -11,7 +11,7 @@
 # format options
 
 --exclude Sources/KlaviyoSwift/Vendor,Tests/KlaviyoSwiftTests/Vendor,Tests/KlaviyoSwiftTests/__Snapshots__
---closingparen same-line
+--closingparen balanced
 --commas inline
 --comments indent
 --decimalgrouping 3,5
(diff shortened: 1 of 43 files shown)


ajaysubra (approved):

belleklaviyo (approved):
do you need to update the pre-commit config for this?

njparadis (approved):

PR #326, opened 2025-04-09
Refactor KlaviyoWebViewController deinitialization

ab1470:
# Description

This PR refactors the `KlaviyoWebViewController`'s deinitialization so that we call `removeScriptMessageHandler()` in the `deinit` method rather than in the `viewDidDisappear` and `dismiss` methods. We need to call `removeScriptMessageHandler()` to avoid a retain cycle when we deallocate the `KlaviyoWebViewController`.

This change is necessary for upcoming work on maintaining a persistent web view. For our MVP IAF release, it was okay for us to remove the script message handlers when the `KlaviyoWebViewController` was dismissed/removed from the view hierarchy. This approach will not work with persistent web views, however, because we need to be able to dismiss the `KlaviyoWebViewController` from the view hierarchy without deallocating it entirely, and the view controller needs to remain alive and listening for script messages so that it can re-present itself if and when it receives another `formWillAppear` event.

I wanted to make these changes prior to working on [CHNL-18661](https://klaviyo.atlassian.net/browse/CHNL-18661) because that ticket will involve some refactoring of the same parts of the codebase that the changes in this PR touches, and I didn't want to make changes for 18661 that will need to be changed again for the work on persistent web views.

Even if, for whatever reason, we don't end up implementing persistent web views (or we employ a different approach than the one we've been discussing), I think the changes in this PR are still valid and an improvement because I think it's better to remove the script message handlers on the viewController's deinitialization, rather than rely on dismiss and viewDidDisappear to handle it.

# Check List

# Manual Test Plan

In addition to the unit tests that I added as part of this PR to validate the changes are working as expected, I also ran the test app and displayed/dismissed a form and validated that the `KlaviyoWebViewController` is successfully getting deinitialized when the form is dismissed.

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
@@ -32,6 +32,7 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
 
     // MARK: - Scripts
 
+    @MainActor
     private var klaviyoJsWKScript: WKUserScript? {
         var apiURL = environment.cdnURL()
         apiURL.path = "/onsite/js/klaviyo.js"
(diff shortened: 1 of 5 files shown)


belleklaviyo (approved):
neat test

PR #328, opened 2025-04-10
[CHNL-18661] remove `dismiss` from delegate

ab1470:
# Description

This PR removes the `dismiss()` requirement from the `KlaviyoWebViewDelegate`. With #326, [the logic within the `KlaviyoWebViewController`'s `dismiss` method was moved to the `deinit` method](https://github.com/klaviyo/klaviyo-swift-sdk/pull/326/commits/9e700274e190e3b45ce33fdd9ecb73fc731a0c9b), leaving the view controller's `dismiss` method not really doing anything, so we no longer need the delegate to implement its own `dismiss` logic. 

# Check List

# Manual Test Plan

I ran the test app and validated that the In-App Form is behaving as expected. The form dismisses successfully, and the KlaviyoWebViewController is deinitialized after the form is dismissed.

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -14,6 +14,8 @@ import UIKit
 class IAFPresentationManager {
     static let shared = IAFPresentationManager()
 
+    let animateDismissal = false
+
     lazy var indexHtmlFileUrl: URL? = {
         do {
             return try ResourceLoader.getResourceUrl(path: "InAppFormsTemplate", type: "html")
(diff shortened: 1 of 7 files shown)


belleklaviyo (approved):

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -14,6 +14,8 @@ import UIKit
 class IAFPresentationManager {
     static let shared = IAFPresentationManager()
 
+    let animateDismissal = false
```
ajaysubra, line 17: why is this a flag?

ajaysubra: also minor nitpick, should it be `private`?

ab1470: good catch, it should be private. Fixed.

It's more like a constant than a flag. We call `dismiss(animated: Bool)` in three places so I figured it would be cleaner to define the `animated` value once

PR #329, opened 2025-04-10
[CHNL-18661] change timeout type

ab1470:
# Description

This PR changes the `preloadWebsite(timeout:)` method to take a `TimeInterval` instead of a `UInt64` as a parameter. I think it's more readable this way, and it's aligned with some of Apple's APIs that use a timer (like `Timer.scheduledTimer(timeInterval: TimeInterval, <...>)`, `Date(timeInterval: TimeInterval, <...>)`, etc).

# Check List

# Manual Test Plan

In addition to updating the unit tests, I also ran the iOS test app and validated that the In-App Forms are behaving as expected. I also simulated a timeout scenario in the test app and validated that the timeout is working as expected.

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
@@ -99,7 +99,7 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
 
     // MARK: - Loading
 
-    func preloadWebsite(timeout: UInt64) async throws {
+    func preloadWebsite(timeout: TimeInterval) async throws {
         guard let delegate else { return }
 
         await delegate.preloadUrl()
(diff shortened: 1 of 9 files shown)


Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -112,7 +112,7 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
                 }
 
                 group.addTask {
-                    try await Task.sleep(nanoseconds: timeout)
+                    try await Task.sleep(nanoseconds: UInt64(timeout * 1_000_000_000))
```
ajaysubra, line 115: not using the extension you added here?

ajaysubra (approved):

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
ab1470: That wouldn't work here. `Task.sleep(nanoseconds:)` expects a `UInt64`, but the `timeout` is a `TimeInterval`. So we need to convert the `TimeInterval` to a `UInt64` in order to pass it into `Task.sleep(nanoseconds:)`.

The extension I added does the opposite, it converts a `UInt64` to a `TimeInterval`

belleklaviyo (approved):

PR #330, opened 2025-04-11
[CHNL-18661] add `withTimeout` function

ab1470:
# Description

This PR adds a `withTimeout` function as a generic way to call an asynchronous task and throw an error if the task doesn't complete before the specified timeout is reached. I refactored the `IAFWebViewModel.preloadWebsite(timeout:)` method to use this new `withTimeout` function.

# Check List

# Manual Test Plan

In addition to passing the unit tests, I also ran the iOS test app and validated that In-App Forms are working as expected. I also simulated a timeout in the test app and validated that the timeout is handled as expected.


Sources/KlaviyoCore/Models/ObjectStateError.swift
```diff
@@ -0,0 +1,10 @@
+//
```
ajaysubra, line 1: Same comment as the other PR - Do we anticipate this being used in other modules if not should be keep it in forms?

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -99,35 +99,19 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
...
-
-                group.addTask { [weak self] in
-                    guard let self else { return }
-
-                    var iterator = self.formWillAppearStream.makeAsyncIterator()
-                    await iterator.next()
-                }
-
-                try await group.next()
+            try await withTimeout(seconds: timeout) { [weak self] in
```
ajaysubra, line 108: where are we stopping the timer once we receive handshook from JS?

Sources/KlaviyoCore/Models/ObjectStateError.swift
ab1470: same comment as [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/329#discussion_r2039926730). This is a common error that's really not specific to KlaviyoForms; there are many parts of the code where we may want to throw an error if an expected object doesn't exist. But I can move it if you prefer.

evan-masseau (approved):

belleklaviyo (approved):

PR #331, opened 2025-04-14
[CHNL-18661] update timeout

ab1470:
# Description

This PR updates the timeout logic for In-App Forms so that it waits for the handshake, rather than formWillAppear, before canceling the timeout timer. As part of this work, I also moved the responsibility for dismissing the `KlaviyoWebViewController` solely within the `IAFPresentationManager`. This allows us to better manage the full view lifecycle and deallocation of the `KlaviyoWebViewController` in a single place, which is important for upcoming work on persistent web views.

[CHNL-18661](https://klaviyo.atlassian.net/browse/CHNL-18661)

# Check List

# Manual Test Plan

- I ran the test app and validated that In-App Forms are working as expected
- I simulated triggering the timeout in the test app, and validated that the app worked as expected
- I simulated an "abort" message in the test app, and validated that the app worked as expected

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -14,7 +14,8 @@ import UIKit
 class IAFPresentationManager {
     static let shared = IAFPresentationManager()
 
-    private let animateDismissal = false
+    private var viewController: KlaviyoWebViewController?
+    private var isLoading: Bool = false
 
     lazy var indexHtmlFileUrl: URL? = {
         do {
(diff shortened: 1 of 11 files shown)


belleklaviyo (approved):

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -52,32 +52,71 @@ class IAFPresentationManager {
...
+                viewController = nil
                 return
             }
 
-            guard let topController = UIApplication.shared.topMostViewController else {
-                viewController.dismiss(animated: animateDismissal)
-                return
+            // now that we've established the handshake, we can start a task that listens for Form events.
+            formEventTask = Task {
+                for await event in viewModel.formLifecycleStream {
```
evan-masseau, line 69: neat

evan-masseau (approved):

PR #333, opened 2025-04-16
[CHNL-17060] fix react native user-agent

ab1470:
# Description

We had a bug in which API calls made using the React Native SDK would show the incorrect user agent if the SDK was imported dynamically. This happened because we were attempting to access the `klaviyo-sdk-configuration.plist` file, which we use to set the user agent, from `Bundle.main`, but that file does not exist in the main bundle when the SDK is imported dynamically. This PR fixes that by searching for the file specifically within the react native framework if it's unable to find it in the main app bundle.

[CHNL-17060](https://klaviyo.atlassian.net/browse/CHNL-17060)

# Check List

# Manual Test Plan

With the changes in this PR, I validated that the user agents are correct when the frameworks are linked statically and dynamically. To reproduce, I:

1. pointed the React Native Test App to consume the local React Native SDK (commit 9fddd58) and the local Swift SDK (this branch)
2. ran `pod install`, built and ran the RN test app, triggered an event, and inspected the user agent in Proxyman to validate that it's what we expect
3. set the `klaviyo-react-native-test-app/iOS/Podfile`'s `use_frameworks!` setting to `dynamic`
4. repeated step 2

User-agent when podfile includes `use_frameworks! :linkage => :static`:
`KlaviyoRNTestApp/1.4.0 (com.klaviyo.rntest; build:1; iOS 18.2.0) klaviyo-react_native/1.2.0`

User-agent when podfile includes `use_frameworks! :linkage => :dynamic`:
`KlaviyoRNTestApp/1.4.0 (com.klaviyo.rntest; build:1; iOS 18.2.0) klaviyo-react_native/1.2.0`

[CHNL-17060]: https://klaviyo.atlassian.net/browse/CHNL-17060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/KlaviyoCore/Utils/FileUtils.swift
+++ b/Sources/KlaviyoCore/Utils/FileUtils.swift
@@ -69,13 +69,22 @@ public func removeFile(at url: URL) -> Bool {
     return false
 }
 
-/// load any plist from app main bundle
+/// load any plist from app main bundle or React Native framework bundle
 /// - Parameter name: the name of the plist
 /// - Returns: the contents of the plist in `[String: AnyObject]` or nil if not found
 func loadPlist(named name: String) -> [String: AnyObject]? {
+    // Try loading from main bundle first
     if let path = Bundle.main.path(forResource: name, ofType: "plist"),
        let dict = NSDictionary(contentsOfFile: path) as? [String: AnyObject] {
         return dict
     }
+
+    // If not found in main bundle, try loading from React Native framework bundle
+    if let reactNativeBundle = Bundle(identifier: "org.cocoapods.klaviyo-react-native-sdk"),
+       let path = reactNativeBundle.path(forResource: name, ofType: "plist"),
+       let dict = NSDictionary(contentsOfFile: path) as? [String: AnyObject] {
+        return dict
+    }
+
     return nil
 }
--- a/Sources/KlaviyoCore/Utils/FileUtils.swift
+++ b/Sources/KlaviyoCore/Utils/FileUtils.swift
@@ -73,18 +73,23 @@ public func removeFile(at url: URL) -> Bool {
 /// - Parameter name: the name of the plist
 /// - Returns: the contents of the plist in `[String: AnyObject]` or nil if not found
 func loadPlist(named name: String) -> [String: AnyObject]? {
-    // Try loading from main bundle first
-    if let path = Bundle.main.path(forResource: name, ofType: "plist"),
-       let dict = NSDictionary(contentsOfFile: path) as? [String: AnyObject] {
-        return dict
-    }
+    let plistPath: String? = {
+        if let path = Bundle.main.path(forResource: name, ofType: "plist") {
+            // Try loading from main bundle first
+            return path
+        } else if let reactNativeBundle = Bundle(identifier: "org.cocoapods.klaviyo-react-native-sdk"),
+                  let path = reactNativeBundle.path(forResource: name, ofType: "plist") {
+            // If not found in main bundle, try loading from React Native framework bundle
+            return path
+        } else {
+            return nil
+        }
+    }()
 
-    // If not found in main bundle, try loading from React Native framework bundle
-    if let reactNativeBundle = Bundle(identifier: "org.cocoapods.klaviyo-react-native-sdk"),
-       let path = reactNativeBundle.path(forResource: name, ofType: "plist"),
-       let dict = NSDictionary(contentsOfFile: path) as? [String: AnyObject] {
+    if let plistPath,
+       let dict = NSDictionary(contentsOfFile: plistPath) as? [String: AnyObject] {
         return dict
+    } else {
+        return nil
     }
-
-    return nil
 }


Sources/KlaviyoCore/Utils/FileUtils.swift
```diff
@@ -69,13 +69,22 @@ public func removeFile(at url: URL) -> Bool {
...
         return dict
     }
+
+    // If not found in main bundle, try loading from React Native framework bundle
+    if let reactNativeBundle = Bundle(identifier: "org.cocoapods.klaviyo-react-native-sdk"),
+       let path = reactNativeBundle.path(forResource: name, ofType: "plist"),
+       let dict = NSDictionary(contentsOfFile: path) as? [String: AnyObject] {
+        return dict
+    }
+
```
evan-masseau, line 88: Oh gross, but good find!

evan-masseau: This might mean we'll need a similar solution for the flutter SDK. I don't remember how far we got with helping them set this. @dan-peluso do you have any idea? Quick google tells me SPM is now encouraged for flutter integration anyway, but cocoapods is still supported.

ndurell: We didn't get too far! But I'm not sure if we tried this method!

Sources/KlaviyoCore/Utils/FileUtils.swift
```diff
@@ -69,13 +69,27 @@ public func removeFile(at url: URL) -> Bool {
...
 /// - Parameter name: the name of the plist
 /// - Returns: the contents of the plist in `[String: AnyObject]` or nil if not found
 func loadPlist(named name: String) -> [String: AnyObject]? {
-    if let path = Bundle.main.path(forResource: name, ofType: "plist"),
-       let dict = NSDictionary(contentsOfFile: path) as? [String: AnyObject] {
+    let plistPath: String? = {
+        if let path = Bundle.main.path(forResource: name, ofType: "plist") {
+            // Try loading from main bundle first
+            return path
+        } else if let reactNativeBundle = Bundle(identifier: "org.cocoapods.klaviyo-react-native-sdk"),
```
ajaysubra, line 80: neat find
QQ - is there anyway we don't use the string? Would this bundle identifier be the same in customers app?

Sources/KlaviyoCore/Utils/FileUtils.swift
evan-masseau: yea but their bundle identifier won't be klaivyo-react-native-sdk

Sources/KlaviyoCore/Utils/FileUtils.swift
ab1470: yeah I'd like to not use a string but from my research and and testing a couple of different things I couldn't find a good way. One suggested solution is to grab the bundle ID by referencing some object within the SDK:

```swift
extension Bundle {
    static var klaviyoSDK: Bundle = {
        return Bundle(for: SomeClassInYourSDK.self)
    }()
}
```
For this to work for the purpose of resolving this bug ticket, `SomeClassInYourSDK` would need to be a class within the React Native SDK; i.e., `SomeClassInYourReactNativeSDK`. But we can't do that because `FileUtils` is defined in `KlaviyoCore`, which has no knowledge of the React Native SDK, and therefore can't access `SomeClassInYourReactNativeSDK`

ajaysubra: Thanks for explaining. Aligned on both your points.

Sources/KlaviyoCore/Utils/FileUtils.swift
ajaysubra: yeah I just learnt from the other thread that we are using the bundle of the cocoapod and not the app. Wondering if there is a general solution for this @ab1470 if not I think what we have now is good and we can revisit when we want this to work for flutter.

Sources/KlaviyoCore/Utils/FileUtils.swift
evan-masseau: actually, Is it possible to edit a .plist from code? Would it be possible to put code in the RN SDK (maybe in that swift bridge file we have in there?) that edits the `klaviyo-sdk-configuration.plist` to override the SDK name prior to calling initialize?

ab1470: the issue isn't overriding the plist, the issue is making sure the plist is successfully bundled in the React Native SDK and will be accessible to the app that hosts the SDK. The plist itself [lives in our RN SDK](https://github.com/klaviyo/klaviyo-react-native-sdk/blob/master/ios/klaviyo-sdk-configuration.plist), and the only purpose of that plist is to supply the SDK name and version values for the RN SDK.

Even if we wanted to, though, it's not possible to programmatically edit a .plist; it's a read-only file that's included in the app bundle.

Another question is whether there's a better way to handle this than using a plist. I'm not sure the full context of why we chose to use a plist for this purpose, but it might be worth investigating a different approach. I would argue that work should be captured in a separate ticket, though. 

@ajaysubra , looks like you added the plist in [this PR](https://github.com/klaviyo/klaviyo-react-native-sdk/pull/175). Do you remember why you decided to use a plist, and if you looked into any alternatives?

belleklaviyo (approved):

Sources/KlaviyoCore/Utils/FileUtils.swift
ajaysubra: Kinda outlined in the PR description [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/204) it's been a while so my memory is a little hazy.

ajaysubra: okay finally jogged my memory so when we had private SDK on native side and using that to set the sdk name and value was problematic when RN devs use native initialization. The private method call was made during RN initialize. 

There may be other ways this can be done but I don't know if I ever explored them myself.

ajaysubra (approved):

Sources/KlaviyoCore/Utils/FileUtils.swift
ab1470: Thanks, that helps. I think we can revisit this if and when it becomes necessary, but for now I think the fix I implemented in this PR should suffice. Merging now.

PR #336, opened 2025-04-28
CHNL-19752: KlaviyoForms app lifecycle listeners

belleklaviyo:
# Description

- Makes the `AppLifeCycleEvents` a package-level singleton 
- Removes the `.handleEvents` modifer out from the singleton instance
- Creates a new publisher that emits lifecycle events and handles reachability
- Implements that publisher where we are subscribing to lifecycle events in KlaviyoSwift (in StateManagement, part of `completeInitialization`)
- KlaviyoForms now listens to the `AppLifeCycleEvents` (entry point being `registerForInAppForms()` as a separate call to a `setupLifecycleEvents()` func)

Tests were tricky to update, some seemed to have duplicated setup, so some were consolidated just to check the lifecycle events.

https://klaviyo.atlassian.net/browse/CHNL-19752

# Check List

# Manual Test Plan

https://github.com/user-attachments/assets/8f31df78-ed17-4b72-ba6b-4c99664494f3

# Supporting Materials


Sources/KlaviyoCore/AppLifeCycleEvents.swift
```diff
@@ -26,25 +26,12 @@ public struct AppLifeCycleEvents {
...
-            })
             .map { _ in LifeCycleEvents.foregrounded }
         let backgrounded = environment
             .notificationCenterPublisher(UIApplication.didEnterBackgroundNotification)
-            .handleEvents(receiveOutput: { _ in
-                environment.stopReachability()
-            })
             .map { _ in LifeCycleEvents.backgrounded }
         // The below is a bit convoluted since network status can be nil.
         let reachability = environment
```
belleklaviyo, line 37: maybe somebody else has more context to this part -- not sure if this needs to be accounted in the newly created publisher in some way

Sources/KlaviyoCore/AppLifeCycleEvents.swift
```diff
@@ -57,18 +44,11 @@ public struct AppLifeCycleEvents {
...
-                }
-            })
             .receive(on: RunLoop.main)
             .eraseToAnyPublisher()
     }) {
         self.lifeCycleEvents = lifeCycleEvents
     }
 
-    static let production = Self()
+    public static let production = Self()
```
ab1470, line 53: you're marking this `public` so the KlaviyoForms module will have access to it, correct? If so, then I think you can mark it as `package` instead of `public`. That will expose it to KlaviyoForms, but it will still remain inaccessible to the host app that consumes `KlaviyoCore`

belleklaviyo: good call yeah will change to that

Tests/KlaviyoSwiftTests/AppLifeCycleEventsTests.swift
```diff
@@ -131,34 +140,26 @@ class AppLifeCycleEventsTests: XCTestCase {
...
+            lifecycleSubject.eraseToAnyPublisher()
+        })
+        environment.appLifeCycle = customLifeCycleEvents
+        let cancellable = environment.lifecycleEventsWithReachability().sink { _ in }
 
-        wait(for: [expection], timeout: 0.1)
+        lifecycleSubject.send(.foregrounded)
+        await fulfillment(of: [expectation], timeout: 1.0)
         cancellable.cancel()
-        XCTAssertEqual(1, expection.expectedFulfillmentCount)
```
ndurell, line 161: is it being called more than once?

ndurell (approved):
I think this is fine just want to make sure we aren't missing anything important with the test mods.

PR #337, opened 2025-05-02
Implement app session with injected events

belleklaviyo:
# Description
A bit large but it makes more sense for testing to have all of these changes together -- I could break this down into smaller PRs if we want but I think the commits generally make sense to walk through
- Created new method `destroyWebView`, renamed presentIAF to `constructWebView`, modified `presentForm` and `dismissForm` accordingly
    - `dismissForm` does not destroy the web view, only hides it

- Added script to inject a `dispatchLifecycleEvent` method we can use to dispatch lifecycle events to the `document.head`
- Added new IAFNativeBridgeEvents `analyticsEvent` and `lifecycleEvent` to handshake
- Implemented app session in the publisher for each lifecycle event according to rules in [ticket](https://klaviyo.atlassian.net/browse/CHNL-19753)

| App State | Action | Event Dispatched |
|-----------|---------|------------------|
| Backgrounded | Record timestamp | CustomEvent('lifecycleEvent', { detail: { type: "background", session: "persist" } }) |
| Launched | Create new webview | CustomEvent('lifecycleEvent', { detail: { type: "foreground", session: "restore" } }) |
| Foregrounded (within session) | No change | CustomEvent('lifecycleEvent', { detail: { type: "foreground", session: "restore" } }) |
| Foregrounded (new session) | Destroy & create webview | CustomEvent('lifecycleEvent', { detail: { type: "foreground", session: "purge" } }) |
| Company ID changed | Destroy & create webview | CustomEvent('lifecycleEvent', { detail: { type: "foreground", session: "purge" } }) |
   

# Check List

# Manual Test Plan

https://github.com/user-attachments/assets/cfc5cd7c-5e04-40a7-915b-d39119edabca

https://github.com/user-attachments/assets/b6381f07-ad8a-48de-9f03-9196f9756f0c

# Supporting Materials


Sources/KlaviyoForms/KlaviyoWebView/KlaviyoWebViewController.swift
```diff
@@ -145,8 +145,8 @@ class KlaviyoWebViewController: UIViewController, WKUIDelegate, KlaviyoWebViewDe
     #endif
 
     @MainActor
-    func evaluateJavaScript(_ script: String) async throws -> Any {
-        try await webView.evaluateJavaScript(script)
+    func evaluateJavaScript(_ script: String) {
+        webView.evaluateJavaScript(script, completionHandler: nil)
```
belleklaviyo, line 149: need to explicitly call this signature to avoid a [fatal error ](https://stackoverflow.com/questions/74364029/fatal-error-while-using-evaluatejavascript-on-wkwebview)

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -12,14 +12,24 @@ import KlaviyoSwift
...
+    private var apiKeyCancellable: AnyCancellable?
     private var viewController: KlaviyoWebViewController?
+
     private var isLoading: Bool = false
     private var formEventTask: Task<Void, Never>?
 
+    private init() {}
+
+    #if DEBUG
+    package init(viewController: KlaviyoWebViewController?) {
```
belleklaviyo, line 28: for testing

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -28,25 +38,80 @@ class IAFPresentationManager {
...
-                case let .reachabilityChanged(status: status):
-                    break
+                Task { @MainActor in
+                    switch event {
+                    case .terminated:
+                        break
+                    case .foregrounded:
+                        if let lastBackgrounded = UserDefaults.standard.object(forKey: "lastBackgrounded") as? Date {
+                            let timeElapsed = Date().timeIntervalSince(lastBackgrounded)
+                            if timeElapsed > 2.0 {
```
belleklaviyo, line 65: for now just have an app session length of 2 seconds for sake of testing/development but will update to something configurable and longer in [CHNL-19775](https://klaviyo.atlassian.net/browse/CHNL-19775)

Tests/KlaviyoFormsTests/KlaviyoFormsTestUtils.swift
```diff
@@ -0,0 +1,185 @@
...
+extension ArchiverClient {
+    static let test = ArchiverClient(
+        archivedData: { _, _ in ARCHIVED_RETURNED_DATA },
+        unarchivedMutableArray: { _ in SAMPLE_DATA }
+    )
+}
+
+extension KlaviyoEnvironment {
+    static var lastLog: String?
+    static var test = {
```
belleklaviyo, line 81: just needed it for this test instance in this module, basically duplicated a TestUtil that already exists for another test group

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -126,14 +189,20 @@ class IAFPresentationManager {
...
             topController.present(viewController, animated: false, completion: nil)
         }
     }
 
-    @MainActor
-    private func dismissForm() {
+    package func dismissForm() {
+        guard let viewController else { return }
+        viewController.view.isHidden = true
+        viewController.view.isUserInteractionEnabled = false
```
ab1470, line 201: just curious why these are needed. Is just `dismiss`ing not sufficient?

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -192,6 +211,10 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
             }
             handshakeContinuation.yield()
             handshakeContinuation.finish()
+        case .analyticsEvent:
+            ()
+        case .lifecycleEvent:
+            ()
```
ab1470, line 217: just placeholders for now?

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
belleklaviyo: ah right I think that was leftover when I ran into [this](https://klaviyo.atlassian.net/browse/CHNL-20493) bug but not actually necessary here anymore, thanks for catching

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -28,25 +38,80 @@ class IAFPresentationManager {
...
+                            self?.constructWebview()
+                        })
+                    }
                 }
             }
     }
 
-    @MainActor
-    func presentIAF(assetSource: String? = nil) {
+    package func constructWebview(assetSource: String? = nil) {
```
ab1470, line 114: nice, much better name

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -126,14 +189,20 @@ class IAFPresentationManager {
...
-            dismissForm()
+            destroyWebView()
         } else {
             topController.present(viewController, animated: false, completion: nil)
         }
     }
 
-    @MainActor
-    private func dismissForm() {
+    package func dismissForm() {
```
ab1470, line 198:
```suggestion
    func dismissForm() {
```
see above

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -28,25 +38,80 @@ class IAFPresentationManager {
...
+            if let additionalAction = additionalAction {
+                await additionalAction()
+            }
+        } catch {
+            print("Javascript evaluation failed; message: \(error.localizedDescription)")
+        }
+    }
+
+    package func setupLifecycleEvents() {
+        lifecycleCancellable = environment.appLifeCycle.lifeCycleEvents()
```
evan-masseau, line 56: Wrestling with this idea on android right now myself: what do you think about extracting all the observers (like lifecycle events and profile changes) into a different class solely responsible for observing those events and injecting them into the webview?

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -126,14 +189,20 @@ class IAFPresentationManager {
...
-    @MainActor
-    private func dismissForm() {
+    package func dismissForm() {
+        guard let viewController else { return }
+        viewController.view.isHidden = true
+        viewController.view.isUserInteractionEnabled = false
+        viewController.dismiss(animated: false)
+    }
+
+    package func destroyWebView() {
```
ab1470, line 203:
```suggestion
    func destroyWebView() {
```
see above

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
evan-masseau: I don't have a strong opinion, but should we merge this work into a feature branch instead of master, until everything pertaining to app session is complete?

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -28,25 +38,80 @@ class IAFPresentationManager {
         }
     }()
 
-    func setupLifecycleEvents() {
-        lifecycleCancellable = AppLifeCycleEvents.production.lifeCycleEvents()
+    package func handleLifecycleEvent(_ event: String, _ session: String, additionalAction: (() async -> Void)? = nil) async throws {
```
ab1470, line 41:
```suggestion
    func handleLifecycleEvent(_ event: String, _ session: String, additionalAction: (() async -> Void)? = nil) async throws {
```
Unless the other modules in our Swift Package (i.e., KlaviyoCore or KlaviyoSwift) need access to this code, I think this function can be `internal` instead of `package`. You can leave off the `internal` modifier because that's the default

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -28,25 +38,80 @@ class IAFPresentationManager {
...
+            }
+            if let additionalAction = additionalAction {
+                await additionalAction()
+            }
+        } catch {
+            print("Javascript evaluation failed; message: \(error.localizedDescription)")
+        }
+    }
+
+    package func setupLifecycleEvents() {
```
ab1470, line 55:
```suggestion
    func setupLifecycleEvents() {
```
see above

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
ab1470, line 114:
```suggestion
    func constructWebview(assetSource: String? = nil) {
```
see above

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
belleklaviyo: that's a good idea! this change definitely made this class pretty beefy, and I could see it being worth separating out in case it gets more complicated

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
belleklaviyo: ah got it okay I thought it was necessary for tests but that makes sense

ab1470: the `@testable` tag should take care of exposing those methods for testing

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -80,6 +80,24 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
...
+                            type: type,
+                            session: session
+                        }
+                    }));
+                };
+            })();
+        """
+        return WKUserScript(source: lifecycleScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
+    }
+
```
evan-masseau, line 100: I like this idea of injecting the function early, and only having to call the method later.
I tried something similar on android (but we don't have this helpful method for injecting scripts with an `injectionTime`, I had to just put the functions in a script tag directly in the HTML file). In that case, I encountered an issue where sometimes the function wasn't defined yet when the very first lifecycle event occurred, so I'm having to re-think my approach.
Have you tested for the edge case where a lifecycle event gets triggered before this script has been injected? Maybe that isn't possible timing-wise in your case?

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
belleklaviyo: makes sense, made a feature branch `feature/IAF-app-session` will point to that

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
evan-masseau: Yeah I'm trying to decide how much to split things up. The super granular way might be something like `WebViewSessionService`, responsible for tracking session, activity timeout, and injecting lifecycle events, and `ProfileObserver` for profile changes, `AnalyticsEventObserver` for analytics, etc.

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
belleklaviyo: yeah I don't really think that is possible right now? a lifecycle event is only possible after the handshake, right, which is done after already creating the view controller (including the view model where this is set up on init) (see: the `constructWebview` Task in `IAFPresentationManager`)

evan-masseau: Gonna go back to my original approach then 🤦, and I'll throw out a suggestion for you too: you could avoid this mixed language situation of putting JS in a swift file, by instead adding this as a script tag the HTML file like so: 

Screenshot 2025-05-07 at 11 37 38

belleklaviyo: good point, less ugly yay!

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
belleklaviyo: I'll make that a followup PR if we decide to do that

evan-masseau: Yeah that makes sense, especially now that you're using a feature branch to merge work!

PR #341, opened 2025-05-15
[CHNL-20505] Plugin user-agent value appending

dan-peluso:
# Description
We want more visibility into when customers are using the react native SDK in a traditional workflow, or using a framework like expo. This logic checks for the added plist ([which we add to the built ios project with this ticket](https://github.com/klaviyo/klaviyo-expo-plugin/pull/8)) and appends it to the user agent if it's included in the app bundle. 

Definitely have a low confidence level in swift so please feel free to rip this apart - I also took a shot at a unit test for validation.

Here is a screenshot from local setup to show it appending:
image
`User-Agent: klaviyopluginexample/1.0.0 (com.klaviyo.expoexample; build:1; iOS 18.1.0) klaviyo-react_native/1.2.0 (klaviyo-expo/0.0.2)`

## Due Diligence

## Test Plan
this is a pain to setup but if you'd like to test locally:
- checkout both this branch and the expo plugin version linked above
- link expo example app the local react native SDK and validate it runs ([instructions on this in the expo pr](https://github.com/klaviyo/klaviyo-expo-plugin/pull/7))
- link local swift SDK to the react native SDK 
- from the expo example directory, run 'npm run clean-ios' and stop it after it get's to the 'installing pods' step. You should now see an 'ios' folder under the example directory
- then, in the Podfile of that directory add the following code change:
```
# Local Swift SDK dependencies
pod 'KlaviyoSwift', :path => '/Users/daniel.peluso/Klaviyo/Repos/klaviyo-swift-sdk'
pod 'KlaviyoForms', :path => '/Users/daniel.peluso/Klaviyo/Repos/klaviyo-swift-sdk'
pod 'KlaviyoCore', :path => '/Users/daniel.peluso/Klaviyo/Repos/klaviyo-swift-sdk'
```
it should now look like this:
image
- run `pod install` from the example/ios folder
- run `npx expo start` from the `example` dir and then type 's' to move away from an expo go build (should say development build). This needs to be running in a terminal or else you will get a white screen on the app.
- run `expo run:ios` from the example folder. *do not* run 'npm run clean-ios' or you will delete your ios folder entirely, which will wipe your podfile changes.

Sorry running this locally is a pain in the ass, I do think it's important for testing though so please reach out if you'd like some assistance.

Then you can either run the expo ios example app by opening the xcworkspace and running from xcode, or by running

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-20505

### DIFF
--- a/Sources/KlaviyoCore/Networking/NetworkSession.swift
+++ b/Sources/KlaviyoCore/Networking/NetworkSession.swift
@@ -35,8 +35,8 @@ public struct NetworkSession {
 
     public static let networkTimeout: UInt64 = 10_000_000_000 // in nanoseconds (10 seconds)
 
-    private static func getPluginConfiguration() -> (name: String, version: String)? {
-        guard let plistURL = Bundle.main.url(forResource: "klaviyo-plugin-configuration", withExtension: "plist") else {
+    static func getPluginConfiguration(bundle: Bundle = .main) -> (name: String, version: String)? {
+        guard let plistURL = bundle.url(forResource: "klaviyo-plugin-configuration", withExtension: "plist") else {
             return nil
         }
 
--- a/Sources/KlaviyoCore/Networking/NetworkSession.swift
+++ b/Sources/KlaviyoCore/Networking/NetworkSession.swift
@@ -35,40 +35,38 @@ public struct NetworkSession {
 
     public static let networkTimeout: UInt64 = 10_000_000_000 // in nanoseconds (10 seconds)
 
-    static func getPluginConfiguration(bundle: Bundle = .main) -> (name: String, version: String)? {
-        guard let plistURL = bundle.url(forResource: "klaviyo-plugin-configuration", withExtension: "plist") else {
+    private static func getPluginConfiguration(bundle: Bundle = .main) -> (name: String, version: String)? {
+        print(type(of: bundle))
+        print(bundle.url(forResource: "klaviyo-plugin-configuration", withExtension: "plist"))
+        guard let plistURL = bundle.url(forResource: "klaviyo-plugin-configuration", withExtension: "plist"),
+              let plistData = try? Data(contentsOf: plistURL),
+              let plist = try? PropertyListSerialization.propertyList(from: plistData, options: [], format: nil) as? [String: Any]
+        else {
             return nil
         }
 
-        guard let plistData = try? Data(contentsOf: plistURL) else {
-            return nil
-        }
-
-        guard let plist = try? PropertyListSerialization.propertyList(from: plistData, options: [], format: nil) as? [String: Any] else {
-            return nil
-        }
-
-        guard let pluginName = plist["klaviyo_sdk_plugin_name_override"] as? String else {
-            return nil
-        }
-        guard let pluginVersion = plist["klaviyo_sdk_plugin_version_override"] as? String else {
+        guard let pluginName = plist["klaviyo_sdk_plugin_name_override"] as? String,
+              let pluginVersion = plist["klaviyo_sdk_plugin_version_override"] as? String
+        else {
             return nil
         }
 
         return (name: pluginName, version: pluginVersion)
     }
 
-    public static let defaultUserAgent = { () -> String in
+    public static var defaultUserAgent: String = defaultUserAgent(bundle: .main)
+
+    internal static func defaultUserAgent(bundle: Bundle) -> String {
         let appContext = environment.appContextInfo()
         let klaivyoSDKVersion = "klaviyo-\(environment.sdkName())/\(environment.sdkVersion())"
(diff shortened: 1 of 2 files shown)


Sources/KlaviyoCore/Networking/NetworkSession.swift
```diff
@@ -35,10 +35,39 @@ public struct NetworkSession {
 
     public static let networkTimeout: UInt64 = 10_000_000_000 // in nanoseconds (10 seconds)
 
+    static func getPluginConfiguration(bundle: Bundle = .main) -> (name: String, version: String)? {
```
ab1470, line 38: this is specific to Expo, right? Can we make that clear, whether in the function name (like `getExpoPluginConfiguration`) and/or in documentation?

Sources/KlaviyoCore/Networking/NetworkSession.swift
```diff
@@ -35,10 +35,39 @@ public struct NetworkSession {
...
+        return (name: pluginName, version: pluginVersion)
+    }
+
     public static let defaultUserAgent = { () -> String in
         let appContext = environment.appContextInfo()
         let klaivyoSDKVersion = "klaviyo-\(environment.sdkName())/\(environment.sdkVersion())"
-        return "\(appContext.executable)/\(appContext.appVersion) (\(appContext.bundleId); build:\(appContext.appBuild); \(appContext.osVersionName)) \(klaivyoSDKVersion)"
+        var userAgent = "\(appContext.executable)/\(appContext.appVersion) (\(appContext.bundleId); build:\(appContext.appBuild); \(appContext.osVersionName)) \(klaivyoSDKVersion)"
+        
+        if let pluginConfig = getPluginConfiguration() {
```
ab1470, line 66:
```suggestion
        // if the SDK is running through an Expo plugin, we want to append the plugin info
        if let pluginConfig = getPluginConfiguration() {
```
not sure if that's exactly the correct way to phrase what's going on but it would help to have a comment explaining why we're adding the plugin config

Tests/KlaviyoCoreTests/NetworkSessionTests.swift
```diff
@@ -32,4 +32,96 @@ class NetworkSessionTests: XCTestCase {
         assertSnapshot(matching: data, as: .dump)
         assertSnapshot(matching: response, as: .dump)
     }
+
+    func testGetPluginConfigurationWithValidPlist() {
+        // Create a temporary plist file
```
ab1470, line 37: nice

Sources/KlaviyoCore/Networking/NetworkSession.swift
ab1470, line 38:
```suggestion
    private static func getPluginConfiguration(bundle: Bundle = .main) -> (name: String, version: String)? {
```
looks like this can be private. Unfortunately you'll need to revise your tests if you make it private, but the updates shouldn't be too difficult.

ab1470 (changes requested):

Sources/KlaviyoCore/Networking/NetworkSession.swift
dan-peluso: Not necessarily - it's for any framework that's being used to build a react native app. So if some expo competitor comes along and claims it has a better system for building RN apps we'd want to be able to use this same method of local resource overriding

Sources/KlaviyoCore/Networking/NetworkSession.swift
```diff
@@ -35,40 +35,38 @@ public struct NetworkSession {
 
     public static let networkTimeout: UInt64 = 10_000_000_000 // in nanoseconds (10 seconds)
 
-    static func getPluginConfiguration(bundle: Bundle = .main) -> (name: String, version: String)? {
-        guard let plistURL = bundle.url(forResource: "klaviyo-plugin-configuration", withExtension: "plist") else {
+    private static func getPluginConfiguration(bundle: Bundle = .main) -> (name: String, version: String)? {
+        print(type(of: bundle))
+        print(bundle.url(forResource: "klaviyo-plugin-configuration", withExtension: "plist"))
```
ab1470, line 40: oops
```suggestion
```

Tests/KlaviyoCoreTests/NetworkSessionTests.swift
```diff
@@ -52,55 +52,53 @@ class NetworkSessionTests: XCTestCase {
                 options: 0
             )
             try plistData.write(to: plistURL)
-            
+            print(plistURL)
```
ab1470, line 55:
```suggestion
```

ab1470 (approved):
looks good (other than removing the print statements). Thanks for making those changes

Sources/KlaviyoCore/Networking/NetworkSession.swift
dan-peluso: feeling dangerous might see how commit from pr handles this lol

PR #343, opened 2025-05-19
Add unregisterFromInAppForms

belleklaviyo:
# Description
Kills webview, removes listeners, resets state + added tests

## Due Diligence

## Changelog / Code Overview
- Adds `unregisterFromInAppForms` method

## Test Plan
- [ ] Open the test app and trigger registerForInAppForms
- [ ] Observe form appears/lifecycle listeners begin listening and printing events with backgrounding/foregrounding. You can use the Safari developer window to confirm the webview being alive
- [ ] Trigger unregisterForInAppForms and observe the Safari developer window closes as there is no more webview. Observe backgrounding/foregrounding produces no logs of injected events

https://github.com/user-attachments/assets/d4bb6159-c92f-4a4b-bc42-c26edf41492e

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-19954

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -52,6 +52,18 @@ class IAFPresentationManager {
         }
     }
 
+    func destroyWebviewAndListeners() {
+        destroyWebView()
+        isLoading = false
+        UserDefaults.standard.removeObject(forKey: "lastBackgrounded")
+        lifecycleCancellable?.cancel()
+        lifecycleCancellable = nil
+        apiKeyCancellable?.cancel()
+        apiKeyCancellable = nil
+        formEventTask?.cancel()
+        formEventTask = nil
+    }
+
     func setupLifecycleEvents(configuration: IAFConfiguration) {
         lifecycleCancellable = environment.appLifeCycle.lifeCycleEvents()
             .sink { [weak self] event in
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -52,18 +52,6 @@ class IAFPresentationManager {
         }
     }
 
-    func destroyWebviewAndListeners() {
-        destroyWebView()
-        isLoading = false
-        UserDefaults.standard.removeObject(forKey: "lastBackgrounded")
-        lifecycleCancellable?.cancel()
-        lifecycleCancellable = nil
-        apiKeyCancellable?.cancel()
-        apiKeyCancellable = nil
-        formEventTask?.cancel()
-        formEventTask = nil
-    }
-
     func setupLifecycleEvents(configuration: IAFConfiguration) {
         lifecycleCancellable = environment.appLifeCycle.lifeCycleEvents()
             .sink { [weak self] event in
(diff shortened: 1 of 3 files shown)


Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -208,6 +208,21 @@ class IAFPresentationManager {
             self?.viewController = nil
         }
     }
+
+    func destroyWebviewAndListeners() {
+        if #available(iOS 14.0, *) {
+            Logger.webViewLogger.info("UnregisterFromInAppForms; destroying webview and listeners")
+        }
+        isLoading = false
+        UserDefaults.standard.removeObject(forKey: "lastBackgrounded")
```
belleklaviyo, line 217: was left as a TBD in the jira ticket but I am assuming/makes the most sense to me to kill the app session by removing the time stamp when we call this method -- (and [this](https://klaviyo.slack.com/archives/C08GN96SF1B/p1747417166602959?thread_ts=1747235903.748809&cid=C08GN96SF1B) Slack message as other context?) but if not could remove this line

Sources/KlaviyoForms/KlaviyoSDK+Forms.swift
```diff
@@ -19,6 +19,15 @@ extension KlaviyoSDK {
         }
     }
 
+    @MainActor
```
ajaysubra, line 22: 🔍  wondering if we should get in the habit of adding docc comments on our APIs?

ajaysubra: if you do, may I ask you do the same for `registerForInAppForms`?

ajaysubra (approved):
LGTM with just a nit

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
evan-masseau: i think thats right. Beyond that though, if session isn't persisted beyond termination, do we need to write this to user defaults even? Keeping it in memory would be enough i think

evan-masseau (approved):

Sources/KlaviyoForms/KlaviyoSDK+Forms.swift
ab1470: > 🔍  wondering if we should get in the habit of adding docc comments on our APIs? 

Agreed. And Cursor often does a pretty good job with documentation so that should make it easier

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -208,6 +208,21 @@ class IAFPresentationManager {
             self?.viewController = nil
         }
     }
+
+    func destroyWebviewAndListeners() {
```
ab1470, line 212: Is the only reason this isn't `private` so that we can unit test it?

ab1470:
Is there a situation where we would want to use just `destroyWebView` instead of `destroyWebViewAndListeners`?

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
belleklaviyo: well I made it similar to `setupLifecycleEvents` and `constructWebview` which are called in `registerForInAppForms` and are both not private

ab1470: Good point, although I think those should both probably be private. We can address those other methods separately, but I think `destroyWebViewAndListeners` should be private. I'd rather not expose it to other parts of the SDK if we don't have to. Then in your unit tests you can call `unregisterFromInAppForms` instead of `destroyWebViewAndListeners`

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
belleklaviyo: that is a good point... yeah that makes sense to me!

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
belleklaviyo: Agreed -- this will be handled in a separate PR as it will involve some slightly involved refactoring for the tests

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
belleklaviyo: I'll update that in a separate PR as similar to the below comment about making the `destroyWebviewAndListeners` method private, it will affect how we have our tests setup.

PR #346, opened 2025-05-29
Enable toggling web/in-app data env

belleklaviyo:
# Description
Enables us to inject a `data-forms-data-environment` attribute that is either set to `in-app` explicitly or `web`. By default, if we do not inject it we will load the in-app forms data. This is toggle-able in the test app updated [here](https://github.com/klaviyo/klaviyo-ios-test-app/pull/78). In production we are defaulting this value to nil so it is never injected as we only care about in-app forms right now.

## Due Diligence

## Changelog / Code Overview
- Added `formsDataEnvironment` on the `KlaviyoEnvironment` that when populated will inject a `data-forms-data-environment` attribute to target different forms data.

## Test Plan
By default, no `data-forms-data-environment` is injected. In-app forms appear when triggered.

When explicitly set (via toggled) to web, `data-forms-data-environment = web` is injected. Web forms appear when triggered.

When explicitly set (via toggled) to in-app, `data-forms-data-environment = in-app` is injected. In-app forms appear when triggered.

https://github.com/user-attachments/assets/a4771f4c-f29e-40a1-9cb9-02a6b049cb81

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-20018

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
@@ -109,8 +109,9 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
         loadScripts?.insert(sdkNameWKScript)
         loadScripts?.insert(sdkVersionWKScript)
         loadScripts?.insert(handshakeWKScript)
-        guard let dataEnvironmentWKScript else { return }
-        loadScripts?.insert(dataEnvironmentWKScript)
+        if let dataEnvironmentWKScript { 
+            loadScripts?.insert(dataEnvironmentWKScript)
+        }
     }
 
     // MARK: - Loading
--- a/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
@@ -74,6 +74,13 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
     }
 
     @MainActor
+    private var dataEnvironmentWKScript: WKUserScript {
+        var environment = environment.formsDataEnvironment()
+        let sdkVersionScript = "document.head.setAttribute('data-forms-data-environment', '\(environment)');"
+        return WKUserScript(source: sdkVersionScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
+    }
+
+    @MainActor
     private var handshakeWKScript: WKUserScript {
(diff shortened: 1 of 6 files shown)


Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -73,6 +73,15 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
...
 
+    @MainActor
+    private var dataEnvironmentWKScript: WKUserScript {
+        guard let formsEnv = environment.formsDataEnvironment() else {
+            return WKUserScript(source: "", injectionTime: .atDocumentEnd, forMainFrameOnly: true)
+        }
+        let sdkVersionScript = "document.head.setAttribute('data-forms-data-environment', '\(formsEnv)');"
+        return WKUserScript(source: sdkVersionScript, injectionTime: .atDocumentEnd, forMainFrameOnly: true)
+    }
+
```
evan-masseau, line 84: Two kinda nit picks:
1. Does creating a WKUserScript object have to run on the main thread?
2. Wouldn't it be cleaner to just not inject anything, rather than inject an empty string script?

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -102,6 +109,8 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
         loadScripts?.insert(sdkNameWKScript)
         loadScripts?.insert(sdkVersionWKScript)
         loadScripts?.insert(handshakeWKScript)
+        guard let dataEnvironmentWKScript else { return }
+        loadScripts?.insert(dataEnvironmentWKScript)
```
ab1470, line 113:
```suggestion
        if let dataEnvironmentWKScript { 
            loadScripts?.insert(dataEnvironmentWKScript)
        }
```
I think it's a bit cleaner this way

ab1470 (approved):

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
belleklaviyo: > if we were to add a new script below line 113

yeah that totally makes sense! guess I was implicitly enforcing that by putting it at the end but that doesn't necessarily safeguard it for any future additions/changes so I agree this more explicit conditional insert is better and clearer (probably was my personal bias of trying to keep it all as one indent level lol)

PR #350, opened 2025-06-02
Prevent double calls to constructWebview

belleklaviyo:
# Description
Prevent `registerForInAppForms` from creating a new webview if there is already one active

## Due Diligence

## Changelog / Code Overview
- Add a guard check that there is not already an active viewController when `registerForInAppForms` is called

## Test Plan

https://github.com/user-attachments/assets/925e4468-0381-476a-8ec4-92ddc240e806

## Related Issues/Tickets
Slack discussion [here](https://klaviyo.slack.com/archives/C08GN96SF1B/p1748638808354229)

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -54,6 +54,13 @@ class IAFPresentationManager {
             return
         }
 
+        guard viewController == nil else {
+            if #available(iOS 14.0, *) {
+                Logger.webViewLogger.log("Webview is already set up; ignoring request.")
+            }
+            return
+        }
+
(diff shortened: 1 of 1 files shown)


Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -54,6 +54,13 @@ class IAFPresentationManager {
             return
         }
 
+        guard viewController == nil else {
+            if #available(iOS 14.0, *) {
+                Logger.webViewLogger.log("Webview is already set up; ignoring request.")
+            }
+            return
+        }
+
```
evan-masseau, line 63: Is there any in-between time that needs to be accounted for, where a webview has been created, but the `viewController` is not?

PR #363, opened 2025-06-16
Check viewController for unofficial backgrounding events

belleklaviyo:
# Description
When we pull down on the notification/control center or slightly slide up on the home bar (not fully backgrounding the app) -- we do not receive a backgrounded lifecycle event. We do however, upon closing the notification/control center or releasing the home bar so we return to the app, receive a foregrounded lifecycle event. We only receive a backgrounded lifecycle event when we actually fully background the app.

This was causing an issue where since we did not have any previous backgrounded time stamp for these cases, the resuming foregrounded event would be treated as a fresh launch and create a new webview. This change checks if there is already a viewController making the conditions for launching a new webview more strict and accurate. It also means we are definitely not considering opening the notification/control center as a backgrounding event (a previous assumption)

## Due Diligence

## Changelog / Code Overview
- Adds a check if a viewController already exists or not to account for unofficial backgrounding events like opening the notification/control center.

## Related Issues/Tickets
Addresses bugs reported [here](https://klaviyo.enterprise.slack.com/lists/T029KGL8E/F090YAV31JT?record_id=Rec091EQ1NC3W) and [here](https://klaviyo.enterprise.slack.com/lists/T029KGL8E/F090YAV31JT?record_id=Rec091CKG9LJH)

https://klaviyo.atlassian.net/browse/CHNL-22084

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -117,11 +117,14 @@ class IAFPresentationManager {
                                 try await self.createFormAndAwaitFormEvents(apiKey: apiKey)
                             }
                         } else {
-                            // launching
                             try await self.handleLifecycleEvent("foreground")
-
-                            let apiKey = try await KlaviyoInternal.fetchAPIKey()
-                            try await self.createFormAndAwaitFormEvents(apiKey: apiKey)
+                            // to prevent case when app is reforegrounded from opening the notification/control center
+                            // no backgrounded lifecycle event is dispatched in that case, only foregrounded event
+                            if self.viewController == nil {
+                                // fresh launch
+                                let apiKey = try await KlaviyoInternal.fetchAPIKey()
+                                try await self.createFormAndAwaitFormEvents(apiKey: apiKey)
+                            }
                         }
                     case .backgrounded:
                         self.lastBackgrounded = Date()


ajaysubra (approved):

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -117,11 +117,14 @@ class IAFPresentationManager {
                                 try await self.createFormAndAwaitFormEvents(apiKey: apiKey)
                             }
                         } else {
-                            // launching
                             try await self.handleLifecycleEvent("foreground")
-
```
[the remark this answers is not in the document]
belleklaviyo: good point yup can do

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -113,15 +113,16 @@ class IAFPresentationManager {
...
+                                try await self.initializeFormWithAPIKey()
                             }
                         } else {
-                            // launching
                             try await self.handleLifecycleEvent("foreground")
-
-                            let apiKey = try await KlaviyoInternal.fetchAPIKey()
-                            try await self.createFormAndAwaitFormEvents(apiKey: apiKey)
+                            // to prevent case when app is reforegrounded from opening the notification/control center
+                            // no backgrounded lifecycle event is dispatched in that case, only foregrounded event
```
ab1470, line 121:
```suggestion
                            // when opening Notification/Control Center, the system will not dispatch a `backgrounded` event,
                            // but it will dispatch a `foregrounded` event when Notification/Control Center is dismissed.
                            // This check ensures that don't reinitialize in this situation
```

ab1470: Assuming my description is accurate, I think it's a little more clear

PR #366, opened 2025-06-17
Add validation of sessionTimeoutDuration

belleklaviyo:
# Description
Adds validation to sessionTimeoutDuration to guard against negative numbers. Android also guards against overflow, but that isn't an issue here since we are using TimeInterval which is safe and provides the `.inifinity` value. Tried to copy over the comments from Android SDK for parity sake.

## Due Diligence

## Changelog / Code Overview
- Added validation to sessionTimeoutDuration

## Test Plan
- Calling registerForInAppForms with a negative TimeDuration throws an error and is treated as a 0s
- Calling registerForInAppForms with any other TimeDuration works as expected

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-21473

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
@@ -245,10 +245,13 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
             ()
         case .formWillAppear:
             if #available(iOS 14.0, *) {
-                Logger.webViewLogger.info("Received `formWillAppear` event from KlaviyoJS")
+                Logger.webViewLogger.info("Received 'formWillAppear' event from KlaviyoJS")
             }
             formLifecycleContinuation.yield(.present)
         case .formDisappeared:
+            if #available(iOS 14.0, *) {
+                Logger.webViewLogger.info("Received 'formDisappeared' event from KlaviyoJS")
+            }
             formLifecycleContinuation.yield(.dismiss)
         case let .trackProfileEvent(data):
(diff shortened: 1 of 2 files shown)


Sources/KlaviyoForms/InAppForms/Models/InAppFormsConfig.swift
```diff
@@ -5,13 +5,23 @@
 //  Created by Isobelle Lim on 5/7/25.
 //
 import Foundation
+import OSLog
 
 /// Configuration for in-app forms
 public struct InAppFormsConfig {
+    /// - Parameter sessionTimeoutDuration: Duration (in seconds) of user inactivity after which the form session is terminated.
+    ///   Defaults to 1 Hour, must be non-negative. Use 0 to timeout as soon as the app is backgrounded.
+    ///   To disable session timeout altogether, use TimeInterval.infinity.
```
evan-masseau, line 14: oh actually just out of curiosity, do the places where we _use_ sessionTimeoutDuration need any update for when it is set to infinity, or will it just work as written.

evan-masseau: oh wait using time interval is way better than using Long, maybe I should update to kotlin's Duration type before we release

ajaysubra (approved):

Sources/KlaviyoForms/InAppForms/Models/InAppFormsConfig.swift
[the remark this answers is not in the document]
belleklaviyo: ahhh I see okay that makes sense!

PR #370, opened 2025-06-20
[CHNL-22218] evaluate timers natively

ab1470:
# Description
This PR is to address [an issue](https://klaviyo.enterprise.slack.com/lists/T029KGL8E/F090YAV31JT?record_id=Rec091CJ7EM0V) that @Aarav-Klaviyo found while testing delays on In-App Forms, in which the form will not appear if the delay is set to anything longer than 4 seconds. This bug was only occurring on physical devices, not on the simulator. 

We already addressed this issue on iOS 17.0 and later in #369.

This PR addresses this issue for iOS <17.0. The fix involves injecting a script into the IAF web view that will relay any timers set within KlaviyoJS to the native Swift layer. Then the Swift layer sets its own timers with durations matching the duration of the timers communicated from the web view. When the native Swift layer timers elapse, they'll evaluate a simple no-op Javascript event, simply to "wake up" the web view, which will cause any of the suspended web view timers to resume and the form to display.

## Due Diligence

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/Assets/timerScript.js
+++ b/Sources/KlaviyoForms/InAppForms/Assets/timerScript.js
@@ -1,0 +1,40 @@
+(function () {
+      // Get the native bridge by name
+      const nativeBridgeName = document.head.getAttribute("data-native-bridge-name");
+
+      // Save original implementations
+      const _nativeSetTimeout = window.setTimeout;
+      const _nativeSetInterval = window.setInterval;
+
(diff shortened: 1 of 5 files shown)


evan-masseau (commented):
Thanks for sharing the code, and great to know this approach works. I'd say let's keep this in our back pocket for now since testing does seem to indicate that iOS<17 is unaffected, and iOS>=17 is covered by your config fix

PR #375, opened 2025-06-26
[CHNL-22412] display forms beneath alerts

ab1470:
# Description
This PR addresses a defect in which a form could be displayed within a UIAlertController. This addresses that issue by detecting if the top-most view controller is a UIAlertController. If so, we find the view controller that presented the alert, and insert our form view controller as a child of that presenting view controller.

[CHNL-22412](https://klaviyo.atlassian.net/browse/CHNL-22412)

## Due Diligence

## Test Plan

Here's a video showing the form working as expected when an alert is shown on the screen:

https://github.com/user-attachments/assets/f42f5704-408e-4b3f-9b00-124fa7663e27

## Related Issues/Tickets

[CHNL-22412]: https://klaviyo.atlassian.net/browse/CHNL-22412?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -313,18 +313,73 @@ class IAFPresentationManager {
             return
         }
 
-        if topController.isKlaviyoVC || topController.hasKlaviyoVCInStack {
-            if #available(iOS 14.0, *) {
-                Logger.webViewLogger.warning("In-App Form is already being presented; ignoring request")
+        // If the top controller is an alert controller, add our form view directly to the view hierarchy
+        if topController is UIAlertController {
+            guard let presentingController = topController.presentingViewController else {
+                if #available(iOS 14.0, *) {
+                    Logger.webViewLogger.warning("Unable to find presenting controller for UIAlertController; ignoring `presentForm()` request.")
+                }
+                return
             }
+
+            if presentingController.isKlaviyoVC || presentingController.hasKlaviyoVCInStack {
+                if #available(iOS 14.0, *) {
+                    Logger.webViewLogger.warning("In-App Form is already being presented; ignoring request")
+                }
+                return
+            }
+
+            // Check if view is already in hierarchy
+            if viewController.view.superview != nil {
+                if #available(iOS 14.0, *) {
+                    Logger.webViewLogger.warning("Form view is already in view hierarchy; ignoring request")
+                }
+                return
+            }
+
+            // Ensure presenting controller's view is loaded
+            guard presentingController.isViewLoaded else {
+                if #available(iOS 14.0, *) {
+                    Logger.webViewLogger.warning("Presenting controller's view is not loaded; ignoring request")
+                }
+                return
+            }
+
+            // Add the form view directly to the presenting controller's view
+            viewController.view.frame = presentingController.view.bounds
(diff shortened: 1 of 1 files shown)


Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -313,18 +313,73 @@ class IAFPresentationManager {
...
+                    Logger.webViewLogger.warning("Presenting controller's view is not loaded; ignoring request")
+                }
+                return
+            }
+
+            // Add the form view directly to the presenting controller's view
+            viewController.view.frame = presentingController.view.bounds
+            viewController.view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+            presentingController.view.addSubview(viewController.view)
+            presentingController.addChild(viewController)
```
evan-masseau, line 352: Am I reading this right, that now we're attaching our view to the host app's VC?

ab1470: *only if* the top-most VC is a UIAlertController, we'll identify the VC that was responsible for presenting that UIAlertController. Then we'll attach our web view to that presenting VC

evan-masseau: Guessing there is a limitation on presenting another VC while an alert is already showing? 
What if we waited for the alert to go away before presenting our form?

evan-masseau: I do realize I'm nitpicking over a pretty edge case scenario, sorry. This is a clever solution, my reservation is just the adding view into a VC we don't own. On Android I moved us to the activity-based architecture to avoid doing this kind of direct modification because it feels intrusive for a library to be doing. If that doesn't alarm you iOS folks with more experience than me, I can withdraw my objection.

ab1470: yeah I see what you're saying, I can imagine us getting into trouble with this approach if the presenting view controller isn't in a state that we expect it to be in. 

Gemini suggested a ("polite") polling approach, so @evan-masseau your idea has been validated by a machine
Screenshot 2025-06-27 at 15 27 35

evan-masseau: Haha, I also I asked gemini two different ways whether it is appropriate for a library to be attaching views directly to the hierarchy, got two fairly different answers depending on how I asked. The more "positive" answer seemed more relative to like a UI tools/components library library though. The other talked about how sticking to our own VC / Activity is better for encapsulation

evan-masseau: "Politely wait its turn" 😆

ab1470: I've only just started using gemini but so far I like its writing style; it's more graceful and has more color than the other models. Here's an analogy it gave my for why I shouldn't attach my view directly to the host app's view hierarchy:

Screenshot 2025-06-27 at 15 55 15

evan-masseau: cooool

ab1470:
closing this in favor of #376

PR #381, opened 2025-07-24
[CHNL-23448] allow endpoint-specific max retries

ab1470:
# Description
This PR modifies our networking logic to allow for endpoint-specific max retries. It also passes the number of max retries into the `X-Klaviyo-Attempt-Count` header; before this PR we were just hard-coding the max retries to 50, regardless of the actual number of max retries.

This is some prerequisite work for the universal linking feature. For that, we have an API call that we want to try only once; if it fails, we give up and don't try again. 

## Due Diligence

## Test Plan
<table>
    <tr><th>Scenario</th><th>Result</th></tr>
    <tr>
        <td>Max retries set to 2</td>
        <td>
Screenshot 2025-07-24 at 10 21 51
        </td>
    </tr>
    <tr>
        <td>Max retries set to 10</td>
        <td>
Screenshot 2025-07-24 at 10 22 36
        </td>
    </tr>
    <tr>
        <td>Max retries set to 1</td>
        <td>
Screenshot 2025-07-24 at 10 23 18
</td>
    </tr>
    <tr>
        <td>Max retries set to 2; API call is failing</td>
        <td>
Screenshot 2025-07-24 at 10 24 18 Although the first two calls have failed, we've already reached the max number of attempts. I validated that the SDK did not try the call again.
        </td>
    </tr>
</table>

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
+++ b/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
@@ -57,7 +57,7 @@ struct KlaviyoState: Equatable, Codable {
     var initalizationState = InitializationState.uninitialized
     var flushing = false
     var flushInterval = StateManagementConstants.wifiFlushInterval
-    var retryInfo = RetryInfo.retry(StateManagementConstants.initialAttempt)
+    var retryState = RetryState.retry(StateManagementConstants.initialAttempt)
     var pendingRequests: [PendingRequest] = []
     var pendingProfile: [Profile.ProfileKey: AnyEncodable]?
 
--- a/Tests/KlaviyoCoreTests/__Snapshots__/KlaviyoAPITests/testSuccessfulResponseWithEvent.1.txt
+++ b/Tests/KlaviyoCoreTests/__Snapshots__/KlaviyoAPITests/testSuccessfulResponseWithEvent.1.txt
@@ -12,7 +12,7 @@
     ▿ some: 1 key/value pair
       ▿ (2 elements)
         - key: "X-Klaviyo-Attempt-Count"
-        - value: "0/50"
+        - value: "1/50"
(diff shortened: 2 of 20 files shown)


Tests/KlaviyoCoreTests/__Snapshots__/KlaviyoAPITests/testSuccessfulResponseWithEvent.1.txt
```diff
@@ -12,7 +12,7 @@
     ▿ some: 1 key/value pair
       ▿ (2 elements)
         - key: "X-Klaviyo-Attempt-Count"
-        - value: "0/50"
+        - value: "1/50"
```
ndurell, line 15: Was this a bug before?

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -57,7 +57,7 @@ struct KlaviyoState: Equatable, Codable {
     var initalizationState = InitializationState.uninitialized
     var flushing = false
     var flushInterval = StateManagementConstants.wifiFlushInterval
-    var retryInfo = RetryInfo.retry(StateManagementConstants.initialAttempt)
+    var retryState = RetryState.retry(StateManagementConstants.initialAttempt)
```
ndurell, line 60: I think this is safe to do since it's not serialized.

ab1470:
> Why did we decide not to retry these out of curiosity?

For universal links? If we're unable to resolve the destination URL the first time, we don't want to try again because it would create a poor user experience to navigate to a deep link well after the fact.

Imagine you tap a deep link in an email. The phone is offline or for some other reason we can't resolve the destination URL. Your phone comes back online a few minutes later, the request goes through, and suddenly you're navigated to a deep link. We want to avoid that situation

PR #382, opened 2025-07-25
CHNL-22957: Inject profile events into webview

belleklaviyo:
# Description
This adds support for injecting profile events into the webview. This adds a publisher to KlaviyoInternal to publish profile events that hook into where we are enqueuing events. Then IAFPresentationManager subscribes to these events and injects the metric along with its properties.

## Due Diligence

## Changelog / Code Overview
- Added profileEvent to handshake
- Added support to inject profile events into the webview

## Test Plan
image

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-22957

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/Models/IAFNativeBridgeEvent.swift
+++ b/Sources/KlaviyoForms/InAppForms/Models/IAFNativeBridgeEvent.swift
@@ -20,6 +20,7 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
     case handShook
     case analyticsEvent
     case lifecycleEvent
+    case profileEvent
     case profileMutation
 
     private enum CodingKeys: String, CodingKey {
@@ -38,6 +39,7 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
         case handShook
         case analyticsEvent
         case lifecycleEvent
+        case profileEvent
         case profileMutation
     }
 
@@ -72,6 +74,8 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
             self = .analyticsEvent
         case .lifecycleEvent:
             self = .lifecycleEvent
+        case .profileEvent:
+            self = .profileEvent
         case .profileMutation:
             self = .profileMutation
         }
@@ -123,6 +127,7 @@ extension IAFNativeBridgeEvent {
             .openDeepLink(URL(string: "https://example.com")!),
             .abort(""),
             .lifecycleEvent,
+            .profileEvent,
             .profileMutation
         ]
     }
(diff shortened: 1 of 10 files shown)


Sources/KlaviyoForms/InAppForms/Models/IAFNativeBridgeEvent.swift
```diff
@@ -123,6 +127,7 @@ extension IAFNativeBridgeEvent {
             .openDeepLink(URL(string: "https://example.com")!),
             .abort(""),
             .lifecycleEvent,
+            .profileEvent,
```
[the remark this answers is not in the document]
ab1470: I'm good with using a feature branch

Sources/KlaviyoSwift/KlaviyoInternal.swift
```diff
@@ -151,6 +153,28 @@ package enum KlaviyoInternal {
         profileDataSubject.send(.failure(.notInitialized))
     }
 
+    // MARK: - Profile Event methods
+
+    /// Publishes an event to subscribers.
+    ///
+    /// - Parameter event: the profile event to publish
+    package static func publishEvent(_ event: Event) {
```
ab1470, line 161: can this be `internal`? I think only `KlaviyoSwift` should be able to call this, right?

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -466,6 +466,9 @@ struct KlaviyoReducer: ReducerProtocol {
 
             event = event.updateEventWithState(state: &state)
 
+            // Publish the event to subscribers
+            KlaviyoInternal.publishEvent(event)
+
```
ab1470, line 471: hmmm, I don't think this implementation is really compatible with TCA's principles. The `publishEvent` action should be executed as an effect of the `enqueueEvent` action.

I ran it through AI and it suggested replacing line 496 (`return event.metric.name...`) with 
```swift
            let baseEffect = event.metric.name == ._openedPush ? EffectTask<KlaviyoAction>.task { .flushQueue } : .none
            
            // Publish the event to subscribers through an effect
            return .merge([
                baseEffect,
                .fireAndForget { KlaviyoInternal.publishEvent(event) }
            ])
```

I haven't validated this but I think it's at least a starting point

(@ndurell, any input here?)

ab1470 (changes requested):

Sources/KlaviyoSwift/KlaviyoInternal.swift
belleklaviyo: good point

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
ndurell: That sounds right though I don't really know what that side effect does. Also if it's truly a side effect could mock some of it in the environment and test it

belleklaviyo: Makes sense, thanks for catching -- fixed this and added a test

PR #385, opened 2025-07-30
[CHNL-23664] refactor networking

ab1470:
# Description
This PR refactors our networking layer to make it more flexible, which is necessary for our work on universal links. Previously our network layer was set up so that every base URL used `klaviyo.com`, but with universal links we'll need to be able to call an endpoint that uses the tracking link as its base URL. 

With this work, I've refactored KlaviyoEndpoint so that the API key is supplied as an associated value in the relevant enum cases, and I've added a function on KlaviyoEndpoint to generate the URLRequest.

[CHNL-23664](https://klaviyo.atlassian.net/browse/CHNL-23664)

## Due Diligence

## Test Plan
I used Proxyman to inspect the HTTP traffic and compare the output between the `master` branch and this branch. I validated that the HTTP requests are functionally identical.


belleklaviyo (approved):
cool makes sense to me

PR #411, opened 2025-09-18
[CHNL-25156] create universal link handling fallback

ab1470:
# Description
This PR creates a fallback mechanism for handling deep links in the event that the host application did not register a deep link handler. 

If a deep link handler has not been registered and a deep link is invoked, the code will:

1. attempt to manually call the deep link handler on the host application's App Delegate
**if that fails**, the code will:
2. attempt to manually call the deep link handler on the host application's Scene Delegate
**if that fails**, the code will:
3. attempt to handle the deep link using `UIApplication.shared.open(_:)`

If a handler is not registered and a deep link is invoked, the console will display a warning:
```
Attempting to handle universal link via a fallback mechanism.
For improved stability and future-proofing, please provide your own
deep link handler logic by calling `KlaviyoSDK().registerDeepLinkHandler(_:)`
on application launch. Refer to the Klaviyo Swift SDK's README for more details.
```

All of this logic has been added to a new `DeepLinkHandler` class.

https://klaviyo.atlassian.net/browse/CHNL-25156

## Due Diligence

## Test Plan
See validation of test cases [here](https://klaviyo.atlassian.net/browse/CHNL-25086?atlOrigin=eyJpIjoiMWFmYjRiYTUxZGI0NGIyODhiMGMxMGU2OWQ0OTk2NGEiLCJwIjoiaiJ9)

### DIFF
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -157,6 +157,10 @@ public struct KlaviyoSDK {
         }
     }
 
+    /// Handles a Klaviyo universal tracking link URL by resolving it to a destination URL asynchronously and invoking the registered Deep Link Handler or invoking the AppDelegate or SceneDelegate's link handling logic.
+    ///
+    /// - Parameter url: the Klaviyo universal tracking link URL.
+    /// - Returns: `true` if the URL is a valid Klaviyo universal tracking link; `false` otherwise.
     public func handleUniversalTrackingLink(_ url: URL) -> Bool {
         if !url.isUniversalTrackingUrl {
             if #available(iOS 14.0, *) {


Sources/KlaviyoCore/Utils/DeepLinkHandler.swift
```diff
@@ -0,0 +1,212 @@
...
+
+        // If that fails, fall back to opening with `UIApplication.shared.open(_:)`
+        await openWithUIApplicationAPI(url)
+    }
+
+    // MARK: - Private Routing Helpers
+
+    /// Routes the URL using the modern SceneDelegate API.
+    @MainActor
+    private static func routeWithSceneSessionActivation(url: URL) async -> Bool {
```
ab1470, line 98: FYI, working on some changes to this method right now

evan-masseau (approved):
Looks cool! As always take my approval with the grain of salt that I'm not well versed in the intracacies of swift/ios. I particularly looked over the DeepLinkHandler part. The logic for routing to appdelegate / scene delegate makes sense for sure!

Sources/KlaviyoCore/Utils/DeepLinkHandler.swift
```diff
@@ -81,81 +81,87 @@ public class DeepLinkHandler {
             """)
         }
 
-        // First try to route with the Scene Delegate
-        if await routeWithSceneSessionActivation(url: url) { return }
-
-        // If that fails, try to route with the App Delegate
+        // First try to route with the App Delegate
```
evan-masseau, line 84: Reading a bit more on the two delegates, I think this makes sense to try first. I had thought the opposite before, but I had an oversimplified idea of how the two delegate types were related (assumed they were basically mutually exclusive, but its more complicated than that)

belleklaviyo:
what was the bug in iOS 18? (re: the last commit)

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -157,8 +157,16 @@ public struct KlaviyoSDK {
         }
     }
 
-    public func handleUniversalTrackingLink(_ url: URL) {
+    public func handleUniversalTrackingLink(_ url: URL) -> Bool {
```
belleklaviyo, line 160: since this is on the public interface could you add a documentation comment for this too

evan-masseau (approved):
Thanks for the heads up about the changes, still cool by me!

Sources/KlaviyoCore/Utils/DeepLinkHandler.swift
ab1470: yeah SceneDelegate and AppDelegate actually aren't mutually exclusive. And a SceneDelegate-based app can still handle deep links with the AppDelegate method (although I don't think it should)

Initially I tried calling SceneDelegate before calling AppDelegate, but that was causing a bug that I wasn't able to resolve until I switched the order.

Ultimately it's up to the host app developer to ensure that they've got their SceneDelegate configured correctly, and that they're following Apple's dev documentation for setting up their universal link handling. If they adhere to Apple's guidelines, they'll *only* handle a universal link within their SceneDelegate. There's only so much we can to to guard against incorrect AppDelegate/SceneDelegate configurations.

That being said, I just played around with different permutations of handling the link within the AppDelegate or SceneDelegate or both, and my code worked fine in all cases

ab1470:
> what was the bug in iOS 18? (re: the last commit)

@belleklaviyo when I tested it on a SceneDelegate version of the iOS test app running iOS 18.6, the link would not get delivered to the sceneDelegate, so the link was getting dropped. 

(When I tried with all of the same conditions except changing the iOS version to 26.0, it worked 🤷‍♂️)

PR #414, opened 2025-09-22
[CHNL-25156] conditionally show "register deep link handler" warning

ab1470:
# Description
After discussing with @evan-masseau, we decided to remove the `registerDeepLinkHandler` method from the React Native SDK, because React Native has a pretty good built-in method for handling deep links and universal links. This means that the warning we display to consumers of our Swift SDK is not relevant to consumers of our React Native SDK. This PR suppresses that warning.

### DIFF
--- a/Sources/KlaviyoCore/Utils/DeepLinkHandler.swift
+++ b/Sources/KlaviyoCore/Utils/DeepLinkHandler.swift
@@ -58,8 +58,10 @@ public class DeepLinkHandler {
                 customDeepLinkHandler(url)
             }
         } else {
-            if #available(iOS 14.0, *) {
-                Logger.navigation.info("A deep link handler has not been registered.\nHandling URL: '\(url.absoluteString, privacy: .public)' using fallback deep link handler")
+            if !KlaviyoEnvironment.isReactNative {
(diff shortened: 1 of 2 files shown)


Sources/KlaviyoCore/Utils/DeepLinkHandler.swift
```diff
@@ -72,13 +74,15 @@ public class DeepLinkHandler {
...
     private static func openWithFallbackHandler(url: URL) async {
-        if #available(iOS 14.0, *) {
-            Logger.navigation.warning("""
-            Attempting to handle universal link via a fallback mechanism.
-            For improved stability and future-proofing, please provide your own
-            deep link handler logic by calling `KlaviyoSDK().registerDeepLinkHandler(_:)`
-            on application launch. Refer to the Klaviyo Swift SDK's README for more details.
-            """)
+        if KlaviyoEnvironment.isReactNative {
+            if #available(iOS 14.0, *) {
```
evan-masseau, line 78: Just curious, is it that #available checks can't be in the same `if` statement as a runtime check?

evan-masseau (approved):

PR #415, opened 2025-09-23
Monitor location permissions

belleklaviyo:
# Description
Removes the hack put in to trigger the permissions prompt from the SDK (this will be done in the host app) and is just reactive to the auth level change to set up or destroy geofencing. 

CLLocationManager is/was difficult to test, but I added a `LocationManagerProtocol` so we could mock the status and test events to abstract the functionality. This will set us up to write more tests later on related to event triggering.

## Due Diligence

## Changelog / Code Overview
- Monitors permission level to set up or remove geofencing

## Test Plan
https://github.com/user-attachments/assets/ae89fdda-bcf9-484d-97c8-4dce76c1690c

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-25112

### DIFF
--- a/Sources/KlaviyoLocation/KlaviyoLocationManager.swift
+++ b/Sources/KlaviyoLocation/KlaviyoLocationManager.swift
@@ -21,16 +21,11 @@ public class KlaviyoLocationManager: NSObject {
 
     internal init(locationManager: LocationManagerProtocol? = nil, geofenceManager: KlaviyoGeofenceManager? = nil) {
         self.locationManager = locationManager ?? CLLocationManager()
-        self.geofenceManager = geofenceManager ?? Self.createDefaultGeofenceManager(with: self.locationManager)
+        self.geofenceManager = geofenceManager ?? KlaviyoGeofenceManager(locationManager: self.locationManager)
 
         super.init()
         self.locationManager.delegate = self
         self.locationManager.allowsBackgroundLocationUpdates = true
-    }
-
-    private static func createDefaultGeofenceManager(with locationManager: LocationManagerProtocol) -> KlaviyoGeofenceManager {
-        let clLocationManager = locationManager as? CLLocationManager ?? CLLocationManager()
-        return KlaviyoGeofenceManager(locationManager: clLocationManager)
     }
 
     deinit {
--- a/Sources/KlaviyoLocation/KlaviyoGeofenceManager.swift
+++ b/Sources/KlaviyoLocation/KlaviyoGeofenceManager.swift
@@ -10,9 +10,9 @@ import KlaviyoCore
 import OSLog
 
 internal class KlaviyoGeofenceManager {
-    private let locationManager: CLLocationManager
+    private let locationManager: LocationManagerProtocol
 
-    internal init(locationManager: CLLocationManager) {
+    internal init(locationManager: LocationManagerProtocol) {
         self.locationManager = locationManager
     }
 


Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -8,23 +8,38 @@
...
+        self.locationManager = locationManager
+        self.geofenceManager = geofenceManager
+        super.init()
+        self.locationManager.delegate = self
+        // Don't set allowsBackgroundLocationUpdates in test environment to avoid simulator issues
+        if NSClassFromString("XCTest") == nil {
+            self.locationManager.allowsBackgroundLocationUpdates = true
+        }
+    }
+    #endif
```
belleklaviyo, line 41: had to do this bit of refactoring to use `LocationManagerProtocol` to set up this class and its functionality with CLLocationManager to be testable

evan-masseau (approved):
looks pretty solid to me

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -8,23 +8,38 @@
...
+    #if DEBUG
+    internal init(locationManager: LocationManagerProtocol, geofenceManager: KlaviyoGeofenceManager) {
+        self.locationManager = locationManager
+        self.geofenceManager = geofenceManager
+        super.init()
+        self.locationManager.delegate = self
+        // Don't set allowsBackgroundLocationUpdates in test environment to avoid simulator issues
+        if NSClassFromString("XCTest") == nil {
+            self.locationManager.allowsBackgroundLocationUpdates = true
+        }
```
evan-masseau, line 39: it does feel a little odd to have test code in the real file, could this be done with a dependency injection pattern where the arguments are either optional, and `init` builds them if not provided, or the arguments are required but have an appropriate default value?
And `allowsBackgroundLocationUpdates = true` could be set from the test / setup scripts so that this kind of setup code doesn't live in the file you're testing.

belleklaviyo: yeah agree, felt iffy about this -- but upon further testing thing I don't actually need to specify that condition for XCTests! that was a remaining condition when I was dealing with a different setup, but guess with the protocol it's not needed here anymore. removing now

evan-masseau: aaand its all gone, never mind haha

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -8,21 +8,29 @@
...
 import Foundation
+import KlaviyoCore
 import KlaviyoSwift
 import OSLog
 
 public class KlaviyoLocationManager: NSObject {
     static let shared = KlaviyoLocationManager()
 
-    private let locationManager = CLLocationManager()
+    private var locationManager: LocationManagerProtocol
```
ab1470, line 18:
```suggestion
    private let locationManager: LocationManagerProtocol
```
can this be a constant?

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -8,21 +8,29 @@
...
     private let geofenceManager: KlaviyoGeofenceManager
-    internal let geofencePublisher: PassthroughSubject<String, Never> = .init()
+    private let geofencePublisher: PassthroughSubject<String, Never> = .init()
+
+    internal init(locationManager: LocationManagerProtocol? = nil, geofenceManager: KlaviyoGeofenceManager? = nil) {
```
ab1470, line 22: In the end I think this is okay, but it gave me pause because making those params optional in the initializer obscures the fact that the `locationManager` and `geofenceManager` aren't actually optional. *Generally* I'd prefer to set the default values in the initializer:

```swift
    internal init(
        locationManager: LocationManagerProtocol = CLLocationManager(),
        geofenceManager: KlaviyoGeofenceManager = Self.createDefaultGeofenceManager(with: self.locationManager)
    ) {
        self.locationManager = locationManager
        self.geofenceManager = geofenceManager
    }
```

to make it clear that those aren't optional and they have default values, but my code above doesn't compile because of the reference to `self.locationManager` when you instantiate the KlaviyoGeofenceManager.

**Tl;dr:** it's fine, leave it

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -8,21 +8,29 @@
...
-        geofenceManager = KlaviyoGeofenceManager(locationManager: locationManager)
         super.init()
-        locationManager.delegate = self
-        locationManager.allowsBackgroundLocationUpdates = true
+        self.locationManager.delegate = self
+        self.locationManager.allowsBackgroundLocationUpdates = true
+    }
+
+    private static func createDefaultGeofenceManager(with locationManager: LocationManagerProtocol) -> KlaviyoGeofenceManager {
+        let clLocationManager = locationManager as? CLLocationManager ?? CLLocationManager()
```
ab1470, line 32: If you modify the `KlaviyoGeofenceManager` to use the `LocationManagerProtocol` instead of a concrete `CLLocationManager` type, you can (a) make the `KlaviyoGeofenceManager` more testable, and (b) remove this optional casting line with a default fallback, and you can actually remove the entire `createDefaultGeofenceManager` method. 

So `KlaviyoGeofenceManager` would look like this:
```swift
internal class KlaviyoGeofenceManager {
    private let locationManager: LocationManagerProtocol

    internal init(locationManager: LocationManagerProtocol) {
        self.locationManager = locationManager
    }

    ....
}
```

And then in this file you can replace line 24 with 
```swift
self.geofenceManager = geofenceManager ?? KlaviyoGeofenceManager(locationManager: self.locationManager)
```

and then delete `createDefaultGeofenceManager`

ab1470 (approved):
looks good, just a few minor comments

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
belleklaviyo: ahhhh okay nice this looks a lot better

PR #420, opened 2025-09-25
Add Swift release branch version validation workflow

evan-masseau:
Description

Adds GitHub action to validate that release branch versions match the version in Sources/KlaviyoCore/Utils/Version.swift. Similar to [Android SDK validation](https://github.com/klaviyo/klaviyo-android-sdk/blob/732240ae050bf223f2426a709c96367f267efb2b/.github/workflows/android-release.yml#L3) but adapted for Swift version detection.

🤖 Generated with [Claude Code](https://claude.ai/code)

## Due Diligence

## Test Plan

Tested it with a pull request trigger first:
Screenshot 2025-09-25 at 11 33 11

but we don't need to keep that trigger.


ab1470 (approved):
Nice! Should we also include version checks for all of the podfiles, or do we want to keep this scope narrow?

evan-masseau:
> Nice! Should we also include version checks for all of the podfiles, or do we want to keep this scope narrow?

@ab1470 yeah good idea, i skipped everything else that looked like test files, but podfiles would be good.

PR #437, opened 2025-10-20
Create Geofence events

belleklaviyo:
# Description

So this does a bit more than just creating geofence events but it all makes sense together:
- Adds the apiKey to the fetchGeofences call
- Updates path for fetchGeofences to `/client/events`
- Uses expected API revision
- Removes mock API call
- Adds guard to setupGeofencing() call to ensure we are initialized and not trying to fetch geofences for a nil/invalid companyId* (this is doubly necessary due to some lifecycle weirdness with CLLocationManager)
- Calls `startMonitoringSignificantLocationChanges` to ensure we can monitor location events in the backgrounded and terminated states
- Introduces an API function `monitorGeofencesFromBackground` to be put in `didFinishLaunchingWithOptions` to ensure we can monitor location events in the backgrounded and terminated states
- Modifies the location event to use the expected dimension names
- Modifies the `enqueueEvent` so if its a goefence event it automatically flushes the queue, regardless of initialization state

**Notes**
- Dwell was originally going to be implemented in this PR as well, but that will be done in a separate PR
- CoreLocation sometimes triggers two identical enter/exit events depending if you're foregrounded or not when triggering an event. We will implement a cooldown period to make sure 2 identical events are not fired within an unreasonable amount of time (more [here](https://klaviyo.slack.com/archives/C092G8ASQH1/p1761673093199389)). This will be done in a separate PR.
- We want to doubly ensure geofence events are delivered in a timely manner, so we want to insert those events at the front of the queue. This will be done in a separate PR.

## Due Diligence

## Changelog / Code Overview
This updates the event metrics going to `client/events` to the expected dimension names. 

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-25309

### DIFF
--- a/Tests/KlaviyoCoreTests/KlaviyoEndpointTests.swift
+++ b/Tests/KlaviyoCoreTests/KlaviyoEndpointTests.swift
@@ -188,5 +188,6 @@ final class KlaviyoEndpointTests: XCTestCase {
         // Then
         XCTAssertEqual(request.httpMethod, "GET")
         XCTAssertEqual(request.url?.path, "/client/geofences")
+        XCTAssertEqual(request.url?.query, "company_id=test_api_key")
     }
 }


Sources/KlaviyoLocation/Models/Geofence.swift
```diff
@@ -41,18 +44,21 @@ public struct Geofence: Equatable, Hashable, Codable {
...
+    ///   - dwell: Optional time in seconds to trigger a dwell event after entering and staying in the geofence for this duration. If not provided, time will default to 5 minutes.
     /// - Throws: `GeofenceError.invalidIdFormat` if the ID doesn't match the expected format
     public init(
         id: String,
         longitude: Double,
         latitude: Double,
         radius: Double,
+        dwell: Int? = 300
     ) throws {
         try Self.validateIdFormat(id)
```
ab1470: just came here to say that I love a failable/throwing initializer

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -90,44 +107,119 @@ extension KlaviyoLocationManager: CLLocationManagerDelegate {
...
 
         Task {
             await MainActor.run {
                 KlaviyoInternal.create(event: enterEvent)
-                geofencePublisher.send("Entered \(region.identifier)")
+                geofencePublisher.send("Entered \(klaviyoLocationId)")
+            }
+        }
+
+        Task {
```
ab1470: was there a reason for putting this in its own task, rather than using the one above it?

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -90,44 +107,119 @@ extension KlaviyoLocationManager: CLLocationManagerDelegate {
...
             await MainActor.run {
                 KlaviyoInternal.create(event: exitEvent)
-                geofencePublisher.send("Exited \(region.identifier)")
+                geofencePublisher.send("Exited \(klaviyoLocationId)")
             }
         }
+
+        Task {
+            await MainActor.run {
+                cancelDwellTimer(for: klaviyoLocationId)
```
ab1470: same question as above about putting this in its own task versus adding it to the existing one

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -90,44 +107,119 @@ extension KlaviyoLocationManager: CLLocationManagerDelegate {
...
+
+    private func cancelDwellTimer(for klaviyoLocationId: String) {
+        if let timer = dwellTimers[klaviyoLocationId] {
+            timer.invalidate()
+            dwellTimers.removeValue(forKey: klaviyoLocationId)
+
+            if #available(iOS 14.0, *) {
+                Logger.geoservices.info("🕐 Cancelled dwell timer for region \(klaviyoLocationId)")
+            }
+        }
```
ab1470:
```suggestion
        } else {
            if #available(iOS 14.0, *) {
                Logger.geoservices.info("🕐 No dwell timer found for region \(klaviyoLocationId)")
            }
        }
```

ab1470 (commented):
overall looks great, just some minor comments and suggested changes

Sources/KlaviyoSwift/Models/Event.swift
```diff
@@ -34,6 +35,16 @@ public struct Event: Equatable {
...
+
+        /// Returns true if this event is a geofence-related event
+        public var isGeofenceEvent: Bool {
+            switch name {
+            case .locationEvent(.geofenceEnter), .locationEvent(.geofenceExit), .locationEvent(.geofenceDwell):
+                return true
+            default:
+                return false
+            }
+        }
```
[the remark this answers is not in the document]
belleklaviyo: ah thanks I was trying to figure out a better syntax

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -90,43 +86,53 @@ extension KlaviyoLocationManager: CLLocationManagerDelegate {
...
+        guard let region = region as? CLCircularRegion,
+              let klaviyoLocationId = region.klaviyoLocationId else {
+            if #available(iOS 14.0, *) {
+                Logger.geoservices.warning("Received non-Klaviyo geofence notification. Skipping.")
+            }
+            return
+        }
         if #available(iOS 14.0, *) {
-            Logger.geoservices.info("🌎 User exited region \"\(region.identifier)\"")
+            Logger.geoservices.info("🌎 User exited region \"\(klaviyoLocationId)\"")
```
ab1470:
```suggestion
            Logger.geoservices.info("🌎 User exited region \"\(klaviyoLocationId, privacy: .public)\"")
```

Sources/KlaviyoLocation/KlaviyoGeofenceManager.swift
```diff
@@ -32,6 +38,13 @@ internal class KlaviyoGeofenceManager {
         }
 
         Task {
+            guard let _ = try? await KlaviyoInternal.fetchAPIKey() else {
+                if #available(iOS 14.0, *) {
+                    Logger.geoservices.info("SDK is not initialized, skipping geofence refresh")
+                }
+                return
+            }
+
```
belleklaviyo: So more context about this -- it's a sensible guard to put in anyways where we would expect there to be a valid API key to use with our `fetchGeofences` call. But there's some weirdness where when CLLocationManager() is instantiated (which we do on launch to ensure background/terminated support) it also triggers `didChangeAuthorization` even if the actual auth level has not changed. This fires before initialization is able to complete, so as a result when we trigger `setupGeofences` because we have the correct auth level, we may not have the actual API key. This results in fetching the geofences for a nil company which returns empty which then clears any existing geofences. 

tldr; we need this to make sure we can support being responsive to authorization level changes (such as if they revoke it and we need to unregister our geofences) while also carefully handling the timing issues between when CoreLocation is setting up and the app can be initialized

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -504,11 +504,12 @@ struct KlaviyoReducer: ReducerProtocol {
...
 
             /*
-             if we receive an opened push event we want to flush the queue right away so that
+             if we receive an opened push event or geofence events we want to flush the queue right away so that
              we don't miss any user engagement events. In all other cases we will flush the queue
              using the flush intervals defined above in `StateManagementConstants`
              */
-            let baseEffect = event.metric.name == ._openedPush ? EffectTask<KlaviyoAction>.task { .flushQueue } : .none
+            let baseEffect = event.metric.name == ._openedPush || event.metric.isGeofenceEvent
+                ? EffectTask<KlaviyoAction>.task { .flushQueue } : .none
```
belleklaviyo: @ajaysubra I know we have discussed extra ensuring our geofence events get processed by inserting the geofence events at the front of the queue. I'd want to handle that in a separate PR as this instant flush seems to handle it well so far in testing, and that work seems like an a further addition that can be tackled separately.

Sources/KlaviyoLocation/Models/Geofence.swift
ajaysubra: I know this is existing logic but does it make sense to move it out of the data model class keep data and logic separate?

Sources/KlaviyoLocation/KlaviyoGeofenceManager.swift
```diff
@@ -7,15 +7,21 @@
...
 
 internal class KlaviyoGeofenceManager {
     private let locationManager: LocationManagerProtocol
+    private weak var locationManagerDelegate: KlaviyoLocationManager?
 
     internal init(locationManager: LocationManagerProtocol) {
         self.locationManager = locationManager
     }
 
+    internal func setLocationManagerDelegate(_ delegate: KlaviyoLocationManager) {
```
ajaysubra: why is there a method to set the delegate?

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
belleklaviyo: https://klaviyo.atlassian.net/browse/CHNL-28059 done, here

Sources/KlaviyoLocation/KlaviyoGeofenceManager.swift
belleklaviyo: good point, realized this/locationManagerDelegate is completely unnecessary

ajaysubra (approved):
Nice work, Belle!

PR #441, opened 2025-11-05
Fix plist loading performance issue (#440)

### LINKED ISSUE #440: Hang because of React Native configuration
### Checklist

- [x] I have read the [contributing guidelines](https://github.com/klaviyo/klaviyo-swift-sdk/blob/master/.github/CONTRIBUTING.md)
- [ ] I have determined whether this bug is also reproducible in a vanilla SwiftUI project.
- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/klaviyo/klaviyo-swift-sdk/issues) or [discussion](https://github.com/pointfreeco/klaviyo/klaviyo-swift-sdk/discussions).

### Description

We have a very basic Klaviyo setup; a simplified version of our AppDelegate looks like that:

```swift
@main
final class AppDelegate: UIResponder, UIApplicationDelegate {
    func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
        KlaviyoSDK().initialze(with: "<redacted>")
        UIApplication.shared.registerForRemoteNotifications()
    }
}

extension AppDelegate: UNUserNotificationCenterDelegate {
    func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
        KlaviyoSDK().set(pushToken: deviceToken)
    }
}
```

We noticed a larger app hang ~500ms on device (~1.5s on simulator) during app startup and I could identify the Klaviyo SDK as the main root cause of this.

The `initialize(with:)` & `set(pushToken:)` start the underlying initialization process. Part of that is initializing the `KlaviyoEnvironment` with [SDK name and version](https://github.com/klaviyo/klaviyo-swift-sdk/blob/db49a64e9feceb6befd8e86bce54908f8daedead/Sources/KlaviyoCore/KlaviyoEnvironment.swift#L232-L233); these two type methods access the [rnSDKConfig](https://github.com/klaviyo/klaviyo-swift-sdk/blob/db49a64e9feceb6befd8e86bce54908f8daedead/Sources/KlaviyoCore/KlaviyoEnvironment.swift#L165) which is lazily initialized and invokes `loadPlist(named:)` the first time. In there it tries to [initialize a React Native specific bundle](https://github.com/klaviyo/klaviyo-swift-sdk/blob/db49a64e9feceb6befd8e86bce54908f8daedead/Sources/KlaviyoCore/Utils/FileUtils.swift#L80) which is blocking the current thread (Main) for a long time despite resulting in a nil instance.

I did a simple test which confirmed my theory; bundle is nil and it takes more than 1 second.
```swift
let date = Date.now
let bundle = Bundle(identifier: "org.cocoapods.klaviyo-react-native-sdk")
print(bundle, -date.timeIntervalSinceNow)
```

We do not use any configuration plists so we just get the [default value](https://github.com/klaviyo/klaviyo-swift-sdk/blob/db49a64e9feceb6befd8e86bce54908f8daedead/Sources/KlaviyoCore/KlaviyoEnvironment.swift#L171) but our current patch to this issue is to add one to our main bundle so that it is [returning before](https://github.com/klaviyo/klaviyo-swift-sdk/blob/db49a64e9feceb6befd8e86bce54908f8daedead/Sources/KlaviyoCore/Utils/FileUtils.swift#L79) the problematic bundle initialization.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
	<key>klaviyo_sdk_name</key>
	<string>swift</string>
	<key>klaviyo_sdk_version</key>
	<string>5.0.3</string>
</dict>
</plist>
```

### Expected behavior

No hang caused by the Klaviyo SDK initialization.

### Actual behavior

The Klaviyo SDK hangs during initialization.

### Steps to reproduce

A simple app setup as mentioned in the description should make this reproducible.

### The Klaviyo Swift SDK version information

5.0.3

### Destination operating system

iOS 18 / iOS 26 (including simulator)

### Xcode version information

16.4

### Swift Compiler version information

```shell
swift-driver version: 1.120.5 Apple Swift version 6.1.2 (swiftlang-6.1.2.1.2 clang-1700.0.13.5)
Target: arm64-apple-macosx15.0
```

ajaysubra:
## Problem

Non-React Native apps experience a 3+ second hang on first API call due to expensive `Bundle(identifier:)` lookup attempting to find the React Native framework bundle, even when React Native is not present.

## Background: PR #333

PR #333 added support for loading RN SDK configuration plists from two locations to handle both CocoaPods linking scenarios:
1. **Main bundle** (static linking - most common): Plist is copied to main bundle during build
2. **RN SDK bundle** (dynamic linking - edge case): Plist remains in framework bundle

This PR preserves that functionality while fixing the performance issue.

## Solution

Added fast `NSClassFromString("RCTBridge")` check (< 1ms) before attempting bundle lookup. For non-RN apps, returns immediately without expensive I/O operations.

**Performance Impact:**
- Before: 3075ms (SLOW)
- After: 0.1ms (EXCELLENT)
- RN iOS apps: Unaffected and performant. For static linking (most common), plist loads from main bundle quickly. For dynamic linking (edge case), the RN framework bundle is present so the lookup succeeds fast (~500-1500ms, but only once on first call).

### 1. Added RCTBridge Guard
`KlaviyoEnvironment.swift:165-182` - Check for React Native presence before plist loading:
```swift
private static let rnSDKConfig: [String: AnyObject] = {
    guard NSClassFromString("RCTBridge") != nil else {
        return [:] // Fast exit for non-RN apps
    }
    // Original PR #333 logic preserved...
}()
```

### 2. Refactored for Separation of Concerns
Split `loadPlist()` into two focused functions:
- `loadPlist(named:)` - Checks main bundle only (general purpose)
- `loadPlistFromReactNativeBundle(named:)` - Checks RN SDK bundle (RN-specific)

**Rationale:** FileUtils should be general-purpose utilities, not contain framework-specific logic. The caller (`KlaviyoEnvironment`) now owns all React Native detection logic.

### 3. Updated Tests
Added performance tests verifying quick completion (< 100ms) for non-RN scenarios.

## Testing

- Unit tests pass with performance validation
- RN SDK configuration loading still works for both static and dynamic linking
- Non-RN apps no longer experience startup hang

Fixes #440

### DIFF
--- a/Sources/KlaviyoCore/Utils/FileUtils.swift
+++ b/Sources/KlaviyoCore/Utils/FileUtils.swift
@@ -72,7 +72,7 @@ public func removeFile(at url: URL) -> Bool {
 /// Load plist from main application bundle
 /// - Parameter name: the name of the plist
 /// - Returns: the contents of the plist or nil if not found
-func loadPlist(named name: String) -> [String: AnyObject]? {
+package func loadPlist(named name: String) -> [String: AnyObject]? {
     guard let path = Bundle.main.path(forResource: name, ofType: "plist"),
           let dict = NSDictionary(contentsOfFile: path) as? [String: AnyObject] else {
         return nil
@@ -83,7 +83,7 @@ func loadPlist(named name: String) -> [String: AnyObject]? {
 /// Load plist from React Native SDK bundle (for dynamic linking scenarios)
 /// - Parameter name: the name of the plist
 /// - Returns: the contents of the plist or nil if not found
-func loadPlistFromReactNativeBundle(named name: String) -> [String: AnyObject]? {
+package func loadPlistFromReactNativeBundle(named name: String) -> [String: AnyObject]? {
     guard let bundle = Bundle(identifier: "org.cocoapods.klaviyo-react-native-sdk"),
           let path = bundle.path(forResource: name, ofType: "plist"),
           let dict = NSDictionary(contentsOfFile: path) as? [String: AnyObject] else {


Sources/KlaviyoCore/Utils/FileUtils.swift
```diff
@@ -69,27 +69,25 @@ public func removeFile(at url: URL) -> Bool {
     return false
 }
 
-/// load any plist from app main bundle or React Native framework bundle
+/// Load plist from main application bundle
 /// - Parameter name: the name of the plist
-/// - Returns: the contents of the plist in `[String: AnyObject]` or nil if not found
+/// - Returns: the contents of the plist or nil if not found
 func loadPlist(named name: String) -> [String: AnyObject]? {
```
ajaysubra: Making this method general purpose and moving RN specific logic to a specialized method.

ajaysubra:
Tested this using the test app, by adding this code to `KlaviyoClient`

```swift
extension KlaviyoClient: DependencyKey {
    static var wasInitialized = false

    // PERFORMANCE LOGGING: This static let will trigger plist loading when the class loads
    static let sdkVersion: String = {
        print("")
        print("========================================")
        print("🔍 PLIST LOADING TEST (Issue #440)")
        print("About to access environment SDK name/version (triggers plist loading)...")
        let startTime = Date()

        let name = environment.sdkName()
        let version = environment.sdkVersion()

        let elapsedMs = (Date().timeIntervalSince(startTime)) * 1000.0

        print("SDK Name: \(name)")
        print("SDK Version: \(version)")
        print("Time: \(String(format: "%.1f", elapsedMs)) ms")

        if elapsedMs < 100 {
            print("✅ EXCELLENT! Very fast (< 100ms)")
        } else if elapsedMs < 2000 {
            print("⚠️  ACCEPTABLE: Within expected range")
        } else {
            print("❌ SLOW! This indicates issue #440 is NOT fixed")
        }

        print("Expected: < 100ms (with fix), 3000ms+ (without fix)")
        print("========================================")
        print("")

        return version
    }()

    static let liveValue = Self(
        initialize: { companyId in
            // Trigger the sdkVersion static property to run the performance test
            _ = sdkVersion

            KlaviyoSDK().initialize(with: companyId)
            wasInitialized = true
        },
        
        ...
```

On testing with our latest swift SDK, 
```
========================================
🔍 PLIST LOADING TEST (Issue #440)
About to access environment SDK name/version (triggers plist loading)...
SDK Name: swift
SDK Version: 5.1.0
Time: 3123.0 ms
❌ SLOW! This indicates issue #440 is NOT fixed
Expected: < 100ms (with fix), 3000ms+ (without fix)
========================================
```

When testing with this branch,

```
========================================
:mag: PLIST LOADING TEST (Issue #440)
About to access environment SDK name/version (triggers plist loading)...
SDK Name: swift
SDK Version: 5.1.0
Time: 0.1 ms
:white_check_mark: EXCELLENT! Very fast (< 100ms)
Expected: < 100ms (with fix), 3000ms+ (without fix)
========================================
```

Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -162,7 +162,24 @@ public struct KlaviyoEnvironment {
...
+            return config
+        }
+
+        // Fallback to RN SDK bundle (dynamic linking - less common)
+        if let config = loadPlistFromReactNativeBundle(named: "klaviyo-sdk-configuration") {
+            return config
+        }
+
+        return [:]
+    }()
```
ndurell: Should this be more lazy? This loads when someone first access the environment. Not sure if delaying to when we actually need it would relieve some pressure on app start?

ndurell: Or maybe this won't try to access the plist until you try to read the var?

ndurell (approved):

Sources/KlaviyoCore/Utils/FileUtils.swift
```diff
@@ -69,27 +69,25 @@ public func removeFile(at url: URL) -> Bool {
...
-                  let path = reactNativeBundle.path(forResource: name, ofType: "plist") {
-            // If not found in main bundle, try loading from React Native framework bundle
-            return path
-        } else {
-            return nil
-        }
-    }()
+/// - Returns: the contents of the plist or nil if not found
+package func loadPlist(named name: String) -> [String: AnyObject]? {
+    guard let path = Bundle.main.path(forResource: name, ofType: "plist"),
```
amber-klaviyo: "guard" seems to be replacing the previous "if" logic. I'm not familiar with swift / guard. What scenarios is guard preferred? Is there a performance benefit to guard over if here?

Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -162,7 +162,24 @@ public struct KlaviyoEnvironment {
             .eraseToAnyPublisher()
     }
 
-    private static let rnSDKConfig: [String: AnyObject] = loadPlist(named: "klaviyo-sdk-configuration") ?? [:]
+    private static let rnSDKConfig: [String: AnyObject] = {
+        // Only check for RN config if React Native is present
+        guard NSClassFromString("RCTBridge") != nil else {
+            return [:]
```
amber-klaviyo: is return [:] the standard response when something isn't found?

amber-klaviyo (approved):
Left some questions to improve my understanding of swift code. Non blocking!

Sources/KlaviyoCore/KlaviyoEnvironment.swift
ab1470: because it's a static property with a closure initializer, it's already lazy

ab1470 (approved):
nice fix!

Sources/KlaviyoCore/Utils/FileUtils.swift
ajaysubra: Good question, `guard` is used to guard code below it from executing if some condition isn't met. For instance, if we were trying to add two optional integers and one of them was nil you wouldn't want the code that does the addition to be executed. You would Essentially use it in cases where you want early return. On the other hand if let is used if a certain condition is met execute a certain set of code. 

In my experience, I've see these used interchangeably and boils down to personal preference.

Sources/KlaviyoCore/KlaviyoEnvironment.swift
ajaysubra: Hmm. I hadn't thought about it but you are right generally, you'd want to return a nil but in this case the way the caller is setup, if it can't find the key `klaviyo_sdk_name` in the dictionary it will just use the hardcoded native swift name.

```
if let sdkName = KlaviyoEnvironment.rnSDKConfig["klaviyo_sdk_name"] as? String {
            return sdkName
        }
        return __klaviyoSwiftName
```

PR #449, opened 2025-11-10
Minor Geofence refactor work

ajaysubra:
# Description

## Due Diligence

### DIFF
--- a/Examples/KlaviyoSwiftExamples/SPMExample/SPMExample/Info.plist
+++ b/Examples/KlaviyoSwiftExamples/SPMExample/SPMExample/Info.plist
@@ -40,11 +40,7 @@
 	<array>
 		<string>remote-notification</string>
 		<string>location</string>
-<<<<<<< HEAD
         <string>fetch</string>
-=======
-    <string>fetch</string>
->>>>>>> feat/geofencing
 	</array>
 	<key>NSLocationWhenInUseUsageDescription</key>
 	<string>This app needs location access to show your current location on the delivery map and help you find nearby delivery zones.</string>


Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -12,132 +12,101 @@ import KlaviyoCore
 import KlaviyoSwift
 import OSLog
 
-class KlaviyoLocationManager: NSObject {
+class KlaviyoLocationManager: NSObject, CLLocationManagerDelegate {
```
ajaysubra: I moved the `CLLocationManagerDelegate` here since I moved the extension to a different file.

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -12,132 +12,101 @@ import KlaviyoCore
...
         self.locationManager = locationManager ?? CLLocationManager()
-        self.geofenceManager = geofenceManager ?? KlaviyoGeofenceManager(locationManager: self.locationManager)
 
         super.init()
         self.locationManager.delegate = self
         self.locationManager.allowsBackgroundLocationUpdates = true
         self.locationManager.startMonitoringSignificantLocationChanges()
     }
 
+    // Q: why do we need this?
```
ajaysubra: testing and good clean up.

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
```diff
@@ -12,132 +12,101 @@ import KlaviyoCore
...
-    }
-}
-
-extension KlaviyoLocationManager: CLLocationManagerDelegate {
-    // MARK: Authorization
-
-    public func locationManager(_ manager: CLLocationManager, didFailWithError error: Error) {
-        if #available(iOS 14.0, *) {
-            Logger.geoservices.error("Core Location services error: \(error.localizedDescription)")
+    func startGeofenceMonitoring() {
```
ajaysubra: `setupGeofencing` -> `startGeofenceMonitoring` 

suggesting this name change to be specific about what the method does.

Sources/KlaviyoLocation/KlaviyoLocationManager+CLLocationManagerDelegate.swift
```diff
@@ -0,0 +1,108 @@
...
+    public func locationManagerDidChangeAuthorization(_ manager: CLLocationManager) {
+        handleCLAuthorizationStatusChange(manager, locationManager.currentAuthorizationStatus)
+    }
+
+    @available(iOS, deprecated: 14.0)
+    public func locationManager(_ manager: CLLocationManager, didChangeAuthorization status: CLAuthorizationStatus) {
+        handleCLAuthorizationStatusChange(manager, status)
+    }
+
+    private func handleCLAuthorizationStatusChange(_ manager: CLLocationManager, _ status: CLAuthorizationStatus) {
```
ajaysubra: @ajaysubra to push the changes he made here.

Sources/KlaviyoLocation/KlaviyoSDK+Location.swift
```diff
@@ -16,11 +16,15 @@ extension KlaviyoSDK {
...
         }
     }
 
+    // Q: have some questions about this?
+    // what code in the manager init does this actually need?
+    // if we know what can we just call that instead?
+    // I still don't think I fully understand why this is needed?
     /// To be called in didFinishLaunchingWithOptions to ensure geofence events that happen in a backgrounded/terminated state are processed.
     public func monitorGeofencesFromBackground() {
         _ = KlaviyoLocationManager.shared
```
ajaysubra: @belleklaviyo to dig into this a bit and also may be a create a internal method in `KlaviyoLocationManager` that is specific to `monitorGeofencesFromBackground` and this method will also be called from `KlaviyoLocationManager.init()`.

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
belleklaviyo: this was declared twice, moved it to the extension instead of here since it's actually implemented there

Sources/KlaviyoLocation/Models/Geofence.swift
```diff
@@ -55,6 +55,7 @@ struct Geofence: Equatable, Hashable, Codable {
         self.radius = radius
     }
 
+    // Q - why do we need this?
```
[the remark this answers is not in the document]
belleklaviyo: just felt good to have imo, don't _really_ need but also helps code wise enforce our expected id format to ensure we don't have parsing errors later on when extracting the company id, geofence id, or (coming soon on here too) dwell time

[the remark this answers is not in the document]
ajaysubra: Gotcha. Thanks! We generally avoid validation in the SDK that assumes something like this to be true just to keep SDK logic opinionated. This is a case where I would highly suspect we would change the size of the company id ever but if it's not really needed I would not enforce since we don't do at the analytics layer either.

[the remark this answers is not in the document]
belleklaviyo: ah got it ok will remove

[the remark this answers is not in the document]
ajaysubra: un-opinionated is what I meant 🤦🏾

Sources/KlaviyoLocation/KlaviyoLocationManager+CLLocationManagerDelegate.swift
ajaysubra: @belleklaviyo pushed.

Sources/KlaviyoLocation/KlaviyoLocationManager.swift
ajaysubra: were you able to set the delegate here w/o this conformance?

PR #451, opened 2025-11-12
Add automated CocoaPods publishing workflow

ajaysubra:
## Summary
Automates CocoaPods publishing when a GitHub release is created. Eliminates manual `pod trunk push` commands and ensures all four podspecs are published in correct dependency order.

## What This Does
- ✅ Automatically triggers when you publish a GitHub release
- ✅ Validates all podspecs before publishing
- ✅ Publishes in dependency order: Core → Extension/Swift → Forms
- ✅ Sends Slack notifications on success/failure
- ✅ Supports manual trigger via workflow_dispatch

## Setup Required

**You need to add the CocoaPods token as a GitHub secret:**

1. Get your CocoaPods trunk token:
   ```bash
   cat ~/.netrc | grep -A 2 cocoapods.org
   ```
   Look for the `password` value after `cocoapods.org`

2. Add to GitHub Secrets:
   - Go to: https://github.com/klaviyo/klaviyo-swift-sdk/settings/secrets/actions
   - Click "New repository secret"
   - Name: `COCOAPODS_TRUNK_TOKEN`
   - Value: <paste token from step 1>
   - Click "Add secret"

## Testing

After merging and adding the secret, the workflow will automatically run on the next GitHub release. You can also test it manually:
1. Go to Actions tab
2. Select "Publish to CocoaPods" workflow
3. Click "Run workflow"
4. Enter version number (e.g., 5.1.1)

## Benefits
- No more manual `pod trunk push` for 4 podspecs
- Guaranteed correct publish order
- Immediate Slack notification on success/failure
- Consistent release process
- Can re-run if CocoaPods fails temporarily

### DIFF
--- a/.github/workflows/cocoapods-publish.yml
+++ b/.github/workflows/cocoapods-publish.yml
@@ -27,7 +27,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
-          ref: ${{ github.event.release.tag_name || github.ref }}
+          ref: ${{ github.event.release.tag_name || github.event.inputs.version || github.ref }}
 
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
@@ -40,11 +40,11 @@ jobs:
 
       - name: Validate podspecs
         run: |
-          echo "Validating all podspecs..."
+          echo "Validating all podspecs in dependency order..."
           pod lib lint KlaviyoCore.podspec --allow-warnings
-          pod lib lint KlaviyoForms.podspec --allow-warnings
           pod lib lint KlaviyoSwiftExtension.podspec --allow-warnings
           pod lib lint KlaviyoSwift.podspec --allow-warnings
+          pod lib lint KlaviyoForms.podspec --allow-warnings
           echo "✅ All podspecs are valid"
 
       - name: Publish to CocoaPods Trunk
@@ -84,11 +84,11 @@ jobs:
               - type: "section"
                 text:
                   type: "mrkdwn"
-                  text: "*Version:* `${{ github.event.release.tag_name || github.event.inputs.version }}`\n*Repository:* ${{ github.repository }}\n*Published Podspecs:*\n• KlaviyoCore\n• KlaviyoSwiftExtension\n• KlaviyoSwift\n• KlaviyoForms"
+                  text: "*Version:* `${{ github.event.release.tag_name || github.event.inputs.version }}`\n*Repository:* ${{ github.repository }}\n*Trigger:* ${{ github.event_name == 'release' && 'Automatic (GitHub Release)' || 'Manual (Workflow Dispatch)' }}\n*Published Podspecs:*\n• KlaviyoCore\n• KlaviyoSwiftExtension\n• KlaviyoSwift\n• KlaviyoForms"
               - type: "section"
                 text:
                   type: "mrkdwn"
-                  text: "*Release:* <${{ github.event.release.html_url }}|View Release>\n*Workflow:* <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View Workflow>"
+                  text: "${{ github.event.release.html_url && format('*Release:* <{0}|View Release>\n', github.event.release.html_url) || '' }}*Workflow:* <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|View Workflow>"
 
       - name: Send Slack notification on failure
         if: failure()


ajaysubra:
## Cursor Bot Feedback Addressed ✅

All three issues have been fixed:

### 1. ✅ Version tag now used for manual triggers
**Before:** Ignored `version` input and published from HEAD  
**After:** `ref: ${{ github.event.release.tag_name || github.event.inputs.version || github.ref }}`

Manual runs now correctly checkout the specified version tag.

### 2. ✅ Validation order fixed
**Before:** KlaviyoForms validated before KlaviyoSwift (dependency violation)  
**After:** Core → Extension → Swift → Forms

Now respects dependencies correctly.

### 3. ✅ Slack links fixed for manual triggers
**Before:** `github.event.release.html_url` was undefined for manual runs, creating broken links  
**After:** Conditionally includes release link only when it exists, plus added "Trigger" field

Manual runs now show clean Slack notifications without broken links.

---

**Re: @evan-masseau's question about `workflow_dispatch`**

Manual trigger is useful for:
- Re-publishing if CocoaPods trunk has temporary issues
- Publishing old releases (like 5.1.1) that pre-date this workflow
- Testing the workflow without creating a GitHub release

.github/workflows/cocoapods-publish.yml
```diff
@@ -0,0 +1,113 @@
...
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version to publish (e.g., 5.1.1)'
+        required: true
+        type: string
```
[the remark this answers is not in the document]
evan-masseau: Yeah workflow dispatch is a way to trigger an action from the github UI. It might be convenient to have in this case actually. Other times claude suggested it for me was like a trigger to run automated test suite, which just seemed silly.

> When workflow_dispatch is included in the on section, a "Run workflow" button becomes available in the GitHub Actions UI for that specific workflow. This allows users to manually initiate a workflow run, selecting a branch to run on.

.github/workflows/cocoapods-publish.yml
```diff
@@ -0,0 +1,113 @@
...
+
+      - name: Publish to CocoaPods Trunk
+        env:
+          COCOAPODS_TRUNK_TOKEN: ${{ secrets.COCOAPODS_TRUNK_TOKEN }}
+        run: |
+          echo "Publishing podspecs to CocoaPods trunk..."
+
+          # Publish in dependency order
+          echo "📦 Publishing KlaviyoCore..."
+          pod trunk push KlaviyoCore.podspec --allow-warnings
```
evan-masseau: do these take a while? you could put in a matrix so that they run in parallel

evan-masseau: Maybe we can just `sleep(10minutes)` ... 😬 good thing GH action minutes are free for open source repos

ajaysubra:
```
Good question! Looking at your current fastlane, I notice it only publishes KlaviyoSwift.podspec (line 21, 30). You're already manually publishing the other three, so this is actually an improvement.

  How pod trunk push Works

  pod trunk push is synchronous - it:
  1. Uploads the podspec to CocoaPods
  2. Waits for validation and processing
  3. Updates the CocoaPods specs repo
  4. Only returns after the pod is live

  So propagation happens before the command completes.

  Real-World Evidence

  Many multi-pod projects publish sequentially without delays:
  - Firebase iOS SDK (20+ pods)
  - Stripe iOS SDK (multiple pods)
  - Your own manual workflow (you publish them one after another)

  Options to Address the Concern

  Option 1: Keep as-is (Recommended)

  pod trunk push blocks until propagation completes. Sequential execution should work fine.

  Option 2: Add explicit delays

  echo "⏳ Waiting 30s for CocoaPods CDN propagation..."
  sleep 30

  Option 3: Add validation between pushes

  pod trunk push KlaviyoCore.podspec --allow-warnings
  pod spec cat KlaviyoCore | head -1  # Verify it's live
  pod trunk push KlaviyoSwift.podspec --allow-warnings

  My recommendation: Start with Option 1. If you see propagation issues in practice, add delays. The fact that pod trunk push is synchronous should handle this.
```

ajaysubra: I like claude suggestion here, start off with no delays and add as needed

ab1470: Here's ChatGPT's suggestion:
https://chatgpt.com/share/e/6914fcc8-662c-8005-9886-af393e91a1bf

If we were starting from scratch I'd do option 1, but I don't think that's feasible because we'd break things for our consumers. Option 2 implements the while loop that @evan-masseau [suggested](https://github.com/klaviyo/klaviyo-swift-sdk/pull/451#discussion_r2519870217)

evan-masseau: Option 3 is smart but won't work for all cases, so yeah i'd just go with the polling option. Where was chat gpt in 2023 when we first released our cocoapods!

ajaysubra:
## Re: @ab1470's CocoaPods propagation concern ✅

Great catch\! You identified the exact issue we just hit in CI.

### The Problem During Validation
During the **validation step**, KlaviyoSwift tries to find its dependency `KlaviyoCore (~> 5.1.1)` in the CocoaPods trunk, but it doesn't exist yet (we only validated it, haven't published). This caused the validation to fail with:

```
ERROR | CocoaPods could not find compatible versions for pod "KlaviyoCore":
  KlaviyoSwift depends on KlaviyoCore (~> 5.1.1)
None of your spec sources contain a spec satisfying the dependency
```

### The Fix
Added `--include-podspecs='*.podspec'` for dependent pods:
```bash
pod lib lint KlaviyoSwift.podspec --allow-warnings --include-podspecs='*.podspec'
pod lib lint KlaviyoForms.podspec --allow-warnings --include-podspecs='*.podspec'
```

This tells validation to use **local podspecs** instead of looking them up in the trunk.

### What About Propagation During Publishing?
`pod trunk push` is **synchronous** - it blocks until the pod is fully published and live on CocoaPods CDN. So when we publish in dependency order:
1. `pod trunk push KlaviyoCore.podspec` → waits until live, then returns
2. `pod trunk push KlaviyoSwift.podspec` → can find KlaviyoCore 5.1.1 in trunk
3. `pod trunk push KlaviyoForms.podspec` → can find KlaviyoSwift 5.1.1 in trunk

**No additional delays needed between pushes\!** The command doesn't return until propagation is complete.

Fixed in commit 9ff0174

PR #454, opened 2025-11-14
Implement cooldown period for geofence events

belleklaviyo:
# Description
Heavily referenced https://github.com/klaviyo/klaviyo-android-sdk/pull/355 as organizationally was a fan of making this cooldown tracker a separate entity. 

This introduces a cooldown tracker to prevent duplicate events from being created. In testing, we found that CoreLocation occasionally triggers double events just due to the timing and conditions of if the app was terminated and then opened or just backgrounded, etc. This tracker writes to a map stored in UserDefaults that tracks when the last type of event for a specific geofence occurred. If an event of that type was created in the last 60 seconds, it does not create the duplicate geofence event. 

## Due Diligence

## Changelog / Code Overview
- Created a GeofenceCooldownTracker to prevent duplicate events from being logged

## Test Plan
I added logs to see when didExit/didEnter was triggered and saw when double logs came in, only one event was actually created.

### DIFF
--- a/Sources/KlaviyoLocation/Utilities/GeofenceCooldownTracker.swift
+++ b/Sources/KlaviyoLocation/Utilities/GeofenceCooldownTracker.swift
@@ -59,6 +59,11 @@ class GeofenceCooldownTracker {
         saveCooldownMap(cooldownMap)
     }
 
+    /// Cleans up the cooldown map to expire old keys, to be triggered when startGeofencingMonitoring is called
+    func clean() {
+        saveCooldownMap(loadCooldownMap())
+    }
+
     /// Generate a unique key for a geofence and transition combination
     private func getCooldownMapKey(geofenceId: String, transition: Event.EventName.LocationEvent) -> String {
         "\(geofenceId):\(transition)"
--- a/Sources/KlaviyoLocation/KlaviyoLocationManager.swift
+++ b/Sources/KlaviyoLocation/KlaviyoLocationManager.swift
@@ -46,6 +46,7 @@ class KlaviyoLocationManager: NSObject {
             }
             return
         }
+        cooldownTracker.clean()
 
         Task {
             guard let apiKey = try? await KlaviyoInternal.fetchAPIKey() else {
--- a/Tests/KlaviyoLocationTests/GeofenceCooldownTrackerTests.swift
+++ b/Tests/KlaviyoLocationTests/GeofenceCooldownTrackerTests.swift
@@ -200,7 +200,7 @@ final class GeofenceCooldownTrackerTests: XCTestCase {
 
     // MARK: - Stale Entry Cleanup Tests
 
-    func test_loadCooldownMapCleansUpStaleEntriesAutomatically() {
+    func test_saveCooldownMapFiltersStaleEntries() {
         // GIVEN - Store map with both recent and stale entries
         let geofence1 = "geofence-1"
         let geofence2 = "geofence-2"
@@ -214,14 +214,11 @@ final class GeofenceCooldownTrackerTests: XCTestCase {
 
         setCooldownData(cooldownData)
 
-        // WHEN - Trigger a load by calling isAllowed (which calls loadCooldownMap internally)
+        // WHEN - Record a new transition (this triggers saveCooldownMap which filters stale entries)
         mockDate = Date(timeIntervalSince1970: baseTime)
-        _ = tracker.isAllowed(geofenceId: geofence1, transition: .geofenceEnter)
-
-        // Record a new transition to trigger save
         tracker.recordTransition(geofenceId: geofence1, transition: .geofenceEnter)
 
-        // THEN - Verify stale entry was removed
+        // THEN - Verify stale entry was removed during save
         let storedData = getCooldownData()
         XCTAssertNotNil(storedData, "Cooldown data should exist")
 
@@ -229,7 +226,7 @@ final class GeofenceCooldownTrackerTests: XCTestCase {
         XCTAssertNotNil(storedData?["\(geofence1):geofenceEnter"], "Recent entry for geofence1 should exist")
         XCTAssertNotNil(storedData?["\(geofence3):geofenceExit"], "Recent entry for geofence3 should exist")
 
-        // Stale entry should be gone (filtered out during load)
+        // Stale entry should be gone (filtered out during save)
         XCTAssertNil(storedData?["\(geofence2):geofenceEnter"], "Stale entry for geofence2 should be removed")
     }
 


Sources/KlaviyoLocation/Utilities/GeofenceCooldownTracker.swift
```diff
@@ -0,0 +1,107 @@
...
+/// This tracker enforces a cooldown period per geofence and transition
+/// type to filter out noise while allowing legitimate events. Cooldown data
+/// is stored in UserDefaults with automatic cleanup of stale entries.
+internal class GeofenceCooldownTracker {
+    /// Key for storing geofence transition cooldown timestamps as a dictionary
+    private static let geofenceCooldownsKey = "geofence_cooldowns"
+
+    /// Cooldown period for geofence transitions in seconds (60 seconds)
+    /// Prevents duplicate events emitted from CoreLocation.
+    private static let geofenceTransitionCooldown: TimeInterval = 60.0
```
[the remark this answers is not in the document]
belleklaviyo: yeah so more context in this (lengthy) [thread](https://klaviyo.slack.com/archives/C092G8ASQH1/p1761673093199389) but tldr; we talked about maybe making this a configurable setting for developers, but ultimately decided not to. since this is more of a technical side effect that might be conflated with a developer being like "I don't wanna miss any events, send them all" it feels safer to just do this for them. and ultimately this is a pretty low pass filter

Sources/KlaviyoLocation/Utilities/GeofenceCooldownTracker.swift
```diff
@@ -0,0 +1,107 @@
...
+                let age = currentTime - timestampDouble
+                if age < Self.geofenceTransitionCooldown {
+                    cleanedMap[key] = timestampDouble
+                }
+            }
+        }
+
+        // Save cleaned map if entries were removed
+        if cleanedMap.count != json.count {
+            saveCooldownMap(cleanedMap)
```
ab1470: hmm, why is the `loadCooldownMap` also doing the work of saving a new cooldown map? Why don't we do the work of filtering stale entries within the `saveCooldownMap` method?

Sources/KlaviyoLocation/Utilities/GeofenceCooldownTracker.swift
```diff
@@ -0,0 +1,107 @@
...
+import KlaviyoCore
+import KlaviyoSwift
+
+/// Manages geofence transition cooldown periods to prevent duplicate
+/// events emitted from CoreLocation.
+///
+/// This tracker enforces a cooldown period per geofence and transition
+/// type to filter out noise while allowing legitimate events. Cooldown data
+/// is stored in UserDefaults with automatic cleanup of stale entries.
+internal class GeofenceCooldownTracker {
```
ajaysubra: we don't need this access modifier since internal is default

Sources/KlaviyoLocation/KlaviyoLocationManager+CLLocationManagerDelegate.swift
```diff
@@ -67,11 +67,23 @@ extension KlaviyoLocationManager: CLLocationManagerDelegate {
             return
         }
 
+        // Check cooldown period before processing event
+        guard cooldownTracker.isAllowed(geofenceId: klaviyoLocationId, transition: eventType) else {
+            if #available(iOS 14.0, *) {
+                let action = eventType == .geofenceEnter ? "entered" : "exited"
+                Logger.geoservices.info("🌎 User \(action) region \"\(klaviyoLocationId, privacy: .public)\" (cooldown active, skipping)")
```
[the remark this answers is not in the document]
belleklaviyo: oooh good to know, appreciate the explanation 👍

evan-masseau (commented):
Minor, but I just added on android an additional `clean` method that I call when the dev registers for geofencing. Its quite trivial, but it gives an extra hook (effectively on app launch) to clean up the expired keys, since geofence transition events may be fairly infrequent.

Sources/KlaviyoLocation/Utilities/GeofenceCooldownTracker.swift
evan-masseau: Personal preference I guess, but I actually preferred the pattern of the `load` function auto-cleaning expired keys

Sources/KlaviyoLocation/Utilities/GeofenceCooldownTracker.swift
```diff
@@ -0,0 +1,106 @@
...
+        let mapKey = getCooldownMapKey(geofenceId: geofenceId, transition: transition)
+        let currentTime = environment.date().timeIntervalSince1970
+
+        cooldownMap[mapKey] = currentTime
+        saveCooldownMap(cooldownMap)
+    }
+
+    /// Generate a unique key for a geofence and transition combination
+    private func getCooldownMapKey(geofenceId: String, transition: Event.EventName.LocationEvent) -> String {
+        "\(geofenceId):\(transition)"
```
amber-klaviyo: I'm not the most familiar with geofencing- what is the transition / location event in this scenario?

belleklaviyo: no worries yeah it's either an `exit` or `enter` event and eventually `dwell`, so this key represents the specific exit and enter event for a specific geofence

Sources/KlaviyoLocation/KlaviyoLocationManager+CLLocationManagerDelegate.swift
```diff
@@ -67,11 +67,21 @@ extension KlaviyoLocationManager: CLLocationManagerDelegate {
...
+        }
+
         if #available(iOS 14.0, *) {
-            let action = eventType == .geofenceEnter ? "entered" : "exited"
-            Logger.geoservices.info("🌎 User \(action) region \"\(klaviyoLocationId, privacy: .public)\"")
+            Logger.geoservices.info("🌎 User \(eventType == .geofenceEnter ? "entered" : "exited") region \"\(klaviyoLocationId, privacy: .public)\"")
         }
 
+        // Record the transition to start cooldown period
+        cooldownTracker.recordTransition(geofenceId: klaviyoLocationId, transition: eventType)
```
amber-klaviyo: I see that we are calling isAllowed, and then recordTransition, with the same parameters. Would it make sense to add the isAllowed check to the recordTransition function? That way the call here is simplified and there is no risk of the isAllowed check getting skipped if recordTransition is called elsewhere in the future?

Or is there a use case where we might want to recordTransition without checking for the cooldown?

belleklaviyo: Hmmm fair point! Thought about it a bit - I think there is still some benefit to keeping them separate, so it's clearer from the KlaviyoLocationManager what we are checking and doing as well as the logging being closer to the actual handler method. If we combine them it would be some compound isAllowedAndUpdateMap type method which might be a bit bloated. Would prefer to keep the GeofenceCooldownTracker with shorter/simpler methods

PR #462, opened 2025-11-24
Reduce significant location changes monitoring

belleklaviyo:
# Description
We can't prevent Apple from popping up the alert every 3 days about how often an app has accessed a user's background location, but we can try to make the number associated with it less scary!

I'm pretty sure the number of times an app is monitoring the user's location is related to our usage of `startMonitoringSignificantLocationChanges` -- this is necessary in order to ensure geofence events can trigger in terminated states. However, maybe we don't need to monitor it always and can stop monitoring for significant location changes when we know we already have the user's location and app processes running to support firing the events.

Hard to tell how much this will help cut down on that number (guess we will see in 3 ish days) but this overall seems like good cleanup.

## Due Diligence

## Changelog / Code Overview
Only monitor significant location changes in the terminated state. Reduce unnecessary monitoring.

## Test Plan
Tested this IRL. Loaded geofences near me and with this code change on my physical device and walked around testing different types of terminated states from manually closing from the app switcher to shutting down the phone and restarting it. All events were received as expected.

### DIFF
--- a/Sources/KlaviyoLocation/KlaviyoLocationManager.swift
+++ b/Sources/KlaviyoLocation/KlaviyoLocationManager.swift
@@ -17,6 +17,7 @@ class KlaviyoLocationManager: NSObject {
 
     private var locationManager: LocationManagerProtocol
     private var apiKeyCancellable: AnyCancellable?
+    private var lifecycleCancellable: AnyCancellable?
     internal let cooldownTracker = GeofenceCooldownTracker()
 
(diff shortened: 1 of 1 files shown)


ajaysubra (approved):

belleklaviyo:
@ab1470 yes however I think having tested that scenario exactly (and on a real device) and still seeing the expected behavior work, I don't think it's right

PR #463, opened 2025-11-25
Add geofencing validation

belleklaviyo:
# Description
Adds some extra validation and careful handling to our geofencing module:
- Adds `monitoringDidFailFor` implementation with logs for relevant errors -- there are [quite a number of different errors](https://developer.apple.com/documentation/corelocation/clerror-swift.struct/code), but just tried to pare down to the relevant ones
- Parses the geofences one by one instead of an all-or-nothing type response and with corresponding logging if any individual geofence failed
- Implements the 20 region limit and ensures we respect what regions CLLocationManager may already be monitoring and only ever add Klaviyo geofences as there are available slots ** effectively deprioritizing our geofences in case other CLRegions are already in place **
   - We still attempt to add all Klaviyo geofences (even if it goes beyond the 20 limit) and leave it to the OS to decide what to monitor and what not as there apparently is some internal filtering based on distance (not well documented, based on SO threads)
   - We provide a warning log about this limitation if we hit the limit

## Due Diligence

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-28196

### DIFF
--- a/Sources/KlaviyoLocation/KlaviyoLocationManager+CLLocationManagerDelegate.swift
+++ b/Sources/KlaviyoLocation/KlaviyoLocationManager+CLLocationManagerDelegate.swift
@@ -53,40 +53,6 @@ extension KlaviyoLocationManager: CLLocationManagerDelegate {
 
     // MARK: Geofencing
 
-    public func locationManager(_ manager: CLLocationManager, monitoringDidFailFor region: CLRegion?, withError error: Error) {
-        guard let region = region as? CLCircularRegion,
-              region.isKlaviyoGeofence else {
-            return
-        }
-
-        guard let clError = error as? CLError else {
-            if #available(iOS 14.0, *) {
-                Logger.geoservices.error("❌ Klaviyo geofence monitoring failed: \(error.localizedDescription, privacy: .public)")
-            }
-            return
-        }
-
-        switch clError.code {
-        case .regionMonitoringFailure:
-            if #available(iOS 14.0, *) {
-                let currentCount = manager.monitoredRegions.count
-                Logger.geoservices.error("❌ Klaviyo geofence monitoring failed. Currently monitoring \(currentCount) regions. iOS limit is 20 regions per app.")
-            }
-        case .regionMonitoringDenied, .denied:
-            if #available(iOS 14.0, *) {
-                Logger.geoservices.error("❌ Klaviyo geofence monitoring denied - insufficient permissions")
-            }
-        case .regionMonitoringResponseDelayed, .regionMonitoringSetupDelayed:
-            if #available(iOS 14.0, *) {
-                Logger.geoservices.warning("⚠️ Klaviyo geofence monitoring delayed")
-            }
-        default:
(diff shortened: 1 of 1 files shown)


Sources/KlaviyoLocation/KlaviyoLocationManager+CLLocationManagerDelegate.swift
```diff
@@ -53,6 +53,40 @@ extension KlaviyoLocationManager: CLLocationManagerDelegate {
...
+
+        switch clError.code {
+        case .regionMonitoringFailure:
+            if #available(iOS 14.0, *) {
+                let currentCount = manager.monitoredRegions.count
+                Logger.geoservices.error("❌ Klaviyo geofence monitoring failed. Currently monitoring \(currentCount) regions. iOS limit is 20 regions per app.")
+            }
+        case .regionMonitoringDenied, .denied:
+            if #available(iOS 14.0, *) {
+                Logger.geoservices.error("❌ Klaviyo geofence monitoring denied - insufficient permissions")
```
ab1470:
```suggestion
                Logger.geoservices.warning("⚠️ Klaviyo geofence monitoring denied - insufficient permissions")
```
nit: it's not really an error condition when the user hasn't granted permission. I think warning is more appropriate here.

Sources/KlaviyoLocation/GeofenceService.swift
```diff
@@ -46,14 +46,31 @@ struct GeofenceService: GeofenceServiceProvider {
...
+                failedCount += 1
+                if #available(iOS 14.0, *) {
+                    Logger.geoservices.warning("⚠️ Failed to parse geofence \(rawGeofence.id): \(error.localizedDescription, privacy: .public)")
+                }
+            }
+        }
+
+        if failedCount > 0, #available(iOS 14.0, *) {
+            Logger.geoservices.warning("⚠️ Failed to parse \(failedCount) of \(response.data.count) geofences. Continuing with \(geofences.count) successfully parsed geofences.")
+        }
```
ajaysubra: Since we are already logging when the failure happens, what is this doing in addition?

Sources/KlaviyoLocation/GeofenceService.swift
```diff
@@ -46,14 +46,31 @@ struct GeofenceService: GeofenceServiceProvider {
...
         let response = try JSONDecoder().decode(GeofenceJSONResponse.self, from: data)
-        return try Set(response.data.map { rawGeofence in
-            try Geofence(
-                id: "_k:\(companyId):\(rawGeofence.id)",
-                longitude: rawGeofence.attributes.longitude,
-                latitude: rawGeofence.attributes.latitude,
-                radius: rawGeofence.attributes.radius
-            )
-        })
+        var geofences: Set<Geofence> = []
```
ajaysubra: Also, general question about this logic, when would there be a case where some geofences are pass parsing whereas other wont? 

FWIW, I think what you are doing here is legit but asking to see if you are already seeing something to made you add this code.

ajaysubra (approved):

PR #489, opened 2025-12-19
Chainable extension functions

evan-masseau:
# Description

I think I might have spotted an oversight while preparing release branches

In the example app code, we had 
```
KlaviyoSDK()
    .initialize(with: "ABC123")
    .registerForInAppForms() // STEP2A: register for in app forms
    .registerGeofencing() // STEP2B: register for in geofencing
```

But `registerForInAppForms` does not return anything. So I added `return self` with `discardableResult` like `initialize`.

## Due Diligence

## Related Issues/Tickets

Currently targeted this at the 5.2.0 prep branch but would re-target it to 5.2.0 once that actually has the location module code merged in


ajaysubra (approved):
Oops! Breaking change?

PR #492, opened 2026-01-05
Add CodingKeys to futureproof geofencing json changes

belleklaviyo:
# Description
Should've added these CodingKeys the first time around. But with `duration` in mind for how we may or may not support dwell in the future, we want to ensure a new JSON response doesn't break the existing decoding.

## Due Diligence


belleklaviyo:
@ab1470 yeah ngl I thought CodingKeys is required to ensure if we began receiving a JSON response with a new field like `duration` or anything else that the current GeofenceJSON struct does not have, it would not break the parsing -- but if that's not the case I can just close this PR and keep it as is

PR #508, opened 2026-02-03
Simplify button category registration

belleklaviyo:
# Description
Realized the category registration was making the assumption that each push had a unique id from the `_k` key but that `_k` key is pointing to a dict, not a string. I originally thought we could just hardcode a plain string for all Klaviyo pushes with action buttons. But we do in fact need a unique id _per each push_ in order to set up its action buttons correctly, or else new incoming pushes can overwrite existing pushes' buttons. (ex. if two were in the Notification Center)

This change:
- updates the isKlaviyoNotification helper to work for multiple UNNotifcation types so we can know whether or not to set up push action buttons for an incoming push if its from Klaviyo or not
- changes the category registration so it uses the push identifier or UUID, instead of a mistaken String on `_k` that was the presumed wrong type, now registers category as `com.klaviyo.button.(push identifier)`
- updates the tests to reflect this accurate type of `_k` being a dict
- removes unnecessary cleanup and processing of categories that overcomplicated it
- introduces a queue to make category registration thread safe in case two pushes are attempted to be processed at the same time
- nit: cleans up some file headers

## Due Diligence

## Test Plan
Using the `apns-cli` I sent several payloads that more correctly reflected the expected formats:

```
{
  "aps": {
    "alert": {
      "title": "New Arrivals Just for You",
      "body": "Check out our latest collection"
    },
    "mutable-content": 1,
    "sound": "default"
  },
  "body": {
    "_k": {},
    "action_buttons": [
      {
        "id": "someId",
        "label": "deep link to form!",
        "action": "deep_link",
        "url": "klaviyotest://forms"
      },
      {
        "id": "someOtherId",
        "label": "just open app!",
        "action": "open_app"
      },
      {
        "id": "someOtherOtherId",
        "action": "invalid_action",
        "label": "Invalid"
      }
    ]
  }
}
```
This is identified as a Klaviyo push and renders the two valid buttons (`deep link to form!` and `just open app!`) and ignored the other.

```
{
  "aps": {
    "alert": {
      "title": "New Arrivals Just for You",
      "body": "Check out our latest collection"
    },
    "mutable-content": 1,
    "sound": "default"
  },
  "body": {
    "action_buttons": [
      {
        "id": "someId",
        "label": "deep link to form!",
        "action": "deep_link",
        "url": "klaviyotest://forms"
      },
      {
        "id": "someOtherId",
        "label": "just open app!",
        "action": "open_app"
      },
      {
        "id": "someOtherOtherId",
        "action": "invalid_action",
        "label": "Invalid"
      }
    ]
  }
}
```
This was ignored as it is not a klaviyo push (missing the `_k` identifier), so it was delivered simply with just the title and body and no buttons

I also sent two valid klaviyo pushes at the same time, not opening either and let them be in my Notification Center. I was able to long press hold both of them to reveal the action buttons and see the correct action buttons on each no matter what order I sent/viewed them.

### DIFF
--- a/Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
+++ b/Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
@@ -224,10 +224,15 @@ public enum KlaviyoExtensionSDK {
         // Create actions
         let actions = KlaviyoActionButtonParser.createActions(from: buttonDefs)
 
-        // Register category dynamically
-        KlaviyoCategoryController.shared.registerCategory(actions: actions)
+        // Generate unique category identifier for this notification
+        let baseIdentifier = "com.klaviyo.button."
+        let uniqueSuffix = request.identifier.isEmpty ? UUID().uuidString : request.identifier
+        let categoryIdentifier = "\(baseIdentifier)\(uniqueSuffix)"
+
+        // Register category dynamically with unique identifier
+        KlaviyoCategoryController.shared.registerCategory(categoryIdentifier: categoryIdentifier, actions: actions)
 
         // Set category on notification content
-        bestAttemptContent.categoryIdentifier = KlaviyoCategoryController.categoryIdentifier
+        bestAttemptContent.categoryIdentifier = categoryIdentifier
     }
 }


Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -209,35 +209,25 @@ public enum KlaviyoExtensionSDK {
         request: UNNotificationRequest,
         bestAttemptContent: UNMutableNotificationContent
     ) {
-        // Respect developer-set categories (don't override non-Klaviyo categories)
+        // Respect developer-set categories for non-Klaviyo notifications
+        // Only process and set up action buttons for Klaviyo notifications
         let existingCategory = request.content.categoryIdentifier
-        if !existingCategory.isEmpty && !existingCategory.hasPrefix("com.klaviyo.") {
+        guard bestAttemptContent.userInfo.isKlaviyoNotification() && existingCategory.isEmpty else {
```
belleklaviyo: the previous check `existingCategory.hasPrefix("com.klaviyo.")` doesn't really make sense imo, its safer to just check if it's a Klaviyo notif and if so set it up, if not do not set it up (as it relates to landing in our SDK)

Sources/KlaviyoSwiftExtension/KlaviyoExtension.swift
```diff
@@ -209,35 +209,30 @@ public enum KlaviyoExtensionSDK {
...
         let actions = KlaviyoActionButtonParser.createActions(from: buttonDefs)
 
-        // Register category dynamically
-        let categoryId = KlaviyoCategoryController.shared.registerCategory(
-            notificationId: notificationId,
-            actions: actions
-        )
+        // Generate unique category identifier for this notification
+        let baseIdentifier = "com.klaviyo.button."
+        let uniqueSuffix = request.identifier.isEmpty ? UUID().uuidString : request.identifier
```
belleklaviyo: assumption is each incoming APNS payload has an identifier. In prod (ex. sending from Klaviyo) an identifier does come through as a timestamp of the push. otherwise we could just use a generated UUID(). All that matters for this category identifier is that it is unique per notification

PR #511, opened 2026-02-06
Clean up old UNNotifcationCategories

belleklaviyo:
# Description
Since we are dynamically registering button categories to support custom action buttons, these categories were accumulating “unbounded” for the length of the app’s lifetime until uninstall. Apple says we can register as many as needed per app, but to be a good SDK citizen, we should clean these up as best as we can!

Now, we remove old UNNotificationCategories when we know a notification has been dismissed from one of three methods:
1. tapping the push (via default tap or action button) 
2. clearing it from the Notification Center via swiping it without opening i t
3. clearing all notifications from the app in the Notification Center by swiping away an entire stack of notifications

Methods 1 and 2 are easily addressed as opened push events are directed to our handle method as expected. Experimenting also revealed clearing a push configured with buttons from the Notification Center also triggers our handle method as it receives a dismiss action. 

Method 3 is a little trickier as there is no known notification for a user mass clearing all notifications from the stacked view in the Notification Center.  

This change introduces a new `pruneCategory` method that we call in our handle method for cases 1 or 2. In addition to removing that category we know has either been opened or cleared, we also check to see if there are any other "stale" categories we no longer need (to address case 3). We utilize `getDeliveredNotifications()` to give us the list of all currently displayed notifications from this app in the Notification Center. From this, we can check if we have any registered categories that are no longer displayed. TLDR; 1 and 2 are pretty immediate ways at cleaning up known categories we are done with, 3 is a little tricker. Overall though, this seems just like a nice to have and not super critical.

## Due Diligence

## Test Plan
Here you can see I first send one push w/ action buttons. It registers the category raising the count to 1 but upon opening it removes it, returning the count to 0.

Then I send 3 pushes w/ action buttons. This raises the registered categories count to 3. Upon opening one, it reduces the count to 2. Clearing one from the Notification Center also removes that category, reducing the count to 1.

I then send 4 pushes w/ action buttons. This raises the registered categories count to 5. I clear all of those pushes en masse. I send 1 more push raising the registered categories count to 6. I open that push resulting in that category being pruned, but the other 5 pushes that had been cleared en masse are also pruned. So the count is back to 0.

https://github.com/user-attachments/assets/ad5a3c2b-d21f-475f-981d-cab57b8ab532

## Related Issues/Tickets
https://klaviyo.atlassian.net/browse/CHNL-31191

### DIFF
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -210,10 +210,13 @@ public struct KlaviyoSDK {
             return false
         }
 
-        // Prune the category if the push with action buttons was dismissed from the Notification Center
-        guard notificationResponse.actionIdentifier != UNNotificationDismissActionIdentifier else {
+        defer {
             let categoryIdentifier = notificationResponse.notification.request.content.categoryIdentifier
             KlaviyoCategoryManager.shared.pruneCategory(categoryIdentifier: categoryIdentifier)
+        }
+
+        // Prune the category if the push with action buttons was dismissed from the Notification Center
+        guard notificationResponse.actionIdentifier != UNNotificationDismissActionIdentifier else {
             return true
         }
 
@@ -227,10 +230,6 @@ public struct KlaviyoSDK {
                 dispatchOnMainThread(action: .openDeepLink(url))
             }
         }
-
-        // Prune the category after all event handling and deep link handling is complete
-        let categoryIdentifier = notificationResponse.notification.request.content.categoryIdentifier
-        KlaviyoCategoryManager.shared.pruneCategory(categoryIdentifier: categoryIdentifier)
 
         Task { @MainActor in
             completionHandler()


Sources/KlaviyoCore/KlaviyoCategoryController.swift
```diff
@@ -0,0 +1,210 @@
...
+                    }
+                } else {
+                    if #available(iOS 14.0, *) {
+                        Logger.actionButtons.info("Removed category '\(categoryIdentifier)'. (Total left: \(updatedCategories.count))")
+                    }
+                }
+            } else {
+                // If we can't fetch delivered notifications, still remove the specific category
+                if #available(iOS 14.0, *) {
+                    Logger.actionButtons.info("Removed category '\(categoryIdentifier)'. (Total left: \(updatedCategories.count))")
```
belleklaviyo: if these logs are too in the weeds for the developer, I'm good removing them. Just was helpful while debugging but lmk if anybody feels strongly

ajaysubra (commented):
Few questions else mostly looking good

PR #516, opened 2026-02-22
fix: Handle empty deep link URLs gracefully

ajaysubra:
## Summary

Fixes a bug where in-app forms with empty deep link URLs would fail silently when CTA buttons were clicked.

## Problem

When a form CTA button has no deep link configured in Klaviyo, the JavaScript sends an event with an empty URL:

```json
{
  "type": "openDeepLink",
  "data": {
    "ios": "",
    "android": ""
  }
}
```

The SDK tried to decode this empty string as a `URL` type, which throws an error and causes the entire event to be **silently dropped**. This means:

- ❌ No indication the CTA was clicked
- ❌ No logging
- ❌ Forms with "Dismiss" or "Remind me later" CTAs fail
- ❌ Poor debugging experience

### 1. Made URL Optional in Decoding

Changed `DeepLinkEventPayload.ios` from `URL` to `URL?` with a custom decoder:

```swift
struct DeepLinkEventPayload: Codable {
    let ios: URL?
    
    init(from decoder: Decoder) throws {
        let container = try decoder.container(keyedBy: CodingKeys.self)
        let urlString = try container.decode(String.self, forKey: .ios)
        
        // Handle empty string gracefully
        if urlString.isEmpty {
            self.ios = nil
        } else {
            self.ios = URL(string: urlString)
        }
    }
}
```

### 2. Use Placeholder for Nil URLs

When URL is nil/empty, use `about:blank` as a placeholder to maintain the existing event structure without changing the enum.

### 3. Validate Before Opening

Added validation to skip attempting to open placeholder/invalid URLs:

```swift
guard url.scheme != "about" && !url.absoluteString.isEmpty else {
    Logger.webViewLogger.info("CTA clicked but no deep link URL configured in form")
    return
}
```

### 4. Enhanced Logging

Added debug logging to see exactly what messages are received from JavaScript, making debugging much easier.

## Impact

- ✅ Forms with empty URLs now work correctly
- ✅ CTAs without URLs work (e.g., "Dismiss", "Not Interested", "Remind me later")
- ✅ Clear, actionable log messages
- ✅ No breaking changes - all existing forms continue to work
- ✅ Better debugging experience

## Edge Cases Handled

| Case | Before | After |
|------|--------|-------|
| Empty string (`""`) | ❌ Silent failure | ✅ Logs "no URL configured" |
| Invalid URL | ❌ Silent failure | ✅ Uses placeholder, logs info |
| Valid URL | ✅ Works | ✅ Still works |
| Missing scheme | ⚠️ Fails at `canOpenURL` | ✅ Validates first |

## Files Changed

- `Sources/KlaviyoForms/InAppForms/Models/IAFNativeBridgeEvent.swift` - Optional URL decoding
- `Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift` - URL validation and enhanced logging

## Test Plan

- [x] Test with empty URL (`""`)
- [x] Test with invalid URL (`"not-a-url"`)
- [x] Test with valid URL (`"klaviyo://debug"`)
- [x] Test with missing URL scheme
- [x] Verify all existing tests pass
- [x] Verify logging output is clear
- [x] Test in example app with real forms

## Breaking Changes

None - this is a backward-compatible bug fix.

## Version Recommendation

Patch version bump (e.g., 5.2.0 → 5.2.1) as this is a bug fix with no API changes.

### Before (with empty URL)
```
Received native bridge message: {"type":"openDeepLink","data":{"ios":"",android":""}}
Failed to decode JSON: dataCorrupted(...)
❌ Silent failure
```

### After (with empty URL)
```
Received native bridge message: {"type":"openDeepLink","data":{"ios":"","android":""}}
Received 'openDeepLink' event from KlaviyoJS with url: about:blank
CTA clicked but no deep link URL configured in form
✅ Clear feedback
```

---

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/Models/IAFNativeBridgeEvent.swift
+++ b/Sources/KlaviyoForms/InAppForms/Models/IAFNativeBridgeEvent.swift
@@ -83,7 +83,7 @@ enum IAFNativeBridgeEvent: Decodable, Equatable {
 }
 
 extension IAFNativeBridgeEvent {
-    struct DeepLinkEventPayload: Codable {
+    struct DeepLinkEventPayload: Decodable {
         let ios: URL?
 
         enum CodingKeys: String, CodingKey {
@@ -98,7 +98,7 @@ extension IAFNativeBridgeEvent {
         }
     }
 
-    struct AbortPayload: Codable {
+    struct AbortPayload: Decodable {
         let reason: String
     }
 }


Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -219,12 +219,24 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
...
+            }
             return
         }
 
         switch handler {
         case .klaviyoNativeBridge:
-            guard let jsonString = message.body as? String else { return }
+            guard let jsonString = message.body as? String else {
+                if #available(iOS 14.0, *) {
+                    Logger.webViewLogger.warning("Message body is not a string: \(type(of: message.body))")
```
ab1470:
```suggestion
                    Logger.webViewLogger.warning("Message body is not a string: \(type(of: message.body), privacy: .public)")
```

there isn't any potential for leaking secrets here so we can mark this as public. Without the `public` tag, the `\(type(of: message.body)` value will be redacted in production logs

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -233,6 +245,7 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
             } catch {
                 if #available(iOS 14.0, *) {
                     Logger.webViewLogger.warning("Failed to decode JSON: \(error)")
+                    Logger.webViewLogger.warning("Raw JSON: \(jsonString)")
```
ab1470: This will just print a difficult-to-read block of text. We could add an extension to the Utilities directory:

```swift
extension String {
    /// Returns a pretty-printed version of the JSON string, or the original string if it's not valid JSON
    var prettyPrintedJSON: String {
        guard let data = data(using: .utf8),
              let jsonObject = try? JSONSerialization.jsonObject(with: data, options: []),
              let prettyData = try? JSONSerialization.data(withJSONObject: jsonObject, options: [.prettyPrinted, .sortedKeys]),
              let prettyString = String(data: prettyData, encoding: .utf8) else {
            return self
        }
        return prettyString
    }
}
```
then use the extension in this log message:
```swift
Logger.webViewLogger.warning("Raw JSON:\n\(jsonString.prettyPrintedJSON)")
```

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
ab1470: no, this doesn't print the actual message, only the message's type

Either way though, I think it's very unlikely that this would get called, because we should always be able to decode the message into a string

PR #517, opened 2026-02-25
docs: Add form lifecycle event hooks section to README

ajaysubra:
## Summary

- Adds documentation for the new `registerFormLifecycleHandler` / `unregisterFormLifecycleHandler` public API introduced in the form lifecycle hooks feature

## Notes

This PR is intentionally kept separate from the feature PR (#515) so it can be merged after the feature is released and the version number is known.

### DIFF
--- a/README.md
+++ b/README.md
@@ -746,6 +746,38 @@ KlaviyoSDK().unregisterFromInAppForms()
 
 Note that after unregistering, the next call to `registerForInAppForms()` will be considered a new session by the SDK.
 
+### Monitoring Form Lifecycle Events
+
+> ℹ️ Form lifecycle events are available in SDK version 5.x.0 and higher
+
+You can register a handler to track when forms are shown, dismissed, or when users tap call-to-action buttons. This is useful for sending form engagement data to third-party analytics platforms like Amplitude, Segment, or Mixpanel.
+
+```swift
+import KlaviyoForms
+
+KlaviyoSDK().registerFormLifecycleHandler { event in
+    switch event {
+    case .formShown:
+        // Track when a form is displayed
+        Analytics.track("Klaviyo Form Shown")
+    case .formDismissed:
+        // Track when a form is dismissed (user, timeout, or programmatic)
+        Analytics.track("Klaviyo Form Dismissed")
+    case .formCTAClicked:
+        // Track when a user taps a call-to-action button
+        Analytics.track("Klaviyo Form CTA Clicked")
+    }
+}
+```
+
+The handler is called on the main thread and will receive events for all form interactions. To unregister the handler:
+
+```swift
+KlaviyoSDK().unregisterFormLifecycleHandler()
+```
+
+**Note:** The handler is optional and does not affect normal form functionality. Forms will continue to display and track analytics in Klaviyo regardless of whether a handler is registered.
+
 ## Geofencing
 
 >  Geofencing support is available in SDK version 5.2.0 and higher.
--- a/README.md
+++ b/README.md
@@ -755,17 +755,26 @@ You can register a handler to track when forms are shown, dismissed, or when use
 ```swift
 import KlaviyoForms
 
-KlaviyoSDK().registerFormLifecycleHandler { event in
+KlaviyoSDK().registerFormLifecycleHandler { event, context in
     switch event {
     case .formShown:
         // Track when a form is displayed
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -746,6 +746,47 @@ KlaviyoSDK().unregisterFromInAppForms()
 
 Note that after unregistering, the next call to `registerForInAppForms()` will be considered a new session by the SDK.
 
+### Monitoring Form Lifecycle Events
```
belleklaviyo, line 797: we should also add this to the table of contents at the top

ajaysubra:
@evan-masseau Do we still need to merge this? Not seeing any readme updates on swift?

README.md
evan-masseau: +1

README.md
```diff
@@ -746,6 +746,47 @@ KlaviyoSDK().unregisterFromInAppForms()
...
+
+> ℹ️ Form lifecycle events are available in SDK version 5.x.0 and higher
+
+You can register a handler to track when forms are shown, dismissed, or when users tap call-to-action buttons. This is useful for sending form engagement data to third-party analytics platforms like Amplitude, Segment, or Mixpanel.
+
+```swift
+import KlaviyoForms
+
+KlaviyoSDK().registerFormLifecycleHandler { event, context in
```
evan-masseau: Considering the commits' dates, and how much we changed more recently, I'd assume this code is outdated?

README.md
```diff
@@ -746,6 +747,48 @@ KlaviyoSDK().unregisterFromInAppForms()
...
+        ])
+    case .formCtaClicked(let formId, let formName, let buttonLabel, let deepLinkUrl):
+        // Track when a user taps a call-to-action button
+        Analytics.track("Klaviyo Form CTA Clicked", properties: [
+            "formId": formId ?? "",
+            "formName": formName ?? "",
+            "buttonLabel": buttonLabel ?? ""
+        ])
+    }
+}
```
evan-masseau, line 829: Two issues in this code block:

1. `formId`, `formName`, `buttonLabel` are non-optional `String` (and `deepLinkUrl` is non-optional `URL`) per [`FormLifecycleEvent.swift`](https://github.com/klaviyo/klaviyo-swift-sdk/blob/master/Sources/KlaviyoForms/InAppForms/Models/FormLifecycleEvent.swift). The `??` calls won't compile.
2. The `formDismissed` inline comment claims it fires for "user, timeout, or programmatic" dismissal — but the API doc states only user-initiated dismissals fire (SDK-internal teardowns like session timeout do not). Simplest fix: drop the parenthetical so the comment matches the call site.

```suggestion
KlaviyoSDK().registerFormLifecycleHandler { event in
    switch event {
    case .formShown(let formId, let formName):
        // Track when a form is displayed
        Analytics.track("Klaviyo Form Shown", properties: [
            "formId": formId,
            "formName": formName
        ])
    case .formDismissed(let formId, let formName):
        // Track when a form is dismissed
        Analytics.track("Klaviyo Form Dismissed", properties: [
            "formId": formId,
            "formName": formName
        ])
    case .formCtaClicked(let formId, let formName, let buttonLabel, let deepLinkUrl):
        // Track when a user taps a call-to-action button
        Analytics.track("Klaviyo Form CTA Clicked", properties: [
            "formId": formId,
            "formName": formName,
            "buttonLabel": buttonLabel
        ])
    }
}
```

evan-masseau (commented):
Spotted a few accuracy issues against the actual feature PR (#515) — the example code wouldn't compile against the current public API and one of the inline comments contradicts the API doc. Suggested fixes inline.

References:
- [`FormLifecycleEvent.swift`](https://github.com/klaviyo/klaviyo-swift-sdk/blob/master/Sources/KlaviyoForms/InAppForms/Models/FormLifecycleEvent.swift) — `formId`, `formName`, `buttonLabel` are non-optional `String`; `deepLinkUrl` is non-optional `URL`. The `??` operator on these is a Swift compile error.
- The same file documents that `formDismissed` only fires for **user-initiated** dismissals (tap outside / close button) and explicitly **does not** fire for SDK-internal teardowns like session timeouts/aborts.
- `formCtaClicked` is only emitted when the CTA has a deep link configured — worth calling out in the prose so integrators understand which CTA taps fire this event.

evan-masseau (approved):

PR #543, opened 2026-04-07
fix: skip profile reset when setProfile called with same identifiers

jason-myers-klaviyo:
## Summary

Fixes excessive `POST /client/push-tokens` requests caused by `setProfile()` being called repeatedly with the same identifiers.

**Root cause:** `enqueueProfile` unconditionally called `state.reset(preserveTokenData: false)` regardless of whether the incoming identifiers matched current state. Each reset regenerated the anonymous ID (in `reset()` line 223), which triggered spurious push-token API requests. Under rate limiting, this compounded into a request storm.

Same bug as the Android SDK: klaviyo/klaviyo-android-sdk#435

**Fix:** Compare incoming (non-nil) identifiers against current state before calling `reset()`. Anonymous ID is the lowest-order identifier — there's no reason to regenerate it when higher-order identifiers (email, phone, externalId) haven't changed. `resetProfile()` remains available for explicitly clobbering all state.

Only non-nil identifiers are compared, since nil means "not specified" per `updateStateWithProfile` semantics.

**Slack thread:** https://klaviyo.slack.com/archives/C09DKBLQ3PX/p1775509483748149

## Test plan

- [ ] Verify unit tests pass in CI
- [ ] `setProfile(same identifiers)` repeatedly — confirm no spurious push-token requests
- [ ] `setProfile(different identifiers)` — confirm reset still fires, anonymous ID regenerated
- [ ] `setProfile(same identifiers, different attributes)` — confirm attributes still update
- [ ] `resetProfile()` — confirm still clobbers all state as before

Part of MAGE-473

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -548,10 +548,14 @@ struct KlaviyoReducer: ReducerProtocol {
             // churn, which triggers spurious push-token API requests.
             // Only compare identifiers that are provided (non-nil) in the incoming
             // profile, since nil means "not specified" per updateStateWithProfile.
+            // Normalize with the same trimming used by updateStateWithProfile
+            // so whitespace-padded inputs match their stored counterparts.
+            // Guard on the original value (nil = "not specified", skip comparison)
+            // but compare the trimmed value against state.
             let identifiersChanged =
-                (profile.email != nil && profile.email != state.email)
-                    || (profile.phoneNumber != nil && profile.phoneNumber != state.phoneNumber)
-                    || (profile.externalId != nil && profile.externalId != state.externalId)
+                (profile.email != nil && profile.email?.trimWhiteSpaceOrReturnNilIfEmpty() != state.email)
+                    || (profile.phoneNumber != nil && profile.phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty() != state.phoneNumber)
+                    || (profile.externalId != nil && profile.externalId?.trimWhiteSpaceOrReturnNilIfEmpty() != state.externalId)
             if identifiersChanged {
                 state.reset(preserveTokenData: false)
             }
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -542,7 +542,6 @@ struct KlaviyoReducer: ReducerProtocol {
 
             let pushTokenData = state.pushTokenData
             let currentIds = [state.email, state.phoneNumber, state.externalId]
-            let isIdentified = currentIds.contains { $0 != nil }
             let incomingIds = [profile.email, profile.phoneNumber, profile.externalId].map {
                 // Normalize with the same trimming used by updateStateWithProfile
                 // so whitespace-padded inputs match their stored counterparts.
@@ -555,7 +554,7 @@ struct KlaviyoReducer: ReducerProtocol {
             // Resetting with the same identifiers causes unnecessary anonymous ID
             // churn, which triggers spurious push-token API requests.
             // resetProfile() remains available for explicitly clobbering all state.
-            if isIdentified, currentIds != incomingIds {
+            if state.isIdentified, currentIds != incomingIds {
                 state.reset(preserveTokenData: false)
             }
             state.updateStateWithProfile(profile: profile)
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -550,8 +550,8 @@ struct KlaviyoReducer: ReducerProtocol {
             // profile, since nil means "not specified" per updateStateWithProfile.
             let identifiersChanged =
                 (profile.email != nil && profile.email != state.email)
-                || (profile.phoneNumber != nil && profile.phoneNumber != state.phoneNumber)
-                || (profile.externalId != nil && profile.externalId != state.externalId)
+                    || (profile.phoneNumber != nil && profile.phoneNumber != state.phoneNumber)
+                    || (profile.externalId != nil && profile.externalId != state.externalId)
             if identifiersChanged {
                 state.reset(preserveTokenData: false)
             }
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -541,22 +541,21 @@ struct KlaviyoReducer: ReducerProtocol {
             }
 
             let pushTokenData = state.pushTokenData
+            let currentIds = [state.email, state.phoneNumber, state.externalId]
+            let isIdentified = currentIds.contains { $0 != nil }
+            let incomingIds = [profile.email, profile.phoneNumber, profile.externalId].map {
+                // Normalize with the same trimming used by updateStateWithProfile
+                // so whitespace-padded inputs match their stored counterparts.
+                $0?.trimWhiteSpaceOrReturnNilIfEmpty()
+            }
+
             // Only reset if the incoming profile has different identifiers.
             // Anonymous ID is the lowest-order identifier, so there's no reason
             // to regenerate it when higher-order identifiers haven't changed.
             // Resetting with the same identifiers causes unnecessary anonymous ID
             // churn, which triggers spurious push-token API requests.
-            // Only compare identifiers that are provided (non-nil) in the incoming
-            // profile, since nil means "not specified" per updateStateWithProfile.
-            // Normalize with the same trimming used by updateStateWithProfile
-            // so whitespace-padded inputs match their stored counterparts.
-            // Guard on the original value (nil = "not specified", skip comparison)
-            // but compare the trimmed value against state.
-            let identifiersChanged =
-                (profile.email != nil && profile.email?.trimWhiteSpaceOrReturnNilIfEmpty() != state.email)
-                    || (profile.phoneNumber != nil && profile.phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty() != state.phoneNumber)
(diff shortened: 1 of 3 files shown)


Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -541,7 +541,20 @@ struct KlaviyoReducer: ReducerProtocol {
...
+            // churn, which triggers spurious push-token API requests.
+            // Only compare identifiers that are provided (non-nil) in the incoming
+            // profile, since nil means "not specified" per updateStateWithProfile.
+            let identifiersChanged =
+                (profile.email != nil && profile.email != state.email)
+                || (profile.phoneNumber != nil && profile.phoneNumber != state.phoneNumber)
+                || (profile.externalId != nil && profile.externalId != state.externalId)
+            if identifiersChanged {
+                state.reset(preserveTokenData: false)
+            }
```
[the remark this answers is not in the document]
evan-masseau: This doesn't sound right, TBH. Will certainly confirm with extensive manual testing.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -541,7 +541,20 @@ struct KlaviyoReducer: ReducerProtocol {
...
+            // to regenerate it when higher-order identifiers haven't changed.
+            // Resetting with the same identifiers causes unnecessary anonymous ID
+            // churn, which triggers spurious push-token API requests.
+            // Only compare identifiers that are provided (non-nil) in the incoming
+            // profile, since nil means "not specified" per updateStateWithProfile.
+            let identifiersChanged =
+                (profile.email != nil && profile.email != state.email)
+                    || (profile.phoneNumber != nil && profile.phoneNumber != state.phoneNumber)
+                    || (profile.externalId != nil && profile.externalId != state.externalId)
+            if identifiersChanged {
```
evan-masseau: Will want to have unit test coverage of this (We'll pick that up, Jason. Just taking notes)

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -541,7 +541,20 @@ struct KlaviyoReducer: ReducerProtocol {
...
 
             let pushTokenData = state.pushTokenData
-            state.reset(preserveTokenData: false)
+            // Only reset if the incoming profile has different identifiers.
+            // Anonymous ID is the lowest-order identifier, so there's no reason
+            // to regenerate it when higher-order identifiers haven't changed.
+            // Resetting with the same identifiers causes unnecessary anonymous ID
+            // churn, which triggers spurious push-token API requests.
+            // Only compare identifiers that are provided (non-nil) in the incoming
+            // profile, since nil means "not specified" per updateStateWithProfile.
```
njparadis: (question): do we actually want to reset when these higher-level identifiers change ? not sure I have enough context on the high level flow here to say one way or the other 😕

njparadis (commented):
reviews for the review god, comments for the comment throne

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
evan-masseau: We've always treated `setProfile` like a "replacement" call, and thats how its documented. The intelligence Jason is adding here is to prevent an unecessary reset if we can identify that the profile's identifiers haven't actually changed, that feels like a safe fix. Removing the reset entirely would mean something like changing email from 
`evan.masseau@klaviyo.com` to `nate.paradis@klaviyo.com` would carry over any profile attributes from me to you, when we used to clear that slate completely, dangerous change.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -541,7 +541,24 @@ struct KlaviyoReducer: ReducerProtocol {
...
+            // so whitespace-padded inputs match their stored counterparts.
+            // Guard on the original value (nil = "not specified", skip comparison)
+            // but compare the trimmed value against state.
+            let identifiersChanged =
+                (profile.email != nil && profile.email?.trimWhiteSpaceOrReturnNilIfEmpty() != state.email)
+                    || (profile.phoneNumber != nil && profile.phoneNumber?.trimWhiteSpaceOrReturnNilIfEmpty() != state.phoneNumber)
+                    || (profile.externalId != nil && profile.externalId?.trimWhiteSpaceOrReturnNilIfEmpty() != state.externalId)
+            if identifiersChanged {
+                state.reset(preserveTokenData: false)
+            }
```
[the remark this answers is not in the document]
jason-myers-klaviyo: Claude says:

pendingProfile is a [ProfileKey: AnyEncodable]? dictionary that accumulates setProfileProperty calls (line 616-621). It gets cleared in two places: reset() (line 226) and
  updateRequestAndStateWithPendingProfile (line 211).

  The bot's claim: when reset() is skipped, pendingProfile isn't cleared, so on the next flushQueue (line 313-314), it triggers enqueueProfileOrTokenRequest() which sends a stale request.

  But this is wrong. Look at line 128-137: enqueueProfileRequest calls updateRequestAndStateWithPendingProfile which merges pendingProfile into the outgoing request and clears it (line 211). The enqueueProfile case
   (our changed code at line 536+) builds and enqueues a request with the current profile data at lines 553-575. On the next flushQueue, if pendingProfile is non-nil, enqueueProfileOrTokenRequest merges those
  pending properties into a new request — which is correct behavior, not a bug.

  The concern is invalid. The pendingProfile was never data belonging to the "old" profile that needs flushing before the new one — it's attribute data (firstName, custom keys, etc.) that should merge into
  whichever profile request goes out next. When identifiers haven't changed, there's no "old vs new" distinction. The merge at flush time is exactly right.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -541,7 +541,23 @@ struct KlaviyoReducer: ReducerProtocol {
...
+
+            // Only reset if the incoming profile has different identifiers.
+            // Anonymous ID is the lowest-order identifier, so there's no reason
+            // to regenerate it when higher-order identifiers haven't changed.
+            // Resetting with the same identifiers causes unnecessary anonymous ID
+            // churn, which triggers spurious push-token API requests.
+            // resetProfile() remains available for explicitly clobbering all state.
+            if isIdentified, currentIds != incomingIds {
+                state.reset(preserveTokenData: false)
```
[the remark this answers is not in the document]
evan-masseau: This is intentional — we considered and rejected the nil-skip approach.

`setProfile` is a clobbering operation, not a merge. The old code unconditionally called `reset()` on every `setProfile` call, so `setProfile(Profile(email: "a@b.com"))` when state has `phone: "+1234"` would have always reset and cleared phone. Our fix preserves that behavior — the only case where we skip reset is when ALL identifiers match exactly.

The alternative (nil = "not specified", skip comparison) silently changes `setProfile` from clobbering to merging for unspecified fields, which is a bigger semantic departure than intended for this hotfix. `resetProfile()` exists for explicit clobbering; individual setters (`setEmail`, etc.) exist for partial updates. `setProfile` with a `Profile` object should continue to represent a complete profile replacement.

Both Android and iOS SDKs are aligned on this behavior.

evan-masseau (approved):
Cleaned up the conditionals a bit, added unit tests. Prepped a patch release branch off of the 5.2.2 tag since master has some dirty commits that are not release-ready, and rebased/retargeted this PR to that rel branch.

Approving, but should get a second teammate's approval since at this point i'm basically approving myself.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -541,7 +541,23 @@ struct KlaviyoReducer: ReducerProtocol {
             }
 
             let pushTokenData = state.pushTokenData
-            state.reset(preserveTokenData: false)
+            let currentIds = [state.email, state.phoneNumber, state.externalId]
+            let isIdentified = currentIds.contains { $0 != nil }
```
ajaysubra: Super minor nit pick - this logic already exists in `KlaviyoState`. Just to avoid duplication can we use that?

```swift
 case let .enqueueProfile(profile):
    guard case .initialized = state.initalizationState
    else {
        state.pendingRequests.append(.profile(profile))
        return .none
    }

    let pushTokenData = state.pushTokenData
    let currentIds = [state.email, state.phoneNumber, state.externalId]
    let incomingIds = [profile.email, profile.phoneNumber, profile.externalId].map {
        // Normalize with the same trimming used by updateStateWithProfile
        // so whitespace-padded inputs match their stored counterparts.
        $0?.trimWhiteSpaceOrReturnNilIfEmpty()
    }

    // Only reset if the incoming profile has different identifiers.
    // Anonymous ID is the lowest-order identifier, so there's no reason
    // to regenerate it when higher-order identifiers haven't changed.
    // Resetting with the same identifiers causes unnecessary anonymous ID
    // churn, which triggers spurious push-token API requests.
    // resetProfile() remains available for explicitly clobbering all state.
    if state.isIdentified, currentIds != incomingIds {
        state.reset(preserveTokenData: false)
    }
```

ajaysubra (approved):
LGTM with a nit pick.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
evan-masseau: Was the `reset` already duplicated before?

evan-masseau: oh oh the `isIdentified` function gotcha

ajaysubra: Not sure I understand 🤔

evan-masseau: No worries i just didn't follow you at first, I thought you were saying move this whole chunk to `KlaviyoState` or something. But I think I follow now, you just meant `isIdentified` exists there already. Cleaned it up, check now.

PR #546, opened 2026-04-10
fix: skip redundant profile API call when setProfile has no changes

ajaysubra:
When `setProfile` is called with the same identifiers and no non-identifier attributes, the SDK was still enqueuing a profile API request. This PR eliminates that unnecessary network call, completing the de-duplication work started in #543 (which stopped anonymous ID churn but left the profile request itself unconditional).

## Due Diligence

## Changelog / Code Overview

**`Profile.swift`** — adds `hasNonIdentifierData: Bool` computed property. Returns `true` if the profile carries any data beyond email/phoneNumber/externalId (i.e. firstName, lastName, org, title, image, location, or custom properties). Used as the de-dupe signal.

**`StateManagement.swift`** — in the `enqueueProfile` handler, after `updateStateWithProfile`, an early return now skips building and enqueuing any request when:
1. identifiers didn't change, AND
2. the profile has no non-identifier attributes, AND
3. no pending profile properties are queued via `setProfileProperty`

This mirrors the existing `shouldSendTokenUpdate` pattern used by `setPushToken`.

**`StateManagementEdgeCaseTests.swift`** — updates `testSetProfileSameIdentifiersDoesNotReset` to reflect the new (correct) behavior: same identifiers + no attributes = no state change, no API call.

**Behavioral matrix:**

| `setProfile` call | identifiers changed? | non-identifier data? | API call? |
|---|---|---|---|
| Same email/phone/externalId, no attributes | No | No | Skipped ✅ |
| Same identifiers, but new `firstName` | No | Yes | Sent |
| Different identifiers | Yes | — | Sent (reset + API call) |
| Fresh anonymous user with new identifiers | Yes (nil→value) | — | Sent |

## Test Plan
1. Call `setProfile` repeatedly with the same `email` — confirm no repeated network requests in Charles/proxyman.
2. Call `setProfile` with the same identifiers but a new `firstName` — confirm a single profile request is sent.
3. Call `setProfile` with a different `email` — confirm reset + profile request fires normally.
4. Run `make test-library` — all 220 tests pass.

## Related Issues/Tickets
- Follows up on #543 (fix: skip profile reset when setProfile called with same identifiers)
- Linear ticket: [MAGE-482](https://linear.app/klaviyo/issue/MAGE-482/skip-redundant-profile-api-call-when-identifiers-unchanged)

### DIFF
--- a/Sources/KlaviyoSwift/Models/Profile.swift
+++ b/Sources/KlaviyoSwift/Models/Profile.swift
@@ -84,6 +84,19 @@ public struct Profile: Equatable, Codable {
 
     let _properties: AnyCodable
 
+    /// Returns `true` if the profile contains any data beyond the three identifier fields
+    /// (email, phoneNumber, externalId). Used to decide whether to skip a redundant API
+    /// call when the caller's identifiers haven't changed.
+    var hasNonIdentifierData: Bool {
+        firstName != nil ||
+            lastName != nil ||
+            organization != nil ||
+            title != nil ||
+            image != nil ||
+            location != nil ||
+            !properties.isEmpty
+    }
+
     /// Create or update properties about a profile without tracking an associated event.
     /// - Parameters:
     ///   - email: Individual's email address
--- a/Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
+++ b/Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
@@ -584,25 +584,10 @@ class StateManagementEdgeCaseTests: XCTestCase {
 
         let store = TestStore(initialState: initialState, reducer: KlaviyoReducer())
 
-        // Same identifiers → no reset → pushTokenData preserved on state, anonymousId unchanged.
-        // The reducer still builds a registerPushToken request using the captured pushTokenData,
-        // but the state's pushTokenData is NOT cleared (only reset() clears it).
-        _ = await store.send(.enqueueProfile(Profile(email: "same@email.com", phoneNumber: "+15555555555", externalId: "ext-123"))) {
-            // anonymousId stays the same — no reset, pushTokenData stays on state
-            let request = KlaviyoRequest(
-                endpoint: .registerPushToken(
-                    TEST_API_KEY,
-                    PushTokenPayload(
-                        pushToken: initialState.pushTokenData!.pushToken,
-                        enablement: initialState.pushTokenData!.pushEnablement.rawValue,
-                        background: initialState.pushTokenData!.pushBackground.rawValue,
-                        profile: Profile(email: "same@email.com", phoneNumber: "+15555555555", externalId: "ext-123")
-                            .toAPIModel(anonymousId: initialState.anonymousId!)
-                    )
-                )
-            )
-            $0.queue = [request]
-        }
+        // Same identifiers + no non-identifier attributes → no reset, no API call, no state change.
+        // Nothing changed, so there's no reason to hit the network.
+        // The pushTokenData and anonymousId both remain untouched on state.
+        _ = await store.send(.enqueueProfile(Profile(email: "same@email.com", phoneNumber: "+15555555555", externalId: "ext-123")))
     }
 
     @MainActor
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -548,16 +548,26 @@ struct KlaviyoReducer: ReducerProtocol {
                 $0?.trimWhiteSpaceOrReturnNilIfEmpty()
             }
 
+            let identifiersChanged = currentIds != incomingIds
+
             // Only reset if the incoming profile has different identifiers.
             // Anonymous ID is the lowest-order identifier, so there's no reason
             // to regenerate it when higher-order identifiers haven't changed.
             // Resetting with the same identifiers causes unnecessary anonymous ID
             // churn, which triggers spurious push-token API requests.
             // resetProfile() remains available for explicitly clobbering all state.
-            if state.isIdentified, currentIds != incomingIds {
+            if state.isIdentified, identifiersChanged {
                 state.reset(preserveTokenData: false)
             }
             state.updateStateWithProfile(profile: profile)
+
+            // Skip the API call entirely when there is nothing new to sync:
+            // identifiers are unchanged, the profile carries no extra attributes,
+            // and no profile properties are queued up via setProfileProperty.
+            if !identifiersChanged, !profile.hasNonIdentifierData, state.pendingProfile == nil {
+                return .none
+            }
+
             guard let anonymousId = state.anonymousId,
                   let apiKey = state.apiKey
             else {


amber-klaviyo (approved):

jasper-hsu-kyvo (approved):

Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
```diff
@@ -584,25 +584,10 @@ class StateManagementEdgeCaseTests: XCTestCase {
...
-                            .toAPIModel(anonymousId: initialState.anonymousId!)
-                    )
-                )
-            )
-            $0.queue = [request]
-        }
+        // Same identifiers + no non-identifier attributes → no reset, no API call, no state change.
+        // Nothing changed, so there's no reason to hit the network.
+        // The pushTokenData and anonymousId both remain untouched on state.
+        _ = await store.send(.enqueueProfile(Profile(email: "same@email.com", phoneNumber: "+15555555555", externalId: "ext-123")))
```
ndurell, line 590: This test update makes sense!

Sources/KlaviyoSwift/Models/Profile.swift
```diff
@@ -84,6 +84,19 @@ public struct Profile: Equatable, Codable {
 
     let _properties: AnyCodable
 
+    /// Returns `true` if the profile contains any data beyond the three identifier fields
+    /// (email, phoneNumber, externalId). Used to decide whether to skip a redundant API
+    /// call when the caller's identifiers haven't changed.
+    var hasNonIdentifierData: Bool {
```
ndurell, line 90: doubt it'll happen but it feels a little brittle if we ever add more fields.

ndurell: Couple ideas from claude: 
```  
1. A unit test on hasNonIdentifierData that constructs a Profile with each non-identifier field individually and asserts true. When a new field is added
  without updating the check, no test fails — but it at least makes the coupling visible and gives a place to add coverage.
  2. Invert the check — instead of listing what IS non-identifier data, check that the profile equals an identifiers-only version of itself:

  var hasNonIdentifierData: Bool {
      self != Profile(email: email, phoneNumber: phoneNumber, externalId: externalId)
  }```

**Comment by ndurell** [comment]
⚠️ **Fragility note for future maintainers**

Both the identifier comparison (`currentIds` / `incomingIds` arrays in `StateManagement.swift`) and the `hasNonIdentifierData` check in `Profile.swift` enumerate fields explicitly. If a new identifier or profile attribute is ever added to `Profile`, both of these must be updated — otherwise `setProfile` may silently skip API calls or fail to reset state.

Consider adding a comment above each check pointing at the other, e.g.:

```swift
// ⚠️ If you add a new identifier field to Profile, update BOTH this array
// AND the hasNonIdentifierData check in Profile.swift — otherwise setProfile
// may silently skip API calls or fail to reset state.
```

Alternatively, the checks could be inverted to fail-safe (compare against an identifiers-only `Profile` rather than listing fields), but that may be overkill given how stable the identifier set is.

Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
ab1470, line 590: Does this modified unit test cover all of the paths that are affected by this PR?

ab1470 (approved):

ndurell (approved):

ab1470:
> 2. Call setProfile with the same identifiers but a new firstName — confirm a single profile request is sent.

Hmm, I just tested this and I'm not sure it's working as expected. Here's what I did:
1. Set a new email and clicked "Create Profile" -> Network request goes through ✅ (ID 1196 in Proxyman)
2. Added a first name and clicked "Create Profile" -> Network request goes through ✅ (ID 1240 in Proxyman)
3. Without changing anything, I clicked "Create Profile" again -> Network request goes through ❌  (ID 1252 in Proxyman)

I think (1) and (2) are the desired behavior, but (3) is not because nothing has changed between (2) and (3).

Screenshot 2026-04-13 at 15 06 10

Screenshot 2026-04-13 at 15 10 12

ab1470:
**EDIT:** maybe the behavior below is expected? I see this message in the Xcode console after (3):

> email is either empty or same as what is already set earlier.
The SDK will ignore this change, please use resetProfile for
resetting profile identifiers

--- 
I'm seeing another possible issue. If I:

1. Set a new email and click "Create Profile" -> Network request goes through ✅ (ID 1501)
2. Add a first name and click "Create Profile" -> Network request goes through ✅ (ID 1503)
3. Delete the first name and click "Create Profile" -> Nothing happens ❌ 

For (3), I think we want another network request to remove the first name, right?

Screenshot 2026-04-13 at 15 19 49

Sources/KlaviyoSwift/Models/Profile.swift
ajaysubra: Good call. I picked approach 2.

Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
ajaysubra: Good call. Adding some more tests.

ajaysubra:
Thanks @ab1470 for testing. Checking both the scenarios now.

ajaysubra:
> > 2. Call setProfile with the same identifiers but a new firstName — confirm a single profile request is sent.
> 
> Hmm, I just tested this and I'm not sure it's working as expected. Here's what I did:
> 
> 1. Set a new email and clicked "Create Profile" -> Network request goes through ✅ (ID 1196 in Proxyman)
> 2. Added a first name and clicked "Create Profile" -> Network request goes through ✅ (ID 1240 in Proxyman)
> 3. Without changing anything, I clicked "Create Profile" again -> Network request goes through ❌  (ID 1252 in Proxyman)
> 
> I think (1) and (2) are the desired behavior, but (3) is not because nothing has changed between (2) and (3).
> 
> Screenshot 2026-04-13 at 15 06 10 Screenshot 2026-04-13 at 15 10 12

This is expected behavior since first name is a profile property and not an identifier. We only dedup based on identifiers (most common use case) at this point. Eventually we want to add dedup for any profile attribute but that would be a bigger change since we don't persist anything other than identifiers in state.

ajaysubra:
> **EDIT:** maybe the behavior below is expected? I see this message in the Xcode console after (3):
> 
> > email is either empty or same as what is already set earlier.
> > The SDK will ignore this change, please use resetProfile for
> > resetting profile identifiers
> 
> I'm seeing another possible issue. If I:
> 
> 1. Set a new email and click "Create Profile" -> Network request goes through ✅ (ID 1501)
> 2. Add a first name and click "Create Profile" -> Network request goes through ✅ (ID 1503)
> 3. Delete the first name and click "Create Profile" -> Nothing happens ❌
> 
> For (3), I think we want another network request to remove the first name, right?
> 
> Screenshot 2026-04-13 at 15 19 49

Yeah this may not be obvious immediately but I think we added this a while ago and this probably happens on the sdk without these changes.

ajaysubra:
Addressed Noah's fragility concern in 2732b15 — replaced the explicit field list in `hasNonIdentifierData` with an inverted Equatable check. Auto-synthesized Equatable catches any future field additions automatically. Also added unit tests.

PR #564, opened 2026-04-28
fix(extension): drop KlaviyoCore dependency by copying shared types

belleklaviyo:
# Description

`KlaviyoSwiftExtension` imports `KlaviyoCore`, but `KlaviyoCore` contains `UIApplication.shared` calls which are unavailable in extension contexts. This can cause issues for customers whose NSE targets enforce `APPLICATION_EXTENSION_API_ONLY`.

This PR removes the `KlaviyoCore` dependency from `KlaviyoSwiftExtension` entirely by copying the three shared types (`ActionType`, `isKlaviyoNotification`, `KlaviyoCategoryManager`) directly into the extension module. Since `KlaviyoSwiftExtension` (NSE) and `KlaviyoSwift` (main app) run in separate processes, they don't share type identity — each can have its own copy.

## Due Diligence

## Changelog / Code Overview

- **`KlaviyoSwiftExtension`** — new `ActionType.swift`, `UNNotificationContent+Klaviyo.swift`, `KlaviyoCategoryManager.swift` (internal copies); `import KlaviyoCore` removed from all source files
- **`KlaviyoCore`** — `KlaviyoCategoryManager.swift` deleted (moved to extension module); `ActionType` kept for `KlaviyoSwift`'s use
- **`KlaviyoSwift`** — `pruneCategory` removed from `KlaviyoSwiftEnvironment` and both `handle(notificationResponse:)` overloads (pruning was the only reason `KlaviyoSwift` referenced `KlaviyoCategoryManager`)
- **`Package.swift`** — `KlaviyoCore` removed from `KlaviyoSwiftExtension` dependencies
- **`KlaviyoSwiftExtension.podspec`** — `KlaviyoCore` dependency removed

**Trade-offs vs. annotation approach (see sibling PR #562):**
- `KlaviyoSwiftExtension` no longer imports `KlaviyoCore` at all — clean separation
- `ActionType` and `isKlaviyoNotification` are duplicated (~42 lines total, stable)
- Pruning removed as a side effect — stale categories accumulate until the app is reinstalled (not a correctness issue)

## Test Plan

- [ ] Build `KlaviyoSwiftExtension` targeting an NSE with `APPLICATION_EXTENSION_API_ONLY = YES` — should compile without errors
- [ ] Run existing unit tests for `KlaviyoSwiftExtension`
- [ ] Verify rich push notifications still render correctly on device/simulator
- [ ] Verify badge count updates still work

## Related Issues/Tickets

Related to PR #562 (annotation approach)


belleklaviyo:
Was having troubles testing this with the CocoapodsExample in this repo due to some unrelated error (seems to've existed even when I went back to a stable version I definitely ran CocoapodsExample on before -- will look into as a follow up) but I tested this by pulling in these remote pods into the KlaviyoReactNativeSdkExample app and was able to get a successful build and confirming it's pulling in this version that removes the KlaviyoCore dependency and duplicates the types

image

Sources/KlaviyoSwiftExtension/ActionType.swift
```diff
@@ -0,0 +1,11 @@
+//
+//  ActionType.swift
+//
+
+import Foundation
+
+/// Represents the supported action types for push notification buttons.
+enum ActionType: String, Equatable {
```
ajaysubra, line 14: can you add a comment that this type is also duplicated in KlaviyoCore by design?

Sources/KlaviyoCore/KlaviyoCategoryManager.swift
```diff
@@ -79,65 +76,6 @@ public class KlaviyoCategoryManager {
         }
     }
 
-    /// Removes a notification category from the registered categories and prunes any other stale categories.
```
ajaysubra, line 82: uhhh this must hurt! I know you spent a lot of time on this!

ajaysubra (approved):
LGTM thanks for looking into this so quickly!

Sources/KlaviyoCore/KlaviyoCategoryManager.swift
belleklaviyo: its ok maybe we'll restore it one day!

evan-masseau: Is dropping pruning necessary, or just doing it for simplification? Whilst reviewing with storytime, asked Claude and it seemed to think it should still work to keep pruning around in the Core/Swift modules

```
  1. KlaviyoCategoryManager is fully extension-safe. Every UIApplication.shared call in KlaviyoCore is in    
  KlaviyoEnvironment.swift and Utils/DeepLinkHandler.swift. None are in KlaviyoCategoryManager.swift.
  pruneCategory only uses UNUserNotificationCenter.current(), which is allowed in extensions.                
  2. No test coupling forced the removal. There were no dedicated KlaviyoCategoryManagerTests — only mocks in
   KlaviyoSDKTests/TestData (which got cleaned up to remove the pruneCategory env field). Nothing structural 
  held pruning hostage.
  3. KlaviyoSwift still happily imports KlaviyoCore. Only the extension couldn't. The simplest               
  preserve-pruning option is the cheapest: leave KlaviyoCategoryManager (with prune) in KlaviyoCore for      
  KlaviyoSwift's use, and just copy a register-only stub into KlaviyoSwiftExtension. No third copy needed.
```

belleklaviyo: yay yeah I originally removed it just for simplification so it's not duplicating over an entire manager. but yeah in an effort to then save the pruning work I did avoid some duplication: there are now 2 KlaviyoCategoryManagers but the one in the KlaviyoSwiftExtension only handles registering categories as that's all it's responsible for. and then the manager in KlaviyoCore has the pruning method -- tested this on a sim on the RN example app and saw it thru that it still registers and prunes as expected!

belleklaviyo:
Tested push action buttons on this as well and confirmed behavior for registering and pruning works as expected. And opened push events behave as expected with the button metadata.

ab1470 (approved):
Looks good to me! I pulled this into Flutter and the Flutter example app compiles + runs ✅

PR #561, opened 2026-04-24
fix(ci): add --include-podspecs for KlaviyoSwiftExtension CocoaPods lint

belleklaviyo:
# Description
Fix CocoaPods publish workflow failing to validate `KlaviyoSwiftExtension` because it can't resolve its `KlaviyoCore` dependency from the CDN during pre-publish validation.

## Due Diligence
- [x] I am confident these changes are compatible with all iOS and XCode versions the SDK currently supports.

## Release/Versioning Considerations
- [x] `Patch` Contains internal changes or backwards-compatible bug fixes.

## Changelog / Code Overview
`KlaviyoSwiftExtension` now depends on `KlaviyoCore`, but the `cocoapods-publish.yml` workflow was still linting it without `--include-podspecs='*.podspec'`. This caused validation to fail because `KlaviyoCore` hasn't been published to the CDN yet at validation time.

**Changes:**
- Added `--include-podspecs='*.podspec'` to the `KlaviyoSwiftExtension` lint command so it resolves `KlaviyoCore` from local podspec files
- Updated the comment from "standalone, no dependencies" to "depends on KlaviyoCore"

## Test Plan
- Re-run the `cocoapods-publish` workflow via `workflow_dispatch` with version `5.3.0` after merging
- The `publish_pod` function will skip already-published pods (KlaviyoCore) and proceed with the rest

## Related Issues/Tickets
5.3.0 CocoaPods publish failure: `KlaviyoSwiftExtension` validation failed with "CocoaPods could not find compatible versions for pod KlaviyoCore"

PR #569, opened 2026-04-30
fix(ci): revert --include-podspecs for KlaviyoSwiftExtension lint (#561)

belleklaviyo:
## Summary

Reverts the CI change from #561 now that it's no longer needed.

- #564 dropped `KlaviyoCore` as a dependency of `KlaviyoSwiftExtension` by copying shared types directly, so the `--include-podspecs='*.podspec'` flag added in #561 is now a no-op
- Restores the standalone lint command for `KlaviyoSwiftExtension`
- Updates the workflow comment to reflect the current state (`standalone, no dependencies`)

## Test plan

- [ ] Verify the CocoaPods publish workflow lint step passes in CI

### DIFF
--- a/.github/workflows/cocoapods-publish.yml
+++ b/.github/workflows/cocoapods-publish.yml
@@ -2,7 +2,7 @@
 #
 # This workflow publishes all five podspecs in dependency order:
 # 1. KlaviyoCore (foundation, no dependencies)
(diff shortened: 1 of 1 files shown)


belleklaviyo:
tbh workflow probably would be fine without this as the flag would just be ignored but for the sake of cleaning up, I wanted to get this in

ndurell (approved):

PR #575, opened 2026-05-14
feat(KlaviyoCore): JWT parser and validator for auth token system

ab1470:
# Description
Foundational layer for the upcoming `AuthTokenManager` (personalized in-app forms work). Adds a pure-Swift JWT parser that extracts `exp` and `iat` (RFC 7519 NumericDate) claims, rejects malformed/expired tokens, and returns a `ValidatedToken` the manager can use for refresh scheduling. No signature verification — the Klaviyo backend is the security boundary.

## Due Diligence

Note: nothing here is part of the public API surface yet (`ValidatedToken`, `JWTParser`, and `JWTValidationFailure` are all `internal` to `KlaviyoCore`). The downstream `AuthTokenManager` ticket will introduce the public `setAuthTokenProvider(_:)` entry point.

## Changelog / Code Overview

**New files in `Sources/KlaviyoCore/Auth/`:**
- `JWTParser.swift` — pure-Swift parser. Three-segment split, Base64URL decode (with `-`/`_` substitution and re-padding), JSON-decodes `exp`/`iat` via a private `JWTClaims` struct, applies the `now >= exp - leeway` check.
- `ValidatedToken.swift` — value type holding `rawToken` + decoded `expiresAt`/`issuedAt` as `Date` values.
- `JWTValidationFailure.swift` — error enum: `malformedStructure`, `malformedBase64`, `malformedJSON`, `missingExpClaim`, `missingIatClaim`, `expiredOnReceipt`.

**Modified:**
- `Sources/KlaviyoCore/Utils/Logger+Ext.swift` — adds `Logger.auth` category.

**Key design choices (per the auth token spec):**
- **15-second clock-skew leeway**: a token is considered expired when `now >= exp - leeway`, so the SDK never injects a token the backend would reject due to client clock drift.
- **No signature verification**: SDK is a transport for the JWT; the backend performs full validation.
- **No `iat` range-checking here**: presence is required (the downstream `AuthTokenManager` needs it for refresh math: `iat + 0.9 * (exp - iat)`), but iat/exp sanity-checks belong with the manager.
- **OSLog inline at every failure site**: literal messages, no token-adjacent interpolation. If future code logs token-adjacent values, it must use `privacy: .private`. Token contents are never logged.

## Test Plan

`Tests/KlaviyoCoreTests/Auth/JWTParserTests.swift` — 19 Swift Testing functions covering 22 cases:

| Group | Cases |
|---|---|
| Happy path | valid token, ignores unknown claims (`sub`/`aud`/`iss`/custom), fractional NumericDate values |
| Malformed structure (parameterized) | `""`, `"only-one-segment"`, `"header.payload"`, `"a.b.c.d"` |
| Malformed Base64URL | invalid characters in payload |
| Malformed JSON | non-JSON bytes, JSON array, `exp` as string |
| Missing claims | missing `exp`, missing `iat`, empty payload |
| Expiration boundaries | `now == exp`, `now > exp`, `now == exp - leeway` (rejected), `now == exp - leeway - epsilon` (accepted), `leeway = 0` boundary |
| Defaults | `JWTParser.defaultLeeway == 15` |
| Base64URL handling | payload with `-`/`_` characters, no padding |

Run locally:
```bash
xcodebuild test -scheme klaviyo-swift-sdk-Package -destination "platform=iOS Simulator,name=iPhone 17 Pro" -only-testing:KlaviyoCoreTests/JWTParserTests
```

All 19 tests passing. Full `KlaviyoCoreTests` suite (106 tests) green. `swiftlint --strict` and `swiftformat --lint` clean.

## Related Issues/Tickets
- Related to [MAGE-613](https://linear.app/klaviyo/issue/MAGE-613)
- Spec: [Personalized In-App Forms — Auth Token Integration](https://linear.app/klaviyo/document/personalized-in-app-forms-auth-token-integration-03d3495f6093)
- Blocks: MAGE-615 (AuthTokenManager actor and public `setAuthTokenProvider` API)

### DIFF
--- a/Sources/KlaviyoCore/Auth/JWTParser.swift
+++ b/Sources/KlaviyoCore/Auth/JWTParser.swift
@@ -1,0 +1,124 @@
+//
+//  JWTParser.swift
+//  KlaviyoCore
+//
+//  Created by Andrew Balmer on 2026-05-14.
+//
+
+import Foundation
+import OSLog
+
+/// Pure-Swift parser for the foundational subset of RFC 7519 that the SDK depends on.
+///
+/// The parser splits a compact JWT into its three segments, Base64URL-decodes the payload,
+/// JSON-decodes the `exp` and `iat` (NumericDate) claims, and applies a clock-skew leeway to
+/// the expiration check. All other claims are ignored.
+///
+/// The SDK never verifies the token's cryptographic signature; the Klaviyo backend is the
+/// security boundary for that. This type exists purely so the auth-token system can decide
+/// whether a returned token is usable and when it should be refreshed.
+enum JWTParser {
+    /// Clock-skew leeway subtracted from `exp` before comparing to the current time.
+    ///
+    /// A token is treated as expired when `now >= exp - leeway`. The 15-second window keeps
+    /// the SDK from injecting a token that the Klaviyo backend would reject due to clock
+    /// drift between the device and the backend.
+    static let defaultLeeway: TimeInterval = 15
+
+    // Cyclomatic complexity is intrinsic here: every validation step inlines its own
+    // OSLog call behind an iOS 14 availability guard, which doubles the apparent branch
+    // count. Routing through a logging wrapper would lose the per-failure call site.
+    // swiftlint:disable cyclomatic_complexity
+
+    /// Parses a JWT string and validates the `exp` and `iat` claims.
+    ///
+    /// - Parameters:
+    ///   - token: The raw JWT string.
+    ///   - currentTime: The current time used for the expiration check. Injectable to make
+    ///     tests deterministic. Defaults to `Date()`.
+    ///   - leeway: Clock-skew leeway subtracted from `exp`. Defaults to ``defaultLeeway``.
+    /// - Returns: A ``ValidatedToken`` on success, or a ``JWTValidationFailure`` describing
+    ///   why the token was rejected.
+    static func parseAndValidate(
+        _ token: String,
+        currentTime: Date = Date(),
+        leeway: TimeInterval = JWTParser.defaultLeeway
+    ) -> Result<ValidatedToken, JWTValidationFailure> {
+        let segments = token.split(separator: ".", omittingEmptySubsequences: false)
+        guard segments.count == 3 else {
+            if #available(iOS 14.0, *) {
+                Logger.auth.warning("JWT validation failed: malformed structure")
+            }
+            return .failure(.malformedStructure)
+        }
+
+        guard let payloadData = base64URLDecode(String(segments[1])) else {
+            if #available(iOS 14.0, *) {
+                Logger.auth.warning("JWT validation failed: malformed base64URL payload")
+            }
+            return .failure(.malformedBase64)
+        }
+
+        let claims: JWTClaims
+        do {
+            claims = try JSONDecoder().decode(JWTClaims.self, from: payloadData)
+        } catch {
+            if #available(iOS 14.0, *) {
+                Logger.auth.warning("JWT validation failed: malformed JSON payload")
+            }
+            return .failure(.malformedJSON)
+        }
+
+        guard let expiresAtSeconds = claims.expiresAtSeconds else {
+            if #available(iOS 14.0, *) {
+                Logger.auth.warning("JWT validation failed: missing exp claim")
+            }
+            return .failure(.missingExpClaim)
+        }
+        guard let issuedAtSeconds = claims.issuedAtSeconds else {
+            if #available(iOS 14.0, *) {
+                Logger.auth.warning("JWT validation failed: missing iat claim")
+            }
+            return .failure(.missingIatClaim)
+        }
+
+        if currentTime.timeIntervalSince1970 >= expiresAtSeconds - leeway {
+            if #available(iOS 14.0, *) {
+                Logger.auth.warning("JWT validation failed: expired on receipt")
+            }
+            return .failure(.expiredOnReceipt)
+        }
+
+        return .success(ValidatedToken(
+            rawToken: token,
+            expiresAt: Date(timeIntervalSince1970: expiresAtSeconds),
+            issuedAt: Date(timeIntervalSince1970: issuedAtSeconds)
+        ))
+    }
+
+    // swiftlint:enable cyclomatic_complexity
+
+    /// Decodes a Base64URL string (RFC 7515 §2): `+` is replaced by `-`, `/` by `_`, and
+    /// trailing `=` padding is omitted. We reverse those substitutions and re-pad before
+    /// handing off to `Data(base64Encoded:)`.
+    private static func base64URLDecode(_ value: String) -> Data? {
+        var normalized = value
+            .replacingOccurrences(of: "-", with: "+")
+            .replacingOccurrences(of: "_", with: "/")
+        let paddingNeeded = (4 - normalized.count % 4) % 4
+        normalized.append(String(repeating: "=", count: paddingNeeded))
+        return Data(base64Encoded: normalized)
+    }
+}
+
+/// Subset of the JWT payload claims that the SDK reads. Other claims are intentionally not
+/// decoded — `JSONDecoder` ignores unknown keys by default.
+private struct JWTClaims: Decodable {
+    let expiresAtSeconds: TimeInterval?
+    let issuedAtSeconds: TimeInterval?
+
+    private enum CodingKeys: String, CodingKey {
+        case expiresAtSeconds = "exp"
+        case issuedAtSeconds = "iat"
+    }
+}
--- a/Sources/KlaviyoCore/Auth/JWTParser.swift
+++ b/Sources/KlaviyoCore/Auth/JWTParser.swift
@@ -17,6 +17,11 @@ import OSLog
 /// The SDK never verifies the token's cryptographic signature; the Klaviyo backend is the
 /// security boundary for that. This type exists purely so the auth-token system can decide
 /// whether a returned token is usable and when it should be refreshed.
+///
+/// Note: `iat` is required to be present (for the downstream `AuthTokenManager`'s refresh
+/// scheduling, which uses `iat + 0.9 * (exp - iat)`), but is otherwise not range-checked
+/// here. Sanity-checks on the iat/exp relationship belong with the manager that consumes
+/// the validated token.
 enum JWTParser {
     /// Clock-skew leeway subtracted from `exp` before comparing to the current time.
     ///
@@ -24,6 +29,10 @@ enum JWTParser {
     /// the SDK from injecting a token that the Klaviyo backend would reject due to clock
     /// drift between the device and the backend.
     static let defaultLeeway: TimeInterval = 15
+
+    /// Shared decoder for JWT payloads. Held statically because `JSONDecoder` is reentrant
+    /// for `decode(_:from:)` and this path is hit on every token acquisition / refresh.
+    private static let decoder = JSONDecoder()
 
     // Cyclomatic complexity is intrinsic here: every validation step inlines its own
     // OSLog call behind an iOS 14 availability guard, which doubles the apparent branch
@@ -61,7 +70,7 @@ enum JWTParser {
 
         let claims: JWTClaims
         do {
-            claims = try JSONDecoder().decode(JWTClaims.self, from: payloadData)
+            claims = try Self.decoder.decode(JWTClaims.self, from: payloadData)
         } catch {
             if #available(iOS 14.0, *) {
                 Logger.auth.warning("JWT validation failed: malformed JSON payload")
--- a/Sources/KlaviyoCore/Auth/ValidatedToken.swift
+++ b/Sources/KlaviyoCore/Auth/ValidatedToken.swift
@@ -1,0 +1,24 @@
+//
+//  ValidatedToken.swift
+//  KlaviyoCore
+//
+//  Created by Andrew Balmer on 2026-05-14.
+//
+
+import Foundation
+
+/// A JWT that has passed SDK-side parsing and validation.
+///
+/// The SDK reads only the `exp` and `iat` claims; all other claims (`sub`, audience, issuer,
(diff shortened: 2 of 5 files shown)


ab1470:
bugbot run

Sources/KlaviyoCore/Auth/JWTParser.swift
```diff
@@ -0,0 +1,133 @@
...
+/// scheduling, which uses `iat + 0.9 * (exp - iat)`), but is otherwise not range-checked
+/// here. Sanity-checks on the iat/exp relationship belong with the manager that consumes
+/// the validated token.
+enum JWTParser {
+    /// Clock-skew leeway subtracted from `exp` before comparing to the current time.
+    ///
+    /// A token is treated as expired when `now >= exp - leeway`. The 15-second window keeps
+    /// the SDK from injecting a token that the Klaviyo backend would reject due to clock
+    /// drift between the device and the backend.
+    static let defaultLeeway: TimeInterval = 15
```
ndurell, line 31: cool idea!

Sources/KlaviyoCore/Auth/JWTParser.swift
```diff
@@ -0,0 +1,133 @@
...
+            issuedAt: Date(timeIntervalSince1970: issuedAtSeconds)
+        ))
+    }
+
+    // swiftlint:enable cyclomatic_complexity
+
+    /// Decodes a Base64URL string (RFC 7515 §2): `+` is replaced by `-`, `/` by `_`, and
+    /// trailing `=` padding is omitted. We reverse those substitutions and re-pad before
+    /// handing off to `Data(base64Encoded:)`.
+    private static func base64URLDecode(_ value: String) -> Data? {
```
ndurell, line 113: Just double checking we can't get this out of the swift standard lib? Just worried this is some copy pasta from a broken SO implementation.

Sources/KlaviyoCore/Auth/JWTParser.swift
```diff
@@ -0,0 +1,133 @@
...
+        _ token: String,
+        currentTime: Date = Date(),
+        leeway: TimeInterval = JWTParser.defaultLeeway
+    ) -> Result<ValidatedToken, JWTValidationFailure> {
+        let segments = token.split(separator: ".", omittingEmptySubsequences: false)
+        guard segments.count == 3 else {
+            if #available(iOS 14.0, *) {
+                Logger.auth.warning("JWT validation failed: malformed structure")
+            }
+            return .failure(.malformedStructure)
```
ndurell, line 61: Question: if we get a failure what are we going to do with it? Report it back to the developer?

Tests/KlaviyoCoreTests/Auth/JWTParserTests.swift
```diff
@@ -0,0 +1,339 @@
...
+
+        #expect(validated.rawToken == token)
+        #expect(validated.issuedAt == Date(timeIntervalSince1970: Self.defaultIat))
+        #expect(validated.expiresAt == Date(timeIntervalSince1970: Self.defaultExp))
+    }
+
+    @Test
+    func validTokenIgnoresUnknownClaims() throws {
+        let token = try makeJWT(extraClaims: [
+            "sub": "user-123",
```
ndurell: Is this pretty accurate to the type of fields we'd expect to find in a jwt in the wild?

Sources/KlaviyoCore/Auth/JWTParser.swift
ab1470: credit goes to Claude for this one. But yeah, I agree!

Sources/KlaviyoCore/Auth/JWTParser.swift
ab1470: yeah that's the idea. We'll log a warning, then any forms that get displayed to the user will be the generic (aka non-personalized) form

Sources/KlaviyoCore/Auth/JWTParser.swift
ab1470: **TL;DR: I've validated this and it's good**

There's nothing in the Swift standard library that will help us directly decode a JWT. I found a bunch of solutions via google search, all of which have a very similar approach to what this method is doing. One of the more robust solutions seems to be this [JWTDecode library](https://github.com/auth0/JWTDecode.swift) (707 stars and actively maintained). I asked Claude to use the library as a model with which to validate that we're doing the correct thing here, and it says we're good. Here's the full output of what Claude found:

<details>
<summary>Full comparison vs auth0/JWTDecode.swift</summary>

### `base64URLDecode` vs auth0's `base64UrlDecode`

**Functionally equivalent for every valid Base64URL input.** Differences:

| Aspect | Ours | auth0 |
|---|---|---|
| Char substitution (`-`→`+`, `_`→`/`) | Same | Same |
| Padding math | `(4 - count % 4) % 4` | `ceil(length / 4.0) * 4 - length` |
| Length source | `String.count` (grapheme clusters) | `lengthOfBytes(using: .utf8)` |
| Unknown-character policy | Strict — `Data(base64Encoded:)` returns `nil` | Lenient — `.ignoreUnknownCharacters` |

The padding math produces identical results across all input lengths (verified for `count % 4 ∈ {0, 1, 2, 3}`).

The length-source difference is theoretical: Base64URL is ASCII by definition, so `String.count` == byte count for any valid input. Non-ASCII input fails downstream either way.

The **unknown-character policy is the one real behavioral difference** — auth0 silently strips whitespace and other unknown chars from segments, we reject the token as `malformedBase64`. For an SDK that consumes JWTs from a trusted auth provider, our strictness is correct: stray whitespace or stray punctuation in a token segment is a bug in the producer that should surface as a rejection, not get papered over.

### `parseAndValidate` vs auth0's `DecodedJWT.init` + `decodeJWTPart`

| Aspect | Ours | auth0 |
|---|---|---|
| Part split | `split(omittingEmptySubsequences: false)` | `components(separatedBy:)` |
| Part-count check | `count == 3`, fail otherwise | Same |
| Header decoding | **Skipped** | Decoded as `[String: Any]` |
| Payload decoding | `JSONDecoder` → typed `JWTClaims` struct | `JSONSerialization` → `[String: Any]` |
| Signature handling | Ignored (we never verify) | Stored, never verified |
| `exp` typing | `TimeInterval` (`Double`). String/null/bool → rejected | `Claim.double` coerces String→Double, etc. |
| `iat` requirement | **Required** | Optional |
| Expiration rule | `now >= exp - leeway` | `exp.compare(now) != .orderedDescending` (i.e. `exp <= now`) — no leeway |
| Return shape | `Result<ValidatedToken, JWTValidationFailure>` | Throws `JWTDecodeError` |

Everywhere we differ, we're either **more strict** (typed JSON decode, required `iat`, no character tolerance) or **more focused** (skip header, only read `exp`/`iat`, leeway built in). All deliberate, all match the MAGE-613 spec.

### Edge cases traced through both implementations

Both libraries handle these the same way:

- `""` → 1 part → reject (structure)
- `"a"` / `"a.b"` → 1–2 parts → reject (structure)
- `"a.b.c.d"` → 4 parts → reject (structure)
- `".."` → 3 empty parts → payload `""` → `Data(base64Encoded: "")` returns empty `Data` → JSON decode fails → reject (malformed JSON)
- `"a.b."` → 3 parts with empty signature → accepted if payload decodes. We never verify signatures so this passes us as well as auth0. Fine for our use case.
- Negative `exp` (NumericDate before 1970) → `Date(timeIntervalSince1970:)` produces a pre-1970 date → `now >= exp - leeway` is true → rejected as expired.
- `"exp": null` or `"exp": true` → `JSONDecoder` throws `typeMismatch` → rejected as `malformedJSON`.

**One edge case we cover but don't explicitly test:** `"header..signature"` (3 parts, empty payload segment). Our parser correctly rejects it as `malformedJSON` because `Data(base64Encoded: "")` returns empty `Data` which then fails JSON decoding. Worth pinning with a test since the failure mode crosses two of our error categories.

### Conclusion

No code changes recommended. Our implementation is functionally equivalent to auth0's for valid input, and intentionally stricter for malformed input in ways that match the MAGE-613 spec. The one improvement worth making is adding a test for the `"header..signature"` empty-payload-segment case to lock in that behavior.
</details>

Tests/KlaviyoCoreTests/Auth/JWTParserTests.swift
ab1470: Fair — that was a thin sample. Updated in b5658f3f to mirror what real OIDC/OAuth providers (Auth0, Okta, Cognito) emit: `iss`, `sub` (UUID), `aud` as an array, `nbf`, `jti`, `auth_time`, `azp`, `email`, `email_verified`, `scope`, `address`. Picks one of every JSON value type our parser might encounter in an unknown position — string, number, boolean, array of strings, nested object — so the test reads as a realistic payload and incidentally confirms we don't choke on any of those types appearing in claims we ignore.

Sources/KlaviyoCore/Auth/JWTParser.swift
ab1470: actually I'm rethinking this a bit now. We probably want something a little more robust than a simple log message. Regardless of what we end up doing though, the error handling will be implemented when I address [MAGE-623](https://linear.app/klaviyo/issue/MAGE-623/ios-authtokenmanager-skeleton-public-setauthtokenprovider-api); this PR won't change

Sources/KlaviyoCore/Auth/JWTParser.swift
ndurell: Sorry I meant base64 decoding, jwt makes sense to reimplement I think!

Sources/KlaviyoCore/Auth/ValidatedToken.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+
+import Foundation
+
+/// A JWT that has passed SDK-side parsing and validation.
+///
+/// The SDK reads only the `exp` and `iat` claims; all other claims (`sub`, audience, issuer,
+/// custom claims) remain opaque and are resolved by the Klaviyo backend. The SDK does not
+/// verify the token's signature — the backend is the security boundary.
+struct ValidatedToken: Equatable, Hashable {
```
dan-peluso, line 15: thoughts on including email in this? even if we don't use it now might be nice to have in the model to avoid some changes down the line

ab1470: If/when we need it, we'll need to make other changes anyway, so I'd say we wait until we have a use for it. It's super easy to add

Sources/KlaviyoCore/Auth/JWTParser.swift
```diff
@@ -0,0 +1,133 @@
...
+///
+/// The parser splits a compact JWT into its three segments, Base64URL-decodes the payload,
+/// JSON-decodes the `exp` and `iat` (NumericDate) claims, and applies a clock-skew leeway to
+/// the expiration check. All other claims are ignored.
+///
+/// The SDK never verifies the token's cryptographic signature; the Klaviyo backend is the
+/// security boundary for that. This type exists purely so the auth-token system can decide
+/// whether a returned token is usable and when it should be refreshed.
+///
+/// Note: `iat` is required to be present (for the downstream `AuthTokenManager`'s refresh
```
ajaysubra, line 21: @ab1470 not blocking this pr but we should reconsider refresh scheduling in light of the rate limit on the klaviyo side oauth servers. I tagged you on slack in the thread.

Sources/KlaviyoCore/Auth/ValidatedToken.swift
```diff
@@ -0,0 +1,24 @@
...
+    /// The raw JWT string as supplied to the SDK.
+    let rawToken: String
+
+    /// The `exp` (expiration time) claim decoded from a NumericDate to a `Date`.
+    let expiresAt: Date
+
+    /// The `iat` (issued at) claim decoded from a NumericDate to a `Date`.
+    let issuedAt: Date
```
ajaysubra, line 23: I'm guessing `rawToken` is the required field that is needed and `expiresAt` and `issuedAt` are potentially optional? I'm thinking specifically about personalization but also this could extend to any request that needs auth - the request would only need `rawToken` for the request to succeed. Having said all this, my question is - in `JWTParser` specifically `parseAndValidate` we return a failure and not this token if say `expiresAt` is missing. Should we be more forgiving and still log the error but go ahead and return the `rawToken`?

This is probably a nit pick so I don't think we'll get into a situation where some of these are missing but just thinking how forgiving we should be in the event they are missing.

ajaysubra (approved):
LGTM some minor questions!

Sources/KlaviyoCore/Auth/ValidatedToken.swift
ab1470: though _technically_ optional, it's_ essentially universal practice for OAuth services to encode both `exp` and `iat` claims when minting access tokens, so I'm not really worried about these not being available. See [this comment](https://linear.app/klaviyo/document/personalized-in-app-forms-auth-token-integration-03d3495f6093#comment-edc8a112) in my spec

ab1470: also, it's good security practice to reject any token that doesn't have both `exp` and `iat`, so if we somehow get a token that's missing those two claims, we should reject it anyway

Sources/KlaviyoCore/Auth/JWTParser.swift
ab1470: Perhaps I misinterpreted, but my takeaway from yesterday's meeting was Harsha's not concerned about this, and that we'll be able to scale our rate limits if necessary to accommodate demand

ab1470: > I tagged you on slack in the thread

([linking that Slack conversation](https://klaviyo.slack.com/archives/C09GNTWEJKE/p1778867709214019) for reference)

evan-masseau (approved):
Lookin good !

PR #576, opened 2026-05-20
feat(KlaviyoCore): AuthTokenManager skeleton + registerAuthTokenProvider API

ab1470:
# Description

Adds the foundational `AuthTokenManager` actor in `KlaviyoCore` and the public `registerAuthTokenProvider(_:)` API on `KlaviyoSDK`. Ships only the happy-path acquisition flow; concurrency hardening, proactive refresh, the refresh stream, and profile-reset interaction are deferred to sibling sub-issues so this can land as a small, focused PR.

## Due Diligence

## Changelog / Code Overview

**New public API (`KlaviyoSwift`)**
- `AuthTokenProvider` typealias — `@Sendable () async throws -> String`, re-aliased from `KlaviyoCore` so hosts only need `import KlaviyoSwift`.
- `KlaviyoSDK.registerAuthTokenProvider(_:)` — registers the host-supplied JWT-producing closure. Re-registering replaces the previous provider; there is no public clear path (deferred to `resetProfile()` in MAGE-626).

**New package API (`KlaviyoCore`)**
- `AuthTokenManager` actor — singleton accessed via `.shared`. Owns cached-token and provider state. `registerProvider(_:)` clears the cache, stores the provider, logs the event, and fires a fire-and-forget eager fetch. `currentToken()` returns a cached token when valid; otherwise invokes the provider, validates via `JWTParser`, caches, and returns.
- `AuthTokenError` — `.noProviderRegistered` and `.validationFailed(JWTValidationFailure)`.
- Stored properties for `inFlightFetch`, `refreshTask`, `refreshContinuations` are declared so sibling sub-issues can fill them in without churning the actor's shape.

**Visibility tweak**
- `JWTValidationFailure` widened from `internal` to `package` so it can be the associated value of the `package` `AuthTokenError`.

**Logging**
- OSLog at the `Auth` category for provider registration, token acquisition (exp/iat with `.private` privacy markers), and validation failure with reason. Token contents are never logged.

**Out of scope (sibling sub-issues)**
- Dedup + timeout enforcement: MAGE-624
- Proactive refresh scheduling: MAGE-625
- Refresh notification stream + profile-reset: MAGE-626

## Test Plan

- [ ] `xcodebuild build -scheme KlaviyoSwift -destination "platform=iOS Simulator,name=iPhone 17 Pro"` succeeds.
- [ ] `xcodebuild test -scheme klaviyo-swift-sdk-Package -destination "platform=iOS Simulator,name=iPhone 17 Pro" -only-testing:KlaviyoCoreTests` — `AuthTokenManagerTests` (8 tests) and `JWTParserTests` (20 tests) all pass.
- [ ] `swiftlint --strict` on changed files reports zero violations.

## Related Issues/Tickets

- Linear: [MAGE-623](https://linear.app/klaviyo/issue/MAGE-623)
- Parent: [MAGE-615](https://linear.app/klaviyo/issue/MAGE-615)
- Unblocks: [MAGE-616](https://linear.app/klaviyo/issue/MAGE-616) (iOS WebView injection)

### DIFF
--- a/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
+++ b/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
@@ -31,14 +31,6 @@ package actor AuthTokenManager {
     /// Starts `nil`; set by ``registerProvider(_:)``.
     private var provider: AuthTokenProvider?
 
-    // MARK: - Reserved storage
-
-    // TODO: - [MAGE-624] dedup concurrent currentToken() callers
-    private var inFlightFetch: Task<String, Error>?
-
-    // TODO: - [MAGE-625] proactively refresh the cached token
-    private var refreshTask: Task<Void, Never>?
-
     init() {}
 
     /// Registers a new provider, discards any cached token from a previous
--- a/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
+++ b/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
@@ -1,0 +1,135 @@
+//
+//  AuthTokenManager.swift
+//  KlaviyoCore
+//
+//  Created by Andrew Balmer on 2026-05-14.
+//
+
+import Foundation
+import OSLog
+
+/// Owns the host-supplied ``AuthTokenProvider`` and serves the current auth JWT to
+/// internal SDK consumers (in-app forms today, future feature modules tomorrow).
+///
+/// This skeleton ships only the happy-path acquisition flow: register a provider,
+/// fetch and validate a token, cache it, and hand the cached string to callers
+/// until the next provider change. Concurrent-caller deduplication, timeout
+/// enforcement, proactive refresh scheduling, the refresh notification stream, and
+/// the `reset()` lifecycle hook will land in sibling sub-issues. Their stored
+/// properties are declared here so adding those behaviors doesn't churn the
+/// actor's shape.
+///
+/// The token cache is in-memory only — never persisted to disk or Keychain.
+package actor AuthTokenManager {
+    /// Shared instance used by `KlaviyoSDK` and other SDK modules. Tests should
+    /// construct their own instances to keep test runs independent.
+    package static let shared = AuthTokenManager()
+
+    /// Most recently validated token, if any. Cleared whenever
+    /// ``registerProvider(_:)`` runs.
+    private var cachedToken: ValidatedToken?
+
+    /// Host-supplied closure that returns a fresh JWT on each invocation.
+    /// Starts `nil`; set by ``registerProvider(_:)``. There is no public API
+    /// for clearing — that happens internally via ``reset()`` (deferred to a
+    /// sibling sub-issue) when `KlaviyoSDK.resetProfile()` is called.
+    private var provider: AuthTokenProvider?
+
+    // MARK: - Deferred state
+
+    // The following properties are declared as part of this actor's shape so
+    // that sibling sub-issues can fill in their behaviors without churning the
+    // signature. They are unused in this skeleton.
+
+    /// Reserved for concurrent-caller deduplication. Sub-issue MAGE-624 will
+    /// populate this from ``currentToken()`` so that multiple callers share a
+    /// single provider invocation.
+    private var inFlightFetch: Task<String, Error>?
+
+    /// Reserved for proactive refresh scheduling. Sub-issue MAGE-625 will schedule
+    /// refreshes at `iat + 0.9 * (exp - iat)` bounded by `[now + 5s, exp - 30s]`.
+    private var refreshTask: Task<Void, Never>?
+
+    /// Reserved for the refresh-notification stream. Sub-issue MAGE-626 will fan
+    /// successful refresh tokens out to consumers via
+    /// `refreshes() -> AsyncStream<String>`.
+    private var refreshContinuations: [AsyncStream<String>.Continuation] = []
+
+    init() {}
+
+    /// Registers a new provider, discards any cached token from a previous
+    /// provider, and triggers an eager fetch to warm the cache.
+    ///
+    /// The eager fetch is fire-and-forget so registration call sites stay
+    /// responsive — failures during the warm-up surface only as logs (the same
+    /// logs ``currentToken()`` would emit). Calling this again later replaces
+    /// the previous provider; there is no public API for clearing a registered
+    /// provider, since clearing is part of the profile-reset lifecycle handled
+    /// in a sibling sub-issue.
+    package func registerProvider(_ newProvider: @escaping AuthTokenProvider) async {
+        cachedToken = nil
+        provider = newProvider
+
+        if #available(iOS 14.0, *) {
+            Logger.auth.info("AuthTokenManager: provider registered")
+        }
+        Task {
+            _ = try? await self.currentToken()
+        }
+    }
+
+    /// Returns the current auth token, fetching one via the registered provider
+    /// if the cache is empty or has gone stale.
+    ///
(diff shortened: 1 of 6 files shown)


Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -0,0 +1,124 @@
...
+    /// A single caller invokes the provider directly. If two callers race, both
+    /// will invoke the provider — the result is extra network calls, not data
+    /// corruption.
+    ///
+    /// - Throws: ``AuthTokenError/noProviderRegistered`` when no provider is
+    ///   registered; the provider's own error when the provider throws;
+    ///   ``AuthTokenError/validationFailed(_:)`` when the returned token fails
+    ///   ``JWTParser`` validation.
+    package func currentToken() async throws -> String {
+        while true {
```
ndurell: why do we have a while true here?

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -0,0 +1,124 @@
...
+    package func registerProvider(_ newProvider: @escaping AuthTokenProvider) async {
+        cachedToken = nil
+        provider = newProvider
+        generation &+= 1
+
+        if #available(iOS 14.0, *) {
+            Logger.auth.info("AuthTokenManager: provider registered")
+        }
+        Task {
+            _ = try? await self.currentToken()
```
ndurell, line 51: Should the previous task be cancelled if a new one starts? Would we spawn a bunch orphaned tasks if they called registered too much? I assume the expectation is that they will provide a token after they call this right?

ab1470: deduplication and task management will be handled separately, in [MAGE-624](https://linear.app/klaviyo/issue/MAGE-624/ios-authtokenmanager-deduplication-and-timeouts). This ticket is just the base implementation; other tickets will build on this to address edge cases and error conditions.

See the [In-flight fetch tracking and deduplication](https://linear.app/klaviyo/issue/MAGE-624/ios-authtokenmanager-deduplication-and-timeouts#in-flight-fetch-tracking-and-deduplication-e177d418) section of MAGE-624 for details on how we'll handle multiple requests

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
ab1470: Good catch — pulled it. The `while true` was guarding a real race (resumed continuation from a since-replaced provider clobbering the cache), but `currentToken()` has no production callers in this PR yet, so the race is latent until consumers wire up. Carrying stopgap scaffolding that would get ripped out in the very next ticket wasn't worth the cognitive load.

Force-pushed to drop `87f8f7f5`. Replaced it with a brief `TODO: [MAGE-624]` at the `await provider()` site so the gap is documented rather than silently re-introduced. The regression test + `Latch` helper from the dropped commit are stashed as a [comment on MAGE-624](https://linear.app/klaviyo/issue/MAGE-624/ios-authtokenmanager-deduplication-and-timeouts) for the cancellation-based fix to land alongside.

ab1470: Claude added this to address [this bugbot comment](https://github.com/klaviyo/klaviyo-swift-sdk/pull/576/changes/BASE..87f8f7f5a9f970d6699beb3b687ae81f7c93a346#r3275982741), but I should have been more careful. The issue that bugbot raised will be addressed in MAGE-624, so I should have deferred the fix to that ticket rather than putting in a temporary fix in this PR. I reverted this change and put in a `TODO` 

If you're curious about why Claude added the `while true`, though:

>   **Purpose of while true in currentToken()**
>
 > The loop exists to handle one specific race: provider swap mid-fetch. Without it, the code below would have nowhere clean to "retry from the top."
>
> **The race it guards against**
>
> `AuthTokenManager` is an actor, but the actor's mutex is released across every `await`. In `currentToken()` the critical `await` is `try await provider()` on line 84. While that suspension is in flight:
>
> 1. `registerProvider(_:)` can run, which clears `cachedToken`, swaps `provider`, and bumps `generation`.
> 2. When our `await` resumes, the `rawToken` we hold came from the old provider — the host has explicitly replaced it, so writing that token back into the cache would be incorrect.
>
>  The `capturedGeneration == generation` check (line 91) detects that swap, and `continue` sends control back to the top of the loop to re-evaluate against the new state.
>
>  **Why a loop, specifically**
>
>  On the retry iteration, the very next thing the loop does is re-check the cache (line 75). That matters because a different concurrent currentToken() call may have already fetched a valid token from the new provider while we were suspended — in which case we just return that cached value and skip an extra network call.
>
>  The alternatives are all worse:
>  - **Throw on swap** → leaks an internal race up to callers and forces them to retry.
>  - **Recurse** → same control flow, but stacks frames and is harder to read.
>  - **One-shot retry** (`if … else { fetch again }`) → handles single swap but not back-to-back swaps; the loop is a small generalization that costs nothing.
>  
>  In the steady state the loop always exits on its first iteration (cache hit, successful fetch, or thrown error). It only ever does a second pass when a provider swap happens to land inside that one `await`, which should be very rare.

Tests/KlaviyoCoreTests/Auth/AuthTokenManagerTests.swift
```diff
@@ -0,0 +1,210 @@
...
+                return waiter
+            }
+        }
+
+        func waitFor(atLeast threshold: Int) async throws {
+            if value >= threshold { return }
+            await withCheckedContinuation { continuation in
+                waiters.append((threshold, continuation))
+            }
+        }
```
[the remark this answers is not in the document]
ab1470: False positive. There's no suspension point between the value >= threshold check on L171 and the waiters.append(...) on L173.
  
The await on line 172 marks a potential suspension inside `withCheckedContinuation`, but the body closure passed to it runs synchronously on the actor before any suspension occurs — it's how the continuation gets registered in the first place. Swift's actor mutex isn't released at the textual await; it's released when the task actually suspends, which is after the body returns. So the check and the append are atomic from the actor's perspective: increment() cannot interleave between them.
 
This is the canonical async-gate pattern for actors.

Sources/KlaviyoCore/Auth/AuthTokenProvider.swift
```diff
@@ -0,0 +1,18 @@
...
+///
+/// The SDK invokes the provider whenever it needs a fresh token: at registration
+/// time (eager fetch), when the cached token has expired, and during proactive
+/// refresh. Returning a token from a long-lived async source (e.g. a network call
+/// to the host's auth backend) is expected and supported — the SDK respects task
+/// cancellation, so timeouts propagate cleanly for well-behaved implementations.
+///
+/// Returning a malformed or already-expired token causes the SDK to discard it and
+/// log a validation failure; the SDK does not retry on its own.
+public typealias AuthTokenProvider = @Sendable () async throws -> String
```
glenn-klaviyo, line 18: ❓ should this be marked as `package` instead of `public`?

ab1470: I wish we could, but we can't because this is re-exported from `KlaviyoSwift` [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/576/changes#diff-f67f2d1527889976a5b080113f464f968521ee37019a1b88dfb5519a2c11d6c7R12). If I make this `package`, it can't be re-exported from `KlaviyoSwift`

ndurell (approved):

PR #578, opened 2026-05-21
fix(KlaviyoCore): reject JWT segments with non-base64url characters (MAGE-613 follow-up)

ab1470:
# Description

Follow-up to [MAGE-613](https://linear.app/klaviyo/issue/MAGE-613/ios-jwt-parsing-and-validation). Tightens `JWTParser.parseAndValidate` to reject tokens whose header or signature segments contain characters outside the RFC 7515 §2 base64url alphabet. Previously only the payload segment was charset-checked (implicitly, via `Data(base64Encoded:)`), so a token like `header.<valid-payload>.'); evilJS(); //` would pass validation and the raw string would flow into downstream consumers verbatim.

This was flagged on [#577](https://github.com/klaviyo/klaviyo-swift-sdk/pull/577) by Cursor Bugbot as a WebView script-injection vector. Fixing at the parser layer closes the concern at the source — `ValidatedToken.rawToken` is now guaranteed to consist of three `.`-separated runs of `[A-Za-z0-9_-]`, with no characters that need escaping when interpolated into a JS string literal.

## Due Diligence

- [x] ` + "`Patch`" + ` Contains internal changes or backwards-compatible bug fixes.
- [ ] ` + "`Minor`" + ` Contains changes to the public API.
- [ ] ` + "`Major`" + ` Contains **breaking** changes.

## Changelog / Code Overview

- ` + "`JWTParser`" + `: new ` + "`isBase64URLShape(_:)`" + ` helper and ` + "`base64URLCharacterSet`" + ` constant. After the segment-count check, every segment is verified to contain only ` + "`[A-Za-z0-9_-]`" + ` before the payload decode proceeds. Failure returns the existing ` + "`.malformedBase64`" + ` case (semantics widened to cover all segments — see below).
- Empty segments are intentionally still accepted at this layer. RFC 7515 §3 permits an empty signature for ` + "`alg: none`" + `, and the existing ` + "`emptyPayloadSegmentIsRejected`" + ` test pins the boundary that empty payload surfaces as ` + "`.malformedJSON`" + ` downstream, not ` + "`.malformedBase64`" + `.
- ` + "`JWTValidationFailure.malformedBase64`" + `: docstring updated to reflect that the case now covers any segment (not just the payload) containing non-base64url characters.
- Extended the existing ` + "`swiftlint:disable`" + ` directive on ` + "`parseAndValidate`" + ` to also suppress ` + "`function_body_length`" + ` — same reasoning as the existing ` + "`cyclomatic_complexity`" + ` suppression (each validation step inlines its own ` + "`OSLog`" + ` call behind an iOS 14 guard).

## Test Plan

- ` + "`singleQuoteInSignatureIsRejected`" + ` — direct reproduction of the bugbot's flagged injection vector.
- ` + "`backslashInHeaderIsRejected`" + ` — the other half of the JS-literal escape concern.
- ` + "`nonBase64URLCharacterInSegmentIsRejected`" + ` — parameterized sweep over the most common non-alphabet characters (space, tab, newline, `\"`, `<`, `>`, `/`, `+`, `=`, `,`). `+`, `/`, and `=` are the standard-Base64 characters that the JWT compact serialization substitutes/strips, so their presence is itself a sign of malformed input.
- ` + "`emptySignatureSegmentIsAccepted`" + ` — pins that ` + "`header.payload.`" + ` (empty signature, ` + "`alg: none`" + `) still validates.
- All existing ` + "`JWTParserTests`" + ` continue to pass without modification — including the pre-existing ` + "`malformedBase64PayloadIsRejected`" + ` (now caught by the new charset check rather than the downstream ` + "`Data(base64Encoded:)`" + ` failure) and ` + "`emptyPayloadSegmentIsRejected`" + ` (empty segments still pass the charset check; the failure surfaces as ` + "`.malformedJSON`" + ` downstream).

## Related Issues/Tickets

- [MAGE-613](https://linear.app/klaviyo/issue/MAGE-613) — reopened for this follow-up
- [#577](https://github.com/klaviyo/klaviyo-swift-sdk/pull/577) (MAGE-616) — where the gap was flagged; will resolve the bugbot thread once this lands and #577 rebases onto it

### DIFF
--- a/Sources/KlaviyoCore/Auth/JWTParser.swift
+++ b/Sources/KlaviyoCore/Auth/JWTParser.swift
@@ -34,10 +34,11 @@ enum JWTParser {
     /// for `decode(_:from:)` and this path is hit on every token acquisition / refresh.
     private static let decoder = JSONDecoder()
 
-    // Cyclomatic complexity is intrinsic here: every validation step inlines its own
-    // OSLog call behind an iOS 14 availability guard, which doubles the apparent branch
-    // count. Routing through a logging wrapper would lose the per-failure call site.
-    // swiftlint:disable cyclomatic_complexity
+    // Cyclomatic complexity and function-body length are intrinsic here: every
+    // validation step inlines its own OSLog call behind an iOS 14 availability
+    // guard, which doubles the apparent branch count. Routing through a logging
+    // wrapper would lose the per-failure call site.
+    // swiftlint:disable cyclomatic_complexity function_body_length
 
     /// Parses a JWT string and validates the `exp` and `iat` claims.
     ///
@@ -59,6 +60,13 @@ enum JWTParser {
                 Logger.auth.warning("JWT validation failed: malformed structure")
             }
             return .failure(.malformedStructure)
+        }
+
+        for segment in segments where !isBase64URLShape(segment) {
+            if #available(iOS 14.0, *) {
+                Logger.auth.warning("JWT validation failed: malformed base64URL segment")
+            }
+            return .failure(.malformedBase64)
         }
 
         guard let payloadData = base64URLDecode(String(segments[1])) else {
@@ -105,7 +113,22 @@ enum JWTParser {
         ))
     }
 
-    // swiftlint:enable cyclomatic_complexity
+    // swiftlint:enable cyclomatic_complexity function_body_length
+
+    /// Base64URL alphabet (RFC 7515 §2): `A-Z`, `a-z`, `0-9`, `-`, `_`. Padding `=` is
+    /// stripped from JWT segments and is intentionally not part of this set — a `=` in
+    /// a JWT segment is malformed.
+    private static let base64URLCharacterSet = CharacterSet(
+        charactersIn: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_"
+    )
+
+    /// Returns `true` when every code unit in `segment` is in the Base64URL alphabet.
+    /// Empty segments are accepted — RFC 7515 §3 permits an empty signature segment
+    /// (`alg: none`); structural emptiness of payload/header is caught downstream by
+    /// the JSON decode step.
+    private static func isBase64URLShape(_ segment: Substring) -> Bool {
+        segment.unicodeScalars.allSatisfy { base64URLCharacterSet.contains($0) }
+    }
 
     /// Decodes a Base64URL string (RFC 7515 §2): `+` is replaced by `-`, `/` by `_`, and
     /// trailing `=` padding is omitted. We reverse those substitutions and re-pad before
--- a/Sources/KlaviyoCore/Auth/JWTParser.swift
+++ b/Sources/KlaviyoCore/Auth/JWTParser.swift
@@ -115,19 +115,18 @@ enum JWTParser {
 
     // swiftlint:enable cyclomatic_complexity function_body_length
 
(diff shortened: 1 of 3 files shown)


Sources/KlaviyoCore/Auth/JWTParser.swift
```diff
@@ -61,6 +62,13 @@ enum JWTParser {
             return .failure(.malformedStructure)
         }
 
+        for segment in segments where !isBase64URLShape(segment) {
+            if #available(iOS 14.0, *) {
```
dan-peluso, line 66: not really relevant to this pr but I see this check getting tossed around all the time before logs - I wonder how many lines of code we could save by having that check built into the logger itself

Sources/KlaviyoCore/Auth/JWTParser.swift
```diff
@@ -105,7 +113,22 @@ enum JWTParser {
         ))
     }
 
-    // swiftlint:enable cyclomatic_complexity
+    // swiftlint:enable cyclomatic_complexity function_body_length
```
dan-peluso, line 116: is there something we could fix in the swift lint to avoid this?

Sources/KlaviyoCore/Auth/JWTParser.swift
```diff
@@ -105,7 +113,22 @@ enum JWTParser {
...
+
+    /// Base64URL alphabet (RFC 7515 §2): `A-Z`, `a-z`, `0-9`, `-`, `_`. Padding `=` is
+    /// stripped from JWT segments and is intentionally not part of this set — a `=` in
+    /// a JWT segment is malformed.
+    private static let base64URLCharacterSet = CharacterSet(
```
dan-peluso: is there some sort of constant we could reference instead of having a character set with the alphabet twice plus symbols? Is this how this is normally being done for this type of encoding? I'm surprised there's not some sort of native checker for this lol

dan-peluso (approved):
basically just style things but the character set code is pretty funny

Sources/KlaviyoCore/Auth/JWTParser.swift
ab1470: lol this has been discussed extensively. Unfortunately we can't really remove the `if #available` check. Building something "into the logger itself" doesn't work because it breaks some key functionality of the `OSLog` framework. Not really easy to describe here but if you're ever curious I can show you.

The good news is that whenever we drop support for iOS 13.0 (which I'd be very in favor of, 13.0 is ancient and last i checked it was so few users on 13.0 that it didn't even show up on the stats), we can immediately find/replace the entire codebase to remove all of these checks

Sources/KlaviyoCore/Auth/JWTParser.swift
ab1470: we could loosen the rules, but I don't think we should. It's fairly standard practice to temporarily disable lint rules for specific circumstances, like this one

ab1470: although the cyclomatic complexity one is due to all of the `if #available(iOS 14.0, *) {...}` checks. Maybe I can get the linter to ignore those? I'll create a tech debt ticket for for it

Sources/KlaviyoCore/Auth/JWTParser.swift
ab1470: I just pushed [e1e3021b](https://github.com/klaviyo/klaviyo-swift-sdk/pull/578/changes/e1e3021b2f05abda82ee904713475272ec761703), which cleans it up a little, but there's no super-easy single property we can query for this

PR #586, opened 2026-05-28
feat(KlaviyoCore): AuthTokenManager dedup, timeouts, and cache integrity (MAGE-624)

ab1470:
# Description
Adds concurrent-caller deduplication, per-caller timeout enforcement, and cache integrity across provider swaps to `AuthTokenManager`. Builds on the skeleton from MAGE-623; addresses the race condition explicitly called out in PR #576 review (and the stashed regression test from that thread is adapted here).

## Due Diligence

All new surface is `package` — host code sees no API change. The IAF call site's bare `currentToken()` is unchanged; only its internal TODO comment moved.

## Changelog / Code Overview

**`AuthTokenManager` (`KlaviyoCore`)**
- New `FetchMode` raw-value enum: `.bestEffort = 0.5` (form-display default, matches the spec's "500ms enforced by the manager" at the IAF call site) and `.proactive = 5.0` (background acquisition: registration-time eager fetch, scheduled refresh in MAGE-625).
- Concurrent callers dedup via a single in-flight slot, `private var inFlight: (id: UUID, task: Task<String, Error>)?`. Always set/cleared as a unit so the invariant "ID and task travel together" is structural.
- Per-caller timeout race via a `OnceResolver` actor over `withCheckedThrowingContinuation` + two detached tasks. `withThrowingTaskGroup` is not suitable here — a group implicitly awaits all children before its closure returns, and `Task.value` does not honor the awaiter's cancellation, so a group-based race would block on the unfinished fetch child until the underlying provider returned regardless of which child won. Documented inline.
- `registerProvider(_:)` cancels any in-flight task and clears the slot before installing the new provider. The fetch body adds an explicit `Task.checkCancellation()` checkpoint immediately after `await provider()` and before any cache write, so a non-cancellation-aware host provider closure cannot poison the cache by writing its stale token after the provider has been swapped.
- The generation-UUID guard in the fetch task's `defer` ensures a cancelled task waking up after a newer in-flight slot has been installed by `registerProvider(_:)` cannot clobber the newer slot.

**`AuthTokenError` (`KlaviyoCore`)**
- New `timedOut` case for the per-caller budget elapsing.

**`IAFPresentationManager` (`KlaviyoForms`)**
- TODO comment refreshed to reflect the now-landed behavior. Method call (`AuthTokenManager.shared.currentToken()`) is unchanged: the new default `.bestEffort` is the 500ms form-display budget the spec requires.

**Design decisions worth flagging in review**
- API shape: `currentToken(mode: FetchMode = .bestEffort)` rather than a `timeout: TimeInterval` parameter or a `bestEffort: Bool` overload. The spec treats the two budgets as named *policies* (configurable from telemetry), not as a continuum. Enum cases keep call sites free of magic numbers and let the policies be tuned in one place.
- Hang-forever contract: the SDK does not enforce an upper bound on the underlying fetch's runtime — if a host's provider hangs without honoring cancellation, the slot is stuck until the next `registerProvider(_:)`. The doc comment makes this explicit. Adding a watchdog was considered and rejected: in practice URLSession-based providers cap out at 60s, and the user-visible degradation is already correct (forms display non-personalized content).

## Test Plan

All 14 tests in `AuthTokenManagerTests` pass. Six new tests added:

- [x] `concurrentCallersShareSingleProviderInvocation` — 8 concurrent callers + eager fetch → exactly 1 provider invocation
- [x] `bestEffortTimeoutThrowsWithinBudget` — 500ms budget caller throws `timedOut` against a 2s-sleeping provider
- [x] `proactiveTimeoutThrowsWithinBudget` — 5s budget caller throws `timedOut` against a 10s-sleeping provider
- [x] `bestEffortTimeoutLeavesFetchTaskRunningForLaterCallers` — first caller bails at ~500ms, second caller (proactive) gets the cached result at ~800ms; single provider invocation across both
- [x] `failureClearsInFlightSoNextCallReinvokesProvider` — failure path clears the slot for the next caller
- [x] `providerReplacedWhileFetchInFlightDoesNotPoisonCache` — adapted from the stashed regression test in #576 review; uses a `Latch` helper to drive deterministic interleaving; asserts the second-provider token survives the stale write attempt

Run: `xcodebuild test -scheme klaviyo-swift-sdk-Package -destination "platform=iOS Simulator,name=iPhone 17 Pro" -only-testing:KlaviyoCoreTests/AuthTokenManagerTests`

## Related Issues/Tickets

- [MAGE-624](https://linear.app/klaviyo/issue/MAGE-624) — this PR
- MAGE-615 — parent
- MAGE-623 — skeleton (#576) this builds on
- MAGE-625, MAGE-626, MAGE-683 — sibling sub-issues that consume the failure surface and dedup machinery

## Summary by CodeRabbit

* **New Features**
  * Token acquisition now supports per-call timeout modes and deduplicated concurrent requests for faster, more efficient auth.
  * Eager background cache warm-up after registering a provider.

* **Bug Fixes**
  * Prevented stale in-flight fetches from overwriting newer cached tokens when providers change.
  * Caller timeouts no longer cancel the shared underlying fetch.

* **Documentation**
  * Clarified timeout behavior and default latency budget for interactive requests.

* **Tests**
  * Added tests for deduplication, timeout semantics, cache safety, and provider-swap cases.

### DIFF
--- a/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
+++ b/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
@@ -17,8 +17,29 @@ import OSLog
 /// has elapsed (with the same clock-skew leeway ``JWTParser`` uses) or the
 /// provider is replaced via ``registerProvider(_:)``.
 ///
+/// Concurrent callers requesting a token while a fetch is in flight share the
+/// result of that fetch — they do not trigger additional provider invocations.
+/// Each caller bounds *its own* wait via ``FetchMode``; the underlying fetch
+/// task is unaffected by individual callers timing out and continues until it
+/// completes naturally (or until ``registerProvider(_:)`` cancels it).
+///
 /// The token cache is in-memory only — never persisted to disk or Keychain.
 package actor AuthTokenManager {
+    /// Latency budget for callers awaiting a token, expressed as a named policy
+    /// rather than a free `TimeInterval` so the two budgets stay
+    /// single-source-of-truth and can be tuned together based on production
+    /// telemetry.
+    package enum FetchMode: TimeInterval {
+        /// 500ms budget. Used at form-display time, where a user is waiting on
+        /// the form to appear — missing the personalization window is preferable
+        /// to delaying the form.
+        case bestEffort = 0.5
+
+        /// 5s budget. Used by background acquisition paths (eager fetch at
+        /// registration, scheduled refresh) where no user is actively waiting.
+        case proactive = 5.0
+    }
+
     /// Shared instance used by `KlaviyoSDK` and other SDK modules. Tests should
     /// construct their own instances to keep test runs independent.
     package static let shared = AuthTokenManager()
@@ -31,16 +52,29 @@ package actor AuthTokenManager {
     /// Starts `nil`; set by ``registerProvider(_:)``.
     private var provider: AuthTokenProvider?
 
-    init() {}
+    /// In-flight fetch slot: a token-fetch task paired with the generation `id`
+    /// that names it. Concurrent callers `await` the `task` rather than starting
+    /// their own fetch. The `id` lets the fetch task itself decide whether to
+    /// clear the slot on completion — a stale (cancelled) task waking up after
+    /// a newer fetch was installed must not clobber the newer slot.
+    ///
+    /// Always set and cleared as a unit: there is no valid state where the
+    /// task is present without its id, or vice versa.
+    private var inFlight: (id: UUID, task: Task<String, Error>)?
+
+    package init() {}
 
     /// Registers a new provider, discards any cached token from a previous
-    /// provider, and triggers an eager fetch to warm the cache.
+    /// provider, cancels any in-flight fetch, and triggers an eager fetch to
+    /// warm the cache.
     ///
     /// The eager fetch is fire-and-forget so registration call sites stay
     /// responsive — failures during the warm-up surface only as logs (the same
-    /// logs ``currentToken()`` would emit). Calling this again later replaces
-    /// the previous provider.
+    /// logs ``currentToken(mode:)`` would emit). Calling this again later
+    /// replaces the previous provider.
     package func registerProvider(_ newProvider: @escaping AuthTokenProvider) async {
+        inFlight?.task.cancel()
+        inFlight = nil
         cachedToken = nil
         provider = newProvider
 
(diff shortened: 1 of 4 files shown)


Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -31,73 +52,188 @@ package actor AuthTokenManager {
...
+            Task {
+                do {
+                    let value = try await fetch.value
+                    await resolver.resolve(.success(value))
+                } catch {
+                    await resolver.resolve(.failure(error))
+                }
+            }
+            Task {
+                try? await Task.sleep(nanoseconds: timeoutNanos)
```
glenn-klaviyo, line 236: ❓ should we catch this error and just return in the event this sleep throws due to a `CancellationError` or some other error resulting in an "unexpected" timeout? or is it safe to say that's unlikely as it's part of a dangling `Task { }`?

ab1470: I think it's fine to not catch the error. We don't keep a reference to `Task.sleep`, meaning there's no real path to cancelling it, and I don't think there's a path to it throwing a `CancellationError`.

If for some (very unlikely) reason `Task.sleep` throws an error, it won't cause any real issues in production, it may just result in a slightly misleading error message saying that the auth token timed out. But I'm not worried about it

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -17,8 +17,29 @@ import OSLog
 /// has elapsed (with the same clock-skew leeway ``JWTParser`` uses) or the
 /// provider is replaced via ``registerProvider(_:)``.
 ///
+/// Concurrent callers requesting a token while a fetch is in flight share the
+/// result of that fetch — they do not trigger additional provider invocations.
+/// Each caller bounds *its own* wait via ``FetchMode``; the underlying fetch
+/// task is unaffected by individual callers timing out and continues until it
```
evan-masseau, line 23: great detail, is there also an overall/independent timeout on the fetch, or do we leave that up to the host app's implementation?

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -31,73 +52,188 @@ package actor AuthTokenManager {
...
-    /// A single caller invokes the provider directly. If two callers race, both
-    /// will invoke the provider — the result is extra network calls, not data
-    /// corruption.
+    /// Concurrent callers share the in-flight fetch — exactly one provider
+    /// invocation runs even under N simultaneous calls. Each caller's wait is
+    /// bounded by ``FetchMode``; on timeout the *caller* throws
+    /// ``AuthTokenError/timedOut`` while the underlying fetch task continues
+    /// (and may still warm the cache for later callers).
+    ///
+    /// The SDK does not enforce an upper bound on the underlying fetch's
```
evan-masseau, line 99: nvm that answers my question

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -31,73 +52,188 @@ package actor AuthTokenManager {
...
+    /// (and the cache-warming path) may still benefit if it eventually succeeds.
+    ///
+    /// Implemented via two detached tasks racing into a one-shot resolver
+    /// rather than `withThrowingTaskGroup`. A task group implicitly awaits all
+    /// children before its closure returns, and `Task.value` does not honor the
+    /// awaiter's cancellation — so a group-based race would block the timeout
+    /// path until the underlying provider returned anyway, defeating the
+    /// purpose. The resolver lets the loser's resume become a no-op while the
+    /// winning task drives the continuation.
+    private nonisolated func race(
```
evan-masseau, line 220: minor: surprised this sort of task timeout is a utility we had to build, and not something built in to swift. But given we built it fairly generically, worth extracting to a util function perhaps?

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -17,8 +17,29 @@ import OSLog
...
 package actor AuthTokenManager {
+    /// Latency budget for callers awaiting a token, expressed as a named policy
+    /// rather than a free `TimeInterval` so the two budgets stay
+    /// single-source-of-truth and can be tuned together based on production
+    /// telemetry.
+    package enum FetchMode: TimeInterval {
+        /// 500ms budget. Used at form-display time, where a user is waiting on
+        /// the form to appear — missing the personalization window is preferable
+        /// to delaying the form.
+        case bestEffort = 0.5
```
evan-masseau: nit: these names aren't obvious to me what kind of timing they'll actually mean. Maybe `short` and `long` or like `foreground` and `background` to indicate a priority level?

evan-masseau (approved):

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
ab1470: good point. I changed them to `interactive` and `background`, which borrows terminology from Swift's `dispatchQueue`'s QOS levels (https://developer.apple.com/documentation/dispatch/dispatchqos)

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
ab1470: this `race` method implements a pretty specific use-case of not cancelling the ongoing task in the event that the timeout expires. There isn't a built-in Swift utility for this, and I think it's specialized enough that a) I don't think it's going to be used elsewhere (YAGNI), and b) if something like it _is_ needed elsewhere it might have specific requirements that differ from this one and I'd rather re-implement it for that situation. If we ever do need to reuse exactly this logic then we can extract it to a shared utility at that point

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -31,73 +52,188 @@ package actor AuthTokenManager {
...
         guard let provider else {
             throw AuthTokenError.noProviderRegistered
         }
 
-        // TODO: [MAGE-624] cancel any in-flight fetch when `registerProvider(_:)`
-        // runs — the resumed continuation below can otherwise overwrite the cache
-        // with a token from a since-replaced provider.
-        let rawToken = try await provider()
+        do {
+            let rawToken = try await provider()
```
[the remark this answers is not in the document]
ab1470: Good catch. Fixed in dee86596 — added `try Task.checkCancellation()` at the top of `runFetch`, before it reads `provider`, so a task cancelled before its body ran bails immediately instead of invoking the swapped-in provider. The existing checkpoint after `provider()` still covers cancellation that lands mid-fetch.

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
glenn-klaviyo: Thanks for explaining!

glenn-klaviyo (approved):
LGTM

PR #588, opened 2026-05-29
feat(KlaviyoCore): AuthTokenManager proactive refresh scheduling (MAGE-625)

ab1470:
# Description

Adds proactive refresh to `AuthTokenManager` so personalized in-app forms have a fresh JWT ready before the cached one expires, without forcing form-display callers to pay the provider round-trip. Stacks on **MAGE-624** (dedup + timeouts) — targeting that branch until 624 merges to `rel/*`.

## Due Diligence

No public API changes — `AuthTokenManager` is `package`. The new `init` parameters (`lifeCycle`, `currentDate`) default to the existing `environment` closures, so `AuthTokenManager.shared` and all current call sites are unaffected.

## Changelog / Code Overview

**Refresh scheduling**
- Target: `iat + 0.9 * (exp - iat)`, clamped to `[now + 5s, exp - JWTParser.defaultLeeway]`. The upper clamp matches the same skew window `JWTParser` uses for the expiry check, so refresh fires before the cache itself would be considered stale. The lower clamp prevents tight refresh loops when a token is acquired close to its own expiration.
- Extracted as a static pure function `AuthTokenManager.refreshTarget(for:currentDate:)` so tests can verify the formula and clamps directly without driving the full schedule-and-sleep lifecycle.

**Wall-clock-aware sleep loop**
- `refreshAtWallClock: Date?` stores the absolute target; the loop re-reads `currentDate()` on every wakeup rather than trusting `Task.sleep`'s elapsed duration. Self-corrects automatically after backgrounding.

**Foreground transition handler**
Triages three cases on each `.foregrounded` event:
1. **Expired cached token** — clear cache, cancel refresh, kick off eager fetch.
2. **Missed scheduled refresh** — cancel stuck refresh task, fire immediately.
3. **Still valid** — no-op.

**Lifecycle observer**
Bridges `AppLifeCycleEvents`'s Combine publisher into an `AsyncStream`:
- iOS 15+: `Publisher.values` (cleanest path).
- iOS 13/14: `AsyncStream + sink` fallback, mirroring the pattern established by `LifecycleObserver` in KlaviyoForms.
- The wrapping `AsyncStream` keeps the consumer call site homogeneous; when the SDK eventually bumps its iOS minimum to 15 the `else` branch is a clean delete.

**Dedup sharing**
`performScheduledRefresh()` routes through the same `inFlight` slot that user-driven `currentToken(mode:)` callers use, so concurrent demand + the scheduled refresh collapse onto a single provider invocation.

**Logging**
All new OSLog lines gated `@available(iOS 14.0, *)`. Refresh failures log at `warning`, not `error` — the cached token remains valid until `exp - leeway`, and a foreground transition or user fetch will retry naturally.

**Out of scope**
- `ConnectivityMonitoring` and network-failure retry → MAGE-683
- Public refresh-stream notification API → MAGE-626 (a placeholder comment marks where the hook will live)
- `clearTokenState()` and profile-reset interaction → MAGE-626

**Scaffolding note**
MAGE-623 intentionally deferred the refresh-related actor properties (`refreshTask`, `refreshAtWallClock`, `lifecycleObserverTask`, `lifeCycle` injection) to this ticket so the design decisions land alongside the implementation that uses them.

## Test Plan

**Automated** — 9 new tests in `AuthTokenManagerRefreshTests.swift`, all passing alongside the 38 existing auth/JWT tests.

Pure-function tests (no real time):
- [x] `refreshTargetLandsAt90PercentOfTokenLifetime` — happy-path 90%-of-lifetime calculation.
- [x] `refreshTargetUsesIdealPointWhenInsideClamps` — neither clamp fires.
- [x] `refreshTargetClampsToExpMinusLeewayWhenIdealExceedsUpperBound` — upper clamp.
- [x] `refreshTargetClampsToNowPlusFiveWhenIdealAlreadyPast` — lower clamp.

Integration tests (real time):
- [x] `refreshFiresAtScheduledTimeAndChainsNextSchedule` — scheduled refresh fires, replaces cache, chains next schedule.
- [x] `foregroundWithStillValidTokenIsNoOp` — no extra provider invocation when cache is healthy.
- [x] `nonForegroundLifecycleEventsAreIgnored` — backgrounded/terminated/reachabilityChanged don't trigger the handler.
- [x] `foregroundWithExpiredCachedTokenClearsCacheAndRefetches` — case A of the foreground handler.
- [x] `registerProviderCancelsPriorScheduledRefresh` — provider swap cancels the pending refresh task.

**Manual smoke (recommended for reviewer)**
- [ ] Integrate a test app with a short-lived JWT provider (60s lifetime). Filter `Console.app` on the `Klaviyo.auth` subsystem; verify `refresh scheduled` → `refresh succeeded` cadence at ~54s.
- [ ] Background the test app past the scheduled refresh time, then foreground. Expect `foreground transition (case=missed-refresh)` or `(case=expired-cached-token)` followed by a fresh fetch.

## Related Issues/Tickets
- Linear: [MAGE-625](https://linear.app/klaviyo/issue/MAGE-625/ios-authtokenmanager-proactive-refresh-scheduling)
- Parent: [MAGE-615](https://linear.app/klaviyo/issue/MAGE-615) (JWT support in SDK/onsite)
- Stacks on: MAGE-624 (`ab/MAGE-624/authtokenmanager-dedup-and-timeouts`)
- Follows up with: MAGE-626 (refresh stream + profile reset), MAGE-683 (connectivity retry)

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable proactive token refreshes with robust handling across background/foreground transitions.
  * Prevents stale refreshes and avoids unnecessary provider calls when returning to foreground.
  * Improved token validity checks to reduce unexpected expirations.

* **Tests**
  * New deterministic tests covering refresh timing, lifecycle behavior, and re-registration scenarios.

* **Chores**
  * Consolidated and centralized authentication test utilities.

### DIFF
--- a/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
+++ b/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
@@ -5,6 +5,7 @@
 //  Created by Andrew Balmer on 2026-05-14.
 //
 
+import Combine
 import Foundation
 import OSLog
 
@@ -62,7 +63,46 @@ package actor AuthTokenManager {
     /// task is present without its id, or vice versa.
     private var inFlight: (id: UUID, task: Task<String, Error>)?
 
-    package init() {}
+    /// Sleep-and-refresh task that fires at ``refreshAtWallClock``. Cancelled
+    /// and replaced when a new token is acquired (chaining), when
+    /// ``registerProvider(_:)`` runs, or when the foreground transition handler
+    /// detects that the wall-clock target has already passed.
+    private var refreshTask: Task<Void, Never>?
+
+    /// Absolute wall-clock target for the next proactive refresh, or `nil` when
+    /// no refresh is scheduled. Stored as an absolute `Date` rather than a
+    /// duration so the sleep loop can re-check against the *current* clock on
+    /// every wakeup — `Task.sleep` drifts during backgrounding, so trusting the
+    /// elapsed-sleep duration alone would fire late.
+    private var refreshAtWallClock: Date?
+
+    /// Long-lived task that consumes the lifecycle event stream and dispatches
+    /// foreground transitions to ``handleForegroundTransition()``. Bound to the
+    /// actor's lifetime (started in ``init`` and survives ``registerProvider(_:)``).
+    private var lifecycleObserverTask: Task<Void, Never>?
+
+    /// Lifecycle event source. Injected for testability; defaults to the
+    /// SDK-wide `environment.appLifeCycle`.
+    private let lifeCycle: AppLifeCycleEvents
+
+    /// Wall-clock source. Injected for testability; defaults to the SDK-wide
+    /// `environment.date` closure. Used by both cached-token validity checks
+    /// and the proactive-refresh scheduling formula so a single time source
+    /// drives every clock-sensitive decision the actor makes.
+    private let currentDate: () -> Date
+
+    /// - Parameters:
+    ///   - lifeCycle: Source of foreground/background events. Defaults to
+    ///     `environment.appLifeCycle`.
+    ///   - currentDate: Wall-clock source. Defaults to `environment.date`.
+    package init(
+        lifeCycle: AppLifeCycleEvents = environment.appLifeCycle,
+        currentDate: @escaping () -> Date = { environment.date() }
+    ) {
+        self.lifeCycle = lifeCycle
+        self.currentDate = currentDate
+        Task { await self.startLifecycleObserver() }
+    }
 
     /// Registers a new provider, discards any cached token from a previous
     /// provider, cancels any in-flight fetch, and triggers an eager fetch to
@@ -75,6 +115,9 @@ package actor AuthTokenManager {
     package func registerProvider(_ newProvider: @escaping AuthTokenProvider) async {
         inFlight?.task.cancel()
         inFlight = nil
+        refreshTask?.cancel()
+        refreshTask = nil
+        refreshAtWallClock = nil
         cachedToken = nil
         provider = newProvider
 
@@ -171,6 +214,7 @@ package actor AuthTokenManager {
             switch JWTParser.parseAndValidate(rawToken) {
             case let .success(validated):
                 cachedToken = validated
+                scheduleRefresh(for: validated)
                 if #available(iOS 14.0, *) {
                     Logger.auth.info(
                         """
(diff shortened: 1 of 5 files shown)


Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -198,6 +242,191 @@ package actor AuthTokenManager {
...
+        }
+    }
+
+    /// Bridges the Combine-backed ``AppLifeCycleEvents`` publisher into an
+    /// `AsyncStream`. iOS 15+ uses `Publisher.values`; iOS 13/14 falls back to
+    /// `sink`-into-continuation (the pattern already established by
+    /// `LifecycleObserver` in `KlaviyoForms`). The wrapping `AsyncStream` keeps
+    /// the consumer call site homogeneous, and when the SDK eventually bumps
+    /// its iOS minimum to 15 the `else` branch is a clean delete.
+    private func makeLifecycleStream() -> AsyncStream<LifeCycleEvents> {
```
evan-masseau: Gentle pushback on the dual-path approach here. The iOS 15+ `publisher.values` path is idiomatic, but when wrapped inside `AsyncStream { continuation in }` it requires a nested `Task` + `onTermination` cancellation — which ends up being roughly the same complexity as the `sink` fallback. There's no meaningful performance difference for a low-frequency lifecycle stream. Since iOS 13 support isn't going away soon and the two paths have to be maintained in parallel, it might be simpler to just use the `sink` path everywhere and delete the `#available` branch. Am I missing something that makes the 15+ path worth keeping?

evan-masseau (approved):
LGTM, left one question on the lifecycle stream

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
ab1470: I simplified it significantly by removing this `makeLifecycleStream()` method. I bias towards `AsyncStream` for emitting events because it's the most modern way to do it, but in this case we already emit lifecycle events via a Combine `AnyPublisher`, and that works perfectly well here. Combine is still useful for certain things even in the "Swift Concurrency" era, so no need to add our own wrapper to convert it to an AsyncStream

As for the dual-path approach, it wasn't about a performance optimization. When there's a more modern API, I prefer to use that when possible, only falling back to an older API when limited by version constraints. The `#available` check and `if/else` kind of serve as an in-code reminder that we have some old code in there. Someday we'll drop support for older OSes, and having the conditional in there it makes it easy to do a find/replace and remove the old implementation.

Not an issue anymore though bc I've removed the method

PR #590, opened 2026-06-01
feat(push): OnceCallback + forwarding-cycle guard for KlaviyoNotificationDelegate

glenn-klaviyo:
# Description
Completes the remaining MAGE-657 acceptance criteria on top of the injection/KVO wiring from PR #587: wraps all UN completion handlers in an at-most-once, always-on-main-thread callback, and adds a forwarding-cycle guard that prevents stack overflow when multiple push SDKs form a circular delegate chain.

## Due Diligence

### `OnceCallback<Input>` — `Sources/KlaviyoSwift/Utilities/OnceCallback.swift`
- New utility class that wraps a completion handler so it fires at most once regardless of how many callers invoke it
- Thread-safe via `NSLock`; first concurrent caller wins
- Always dispatches on the main thread — `UNUserNotificationCenter` completion handlers trap with `NSInternalInconsistencyException` when called off-main

### `ForwardingCycleGuard` — `Sources/KlaviyoSwift/Utilities/ForwardingCycleGuard.swift`
- New utility class that tracks in-flight notification request IDs to detect and break delegate forwarding cycles
- `begin(_:)` atomically checks and inserts; `end(_:)` removes — both `NSLock`-protected
- `KlaviyoNotificationDelegate` holds two private instances: one for `didReceive`, one for `willPresent`

### `KlaviyoNotificationDelegate` — `Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift`
- `didReceive`: calls `KlaviyoSDK().handle()` + forwards to `existingDelegate` via `OnceCallback`; trailing `once()` is the safety-net fallback
- `willPresent`: forwards to `existingDelegate` first (so host options win on sync path), then falls back to Klaviyo defaults via `OnceCallback`
- Both methods guard against re-entrant calls with `ForwardingCycleGuard`; cycle detection logs a warning and aborts the nested forward

## Test Plan
- `make test-library` passes
- `swiftlint --strict` passes on all changed files
- New test suites: `OnceCallbackTests` (5 tests incl. off-main dispatch + 32-thread concurrent load), `ForwardingCycleGuardTests` (4 tests)
- Basic proxy flow (KVO re-injection, `SceneDelegate` scenario) was manually verified in the prior session

## Related Issues/Tickets
- Linear: MAGE-657
- Depends on: #587 (squash-merged into `feat/auto-push-tracking`)
- Blocks: MAGE-659, MAGE-660, MAGE-661

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented notification forwarding loops and duplicate callback invocations, ensuring each notification is handled exactly once.
  * Improved reliability and consistent presentation behavior for incoming notifications across concurrent scenarios.

* **Tests**
  * Added tests covering cycle detection and single-invocation guarantees for notification callbacks.

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift
+++ b/Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift
@@ -72,6 +72,11 @@ final class KlaviyoNotificationDelegate: NSObject {
     /// without requiring `nonisolated(unsafe)` or `@unchecked Sendable` annotations.
     private var centerObservation: AnyObject?
 
+    // MARK: - Forwarding-cycle guard
+
+    private let didReceiveGuard = ForwardingCycleGuard()
+    private let willPresentGuard = ForwardingCycleGuard()
+
     // MARK: - Injection
 
     /// Reads the opt-in flag and notification center from `KlaviyoSwiftEnvironment`,
@@ -122,10 +127,22 @@ extension KlaviyoNotificationDelegate: UNUserNotificationCenterDelegate {
         didReceive response: UNNotificationResponse,
         withCompletionHandler completionHandler: @escaping @Sendable () -> Void
     ) {
+        let requestId = response.notification.request.identifier
+        guard didReceiveGuard.begin(requestId) else {
+            if #available(iOS 14.0, *) {
+                Logger.notifications.warning("Klaviyo: forwarding cycle detected in didReceive, skipping.")
+            }
+            completionHandler()
+            return
+        }
+        defer { didReceiveGuard.end(requestId) }
+
         let once = OnceCallback(completionHandler)
-        // TODO: [MAGE-660] Add double-track guard
+        // MAGE-660: double-track guard (auto + manual paths) is a follow-on ticket.
         _ = KlaviyoSDK().handle(notificationResponse: response, withCompletionHandler: { once() })
-        existingDelegate?.userNotificationCenter?(center, didReceive: response, withCompletionHandler: { once() })
+        existingDelegate?.userNotificationCenter?(
+            center, didReceive: response, withCompletionHandler: { once() }
+        )
         once()
     }
 
@@ -135,8 +152,24 @@ extension KlaviyoNotificationDelegate: UNUserNotificationCenterDelegate {
         withCompletionHandler completionHandler: @escaping @Sendable
         (UNNotificationPresentationOptions) -> Void
     ) {
+        let requestId = notification.request.identifier
+        guard willPresentGuard.begin(requestId) else {
+            if #available(iOS 14.0, *) {
+                Logger.notifications.warning("Klaviyo: forwarding cycle detected in willPresent, skipping.")
+            }
+            if #available(iOS 14.0, *) {
+                completionHandler([.list, .banner, .badge, .sound])
+            } else {
+                completionHandler([.alert, .badge, .sound])
+            }
+            return
+        }
+        defer { willPresentGuard.end(requestId) }
+
         let once = OnceCallback(completionHandler)
-        existingDelegate?.userNotificationCenter?(center, willPresent: notification, withCompletionHandler: { once($0) })
+        existingDelegate?.userNotificationCenter?(
+            center, willPresent: notification, withCompletionHandler: { once($0) }
+        )
         if #available(iOS 14.0, *) {
             once([.list, .banner, .badge, .sound])
         } else {
--- a/Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift
+++ b/Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift
@@ -131,7 +131,7 @@ extension KlaviyoNotificationDelegate: UNUserNotificationCenterDelegate {
         let once = OnceCallback(completionHandler)
         guard didReceiveGuard.begin(requestId) else {
             if #available(iOS 14.0, *) {
-                Logger.notifications.warning("Klaviyo: forwarding cycle detected in didReceive, skipping.")
+                Logger.notifications.warning("Forwarding cycle detected in didReceive, skipping delegate forward.")
             }
             once()
             return
@@ -155,7 +155,7 @@ extension KlaviyoNotificationDelegate: UNUserNotificationCenterDelegate {
         let once = OnceCallback(completionHandler)
         guard willPresentGuard.begin(requestId) else {
             if #available(iOS 14.0, *) {
-                Logger.notifications.warning("Klaviyo: forwarding cycle detected in willPresent, skipping.")
+                Logger.notifications.warning("Forwarding cycle detected in willPresent, skipping delegate forward.")
                 once([.list, .banner, .badge, .sound])
             } else {
                 once([.alert, .badge, .sound])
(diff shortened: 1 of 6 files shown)


Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift
```diff
@@ -122,10 +127,23 @@ extension KlaviyoNotificationDelegate: UNUserNotificationCenterDelegate {
...
-        // TODO: [MAGE-657] Call KlaviyoSDK().handle() for auto-tracking
-        // TODO: [MAGE-660] Add double-track guard
-        completionHandler()
+        let requestId = response.notification.request.identifier
+        guard didReceiveGuard.begin(requestId) else {
+            if #available(iOS 14.0, *) {
+                Logger.notifications.warning("Klaviyo: forwarding cycle detected in didReceive, skipping.")
+            }
+            completionHandler()
+            return
```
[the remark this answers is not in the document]
glenn-klaviyo: should be addressed now

Sources/KlaviyoSwift/Utilities/OnceCallback.swift
```diff
@@ -0,0 +1,45 @@
...
+/// Wraps a callback so it fires at most once, regardless of how many callers invoke it.
+///
+/// Used in `UNUserNotificationCenterDelegate` proxy methods to ensure a completion
+/// handler is called exactly once even when both the proxy and a forwarded host delegate
+/// each attempt to invoke it.
+///
+/// Thread-safe: the first concurrent caller wins; all subsequent calls are no-ops.
+/// The body is always dispatched on the main thread — `UNUserNotificationCenter`
+/// completion handlers trap if called from a background thread.
+final class OnceCallback<Input>: @unchecked Sendable {
```
ab1470: I recently introduced [`OnceResolver`](https://github.com/klaviyo/klaviyo-swift-sdk/blob/feat/jwts-for-personalized-forms/Sources/KlaviyoCore/Auth/OnceResolver.swift) that does something similar. I'm about to move it into `KlaviyoCore/Utils` so it's more discoverable. This new `OnceCallback` could potentially go in `KlaviyoCore/Utils` as well if you think it might be reused, but I'll leave it up to you

ab1470: also, I asked Claude to review this so full disclosure this is from Claude, not me. But I think it's valid

--- 

### `OnceCallback`: tighten `Sendable` so `@unchecked` only covers the lock
  
  Nice utility — the latch logic looks correct (atomic check-and-clear under the
  lock, callback invoked *outside* the lock to avoid re-entrancy, and `body = nil`
  releases the captured closure). One change I'd suggest, plus a couple of optional
  notes.
  
  #### Make `@unchecked Sendable` honest
  
  The class legitimately needs `@unchecked` because `body` is a `var` mutated under
  a manual `NSLock` the compiler can't reason about. But `@unchecked` is currently
  also silently covering a *second*, unrelated gap: `callAsFunction` sends
  `callback` and `input` across a thread boundary via `DispatchQueue.main.async`,
  yet `Input` is unconstrained and `body` isn't `@Sendable`. The compiler can't
  verify that cross-thread send is safe — `@unchecked` just suppresses the question.
  
  I'd constrain it:
  
  ```swift
  final class OnceCallback<Input: Sendable>: @unchecked Sendable {
      private var body: (@Sendable (Input) -> Void)?
      private let lock = NSLock()
      
      init(_ body: @escaping @Sendable (Input) -> Void) {
          self.body = body
      }   
      // ...rest unchanged...
  }   
```
  
  This costs nothing at the current call sites — the completion handlers being
  wrapped are already @escaping @Sendable, and the willPresent payload
  (UNNotificationPresentationOptions) is a Sendable value type. It just
  documents the real requirement and confines @unchecked to the one thing it
  should cover (the lock-guarded var), instead of papering over a genuine
  cross-thread Sendable hole.

  It's latent rather than active today (we don't build under complete concurrency
  checking yet, so nothing warns), but the current form would hide a real data-race
  risk the moment this module enables strict concurrency or someone instantiates
  OnceCallback with a non-Sendable Input. Cheap to lock down now.

File: Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift [hunk 3]
```diff
@@ -122,10 +127,22 @@ extension KlaviyoNotificationDelegate: UNUserNotificationCenterDelegate {
...
     ) {
-        // TODO: [MAGE-657] Forward to existingDelegate via OnceCallback
-        // TODO: [MAGE-657] Call KlaviyoSDK().handle() for auto-tracking
-        // TODO: [MAGE-660] Add double-track guard
-        completionHandler()
+        let requestId = response.notification.request.identifier
+        let once = OnceCallback(completionHandler)
+        guard didReceiveGuard.begin(requestId) else {
+            if #available(iOS 14.0, *) {
+                Logger.notifications.warning("Klaviyo: forwarding cycle detected in didReceive, skipping.")
```
ab1470:
```suggestion
                Logger.notifications.warning("Forwarding cycle detected in didReceive, skipping.")
```
I think I'd leave this out. We don't add `Klaviyo` in any of our other log messages, because (a) doing so would add a lot of noise, and (b) OSLog automatically adds filterable metadata with Klaviyo identifiers, including the Library and Subsystem

Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift
```diff
@@ -134,11 +151,25 @@ extension KlaviyoNotificationDelegate: UNUserNotificationCenterDelegate {
         withCompletionHandler completionHandler: @escaping @Sendable
         (UNNotificationPresentationOptions) -> Void
     ) {
-        // TODO: [MAGE-657] Forward to existingDelegate when present
+        let requestId = notification.request.identifier
+        let once = OnceCallback(completionHandler)
+        guard willPresentGuard.begin(requestId) else {
+            if #available(iOS 14.0, *) {
+                Logger.notifications.warning("Klaviyo: forwarding cycle detected in willPresent, skipping.")
```
ab1470:
```suggestion
                Logger.notifications.warning("Forwarding cycle detected in willPresent, skipping.")
```

[same as above](https://github.com/klaviyo/klaviyo-swift-sdk/pull/590/changes#r3336220585)

Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift
ab1470: > Klaviyo: forwarding cycle detected in didReceive, skipping.

also, could you either write a code comment or expand the log message to explain _what_ you're skipping? When I see "skipping" I generally think "no-op", but it's not actually a no-op because you're calling `once()`

ab1470 (approved):

Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift
glenn-klaviyo: Nice suggestion! I will update the log to be clearer as to what it's skipping and remove the `"klaviyo:"` prefix.

Sources/KlaviyoSwift/Utilities/OnceCallback.swift
glenn-klaviyo: I like this suggestion. Now @unchecked Sendable only covers the ns lock

glenn-klaviyo: hmm I spent some time thinking about this one - for the time being I'll leave it in `KlaviyoSwift/Utilities`. I don't know this is general purpose enough to be used outside the UNUserNotificationCenter at this time (yet). It would be nice to be able to use `OnceResolver` as they have a similar idea of firing once, but this use case is different to the once resolver (async continuation) need. Happy to revisit this though

PR #594, opened 2026-06-02
feat(KlaviyoForms): push refreshed auth tokens into live IAF WebView (MAGE-616)

ab1470:
# Description
Subscribes In-App Forms to `AuthTokenManager.refreshes()` so proactively-refreshed JWTs are pushed into the live WebView, keeping `data-klaviyo-jwt` fresh whether or not a form is currently on screen. Completes the live-update portion (Phase 3) of MAGE-616.

## Due Diligence
  - Unit tests run on the iOS simulator; end-to-end manual verification of the JWT flow is pending the full feature stack landing.
  - Uses `AsyncStream` / `for await` / `Task` only; `make CONFIG=release test-library` (Swift 5.9) not yet run locally — relying on CI to confirm.

  - This PR is internal wiring only (no public API change); it's part of the larger public JWT feature (`registerAuthTokenProvider`) tracked across the MAGE-61x/62x stack.

## Changelog / Code Overview
Phase 3 of MAGE-616 — live token updates. The subscription is bound to the **WebView's lifetime** (the WebView sits alive between form displays), not to a single form display.

- **`IAFPresentationManager`** owns the subscription (`tokenRefreshTask`):
  - `startTokenRefreshObservation()` subscribes to `AuthTokenManager.shared.refreshes()` and pushes each token to the view model. Started in `createFormWebView` (the single creation chokepoint, after the webview is built) and cancelled in `destroyWebView` (the single destruction chokepoint — covers the API-key-change and session-expiry recreate paths too).
  - Consolidates all `AuthTokenManager` interaction in one layer (it already does the initial `currentToken()` injection) and matches its existing observer-task pattern (`companyEventsTask`, `lifecycleEventsTask`, `profileEventsTask`, `formEventTask`).
- **`IAFWebViewModel`** gains `pushAuthToken(_:)`, a direct analogue of the existing `pushDeviceInfo()` — builds the `data-klaviyo-jwt` script via the existing `createAuthTokenScript` (so initial-load injection and live updates share one builder) and pushes it via `delegate.evaluateJavaScript`. The token value is never logged.
- **Tests**: `MockIAFWebViewDelegate` records evaluated scripts; two focused unit tests cover `pushAuthToken` (single update carries the token; sequential updates apply in order).

## Test Plan
1. `xcodebuild test -scheme klaviyo-swift-sdk-Package -destination "platform=iOS Simulator,name=iPhone 17 Pro" -only-testing KlaviyoFormsTests` → all 60 pass, including `testPushAuthTokenUpdatesWebView` and `testPushAuthTokenAppliesUpdatesInOrder`.
2. Manual (pending full stack): with a registered auth-token provider and a form displayed, force a proactive refresh and confirm `document.head`'s `data-klaviyo-jwt` updates without a reload.

## Related Issues/Tickets
- [MAGE-616](https://linear.app/klaviyo/issue/MAGE-616) — iOS: inject auth token into In-App Forms WebView
- Stacked on `ab/MAGE-626/refresh-stream-and-profile-reset` (MAGE-626, in review); retarget to the appropriate base once that merges.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced authentication token refresh mechanism for in-app forms, ensuring tokens remain valid and properly updated during active form sessions.

* **Tests**
  * Added comprehensive test coverage for token refresh functionality, including validation that updates are applied correctly and in proper order.

[Review Change Stack](https://app.coderabbit.ai/change-stack/klaviyo/klaviyo-swift-sdk/pull/594?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -176,11 +176,20 @@ class IAFPresentationManager {
     /// via ``IAFWebViewModel/pushAuthToken(_:)`` so onsite always has a fresh
     /// token — whether or not a form is currently on screen.
     ///
-    /// Bound to the WebView's lifetime, not a single form display: started here
-    /// (the single WebView-creation site) and cancelled in ``destroyWebView()``
-    /// (the single destruction site). `self` (the shared manager) is captured
-    /// weakly and re-acquired inside the loop, matching the other observer tasks.
+    /// Bound to the WebView's lifetime, not a single form display: started on
+    /// WebView creation and cancelled in ``destroyWebView()``. `self` (the shared
+    /// manager) is captured weakly and re-acquired inside the loop, matching the
+    /// other observer tasks.
+    ///
+    /// Cancels any existing task before replacing it: `viewController` can be
+    /// cleared without going through ``destroyWebView()`` (e.g. a failed
+    /// presentation in ``presentFormAsModal(viewController:)``), after which a
+    /// reinit can call this again — without this cancel the prior task's handle
+    /// would be overwritten and its `refreshes()` loop would leak (the shared
+    /// manager never deallocates, so the `[weak self]` guard never trips),
+    /// double-pushing every future token.
     private func startTokenRefreshObservation() {
+        tokenRefreshTask?.cancel()
         tokenRefreshTask = Task { [weak self] in
             let stream = await AuthTokenManager.shared.refreshes()
             for await token in stream {
@@ -484,12 +493,16 @@ class IAFPresentationManager {
     // MARK: - Cleanup & Destruction
 
     func destroyWebView() {
+        // Cancel before the guard: `viewController` may already have been cleared
+        // elsewhere (e.g. a failed presentation) while the token-refresh task is
+        // still running, so gating the cancel on `viewController` would leak it.
+        tokenRefreshTask?.cancel()
+        tokenRefreshTask = nil
+
         guard let viewController else { return }
 
         performDismiss(viewController: viewController)
 
-        tokenRefreshTask?.cancel()
-        tokenRefreshTask = nil
         self.viewController = nil
         viewModel = nil
     }
--- a/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
@@ -36,6 +36,7 @@ class IAFPresentationManager {
 
     private var formEventTask: Task<Void, Never>?
     private var delayedPresentationTask: Task<Void, Never>?
+    private var tokenRefreshTask: Task<Void, Never>?
 
     lazy var indexHtmlFileUrl: URL? = {
         do {
@@ -166,6 +167,27 @@ class IAFPresentationManager {
         self.viewModel = viewModel
         viewController = KlaviyoWebViewController(viewModel: viewModel)
         viewController?.modalPresentationStyle = .overCurrentContext
+
+        startTokenRefreshObservation()
+    }
+
+    /// Subscribes to ``AuthTokenManager``'s proactive-refresh stream for the
+    /// lifetime of the WebView, pushing each refreshed token into the live page
+    /// via ``IAFWebViewModel/pushAuthToken(_:)`` so onsite always has a fresh
+    /// token — whether or not a form is currently on screen.
+    ///
+    /// Bound to the WebView's lifetime, not a single form display: started here
+    /// (the single WebView-creation site) and cancelled in ``destroyWebView()``
+    /// (the single destruction site). `self` (the shared manager) is captured
+    /// weakly and re-acquired inside the loop, matching the other observer tasks.
+    private func startTokenRefreshObservation() {
+        tokenRefreshTask = Task { [weak self] in
+            let stream = await AuthTokenManager.shared.refreshes()
+            for await token in stream {
+                guard let self else { return }
+                await self.viewModel?.pushAuthToken(token)
+            }
+        }
     }
 
     // MARK: - Form Lifecycle Listener Setup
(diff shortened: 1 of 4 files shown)


Tests/KlaviyoFormsTests/IAFWebViewModelTests.swift
```diff
@@ -377,6 +379,41 @@ final class IAFWebViewModelTests: XCTestCase {
...
+        let firstToken = "header.first.signature"
+        let secondToken = "header.second.signature"
+        await viewModel.pushAuthToken(firstToken)
+        await viewModel.pushAuthToken(secondToken)
+
+        // Then — both are applied, in order
+        let scripts = tokenScripts(delegate)
+        XCTAssertEqual(scripts.count, 2)
+        XCTAssertTrue(scripts[0].contains(firstToken))
+        XCTAssertTrue(scripts[1].contains(secondToken))
```
glenn-klaviyo, line 415: **suggestion:** since `[String]` is equatable, instead of asserting each element could we just assert that the arrays are equal?
```
XCTAssertEqual(scripts, [firstToken, secondToken])
```

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -240,6 +240,11 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
         return "document.head.setAttribute('data-klaviyo-profile', '\(profileDataString)');"
     }
 
+    @MainActor
+    private func createAuthTokenScript(from token: String) -> String {
+        "document.head.setAttribute('data-klaviyo-jwt', '\(token)');"
+    }
```
glenn-klaviyo, line 246: nice! 👍

Tests/KlaviyoFormsTests/IAFWebViewModelTests.swift
ab1470: the `.contains` is the key part here; `tokenScripts` returns the entire `document.head.setAttribute(...` script, not just the tokens, so your suggested assert would fail. We only want to validate that the returned scripts contain the `firstToken` and `secondToken`.

glenn-klaviyo: Ah I missed that. Thanks for the explanation!

Sources/KlaviyoForms/InAppForms/IAFPresentationManager.swift
```diff
@@ -166,6 +167,36 @@ class IAFPresentationManager {
...
+    /// double-pushing every future token.
+    private func startTokenRefreshObservation() {
+        tokenRefreshTask?.cancel()
+        tokenRefreshTask = Task { [weak self] in
+            let stream = await AuthTokenManager.shared.refreshes()
+            for await token in stream {
+                guard let self else { return }
+                await self.viewModel?.pushAuthToken(token)
+            }
+        }
```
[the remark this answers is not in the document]
ab1470: Good catch on the asymmetry — the `presentFormAsModal` "no `topMostViewController`" branch was only niling `viewController`, leaving `viewModel` and `tokenRefreshTask` alive. I've funneled teardown through a single `teardownWebViewState()` helper (cancels the task + nils `tokenRefreshTask`, `viewController`, and `viewModel`), called from both that branch and `destroyWebView()`, so the invariant "no `viewController` ⇒ no live `tokenRefreshTask` and no `viewModel`" always holds.

Two corrections on the finding itself:

- **Impact is lower than Major.** After that branch nils the manager's only strong ref to the never-presented VC, the VC deallocates and `viewModel.delegate` (weak) goes nil, so `pushAuthToken` → `delegate?.evaluateJavaScript` becomes a no-op. There's no on-screen WebView, so dropping those pushes is correct, not a silent loss of a visible update — and the task was already bounded (cancelled on the next `createWebView`/`destroyWebView`). It was a state-consistency gap, not a functional defect.
- **"Also applies to: 495–500" is `destroyWebView()`**, which already cancelled the task correctly — no change needed there. (It did have a smaller gap: when `viewController` was already nil it cancelled the task but left `viewModel` dangling; the shared helper now closes that too.)

glenn-klaviyo (approved):

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
```diff
@@ -261,6 +266,34 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
...
+    /// The token value is never logged — only the success/failure of the update.
+    @MainActor
+    func pushAuthToken(_ token: String) async {
+        if #available(iOS 14.0, *) {
+            Logger.webViewLogger.info("Auth token refreshed; updating In-App Forms HTML")
+        }
+        let authTokenScript = createAuthTokenScript(from: token)
+        do {
+            _ = try await delegate?.evaluateJavaScript(authTokenScript)
+            if #available(iOS 14.0, *) {
```
ndurell, line 287: Have we looked at what percentage of traffic is on 13? Can we bump the minimum?

ndurell (approved):

Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
ab1470: lol this has been discussed extensively. I would _love_ to bump the minimum, but the last time we chatted about it we decided not to because (a) our competitors still support iOS 13 and (b) we didn't want to take any chance of upsetting any customers who still support iOS 13.

According to Gemini, only 0.2% of iOS devices are running iOS 13. My personal opinion is that bumping our minimum to 14 would affect almost nobody. But I never pulled our own stats for iOS 13 usage in our SDK, so that would be a useful exercise to help inform this decision

ab1470: > only 0.2% of iOS devices are running iOS 13

and if I were to guess, I'd wager that those are people's secondary devices, like super old iPads that don't support newer iOS versions and sit on peoples' shelves most of the time

ab1470: Update: iOS 13 usage of our SDK's is very close to nil. See [this Slack post](https://klaviyo.enterprise.slack.com/archives/C08GN96SF1B/p1780598870000609)

PR #595, opened 2026-06-03
feat(KlaviyoCore): connectivity-driven auth token refresh retry (MAGE-683)

ab1470:
# Description
Adds connectivity-driven retry to `AuthTokenManager`: when a proactive token refresh fails because the device is offline, the manager waits for connectivity to be restored and retries, instead of letting the cached token go stale until the next foreground transition or user fetch.

## Due Diligence

## Changelog / Code Overview
**As-built note:** this reuses the SDK's existing reachability signal rather than the spec's `ConnectivityMonitoring`/`NWPathMonitor` design — decided with the assignee to keep the change DRY and avoid introducing `Network.framework`.

- **`URLError+Ext.swift` (new):** `URLError.isConnectivityError` classifies the four genuine-offline codes (`.notConnectedToInternet`, `.networkConnectionLost`, `.dataNotAllowed`, `.internationalRoamingOff`). Other failures (HTTP errors, validation rejections, generic timeouts) are deliberately excluded — waiting on connectivity wouldn't address them.
- **`AuthTokenManager`:**
  - A network-classified proactive-refresh failure arms an `isAwaitingConnectivityRetry` flag (in `performScheduledRefresh()`'s `catch`).
  - The actor's already-running lifecycle subscription now also handles `.reachabilityChanged(status:)`; on the next satisfied status it consumes the flag and re-fires `performScheduledRefresh()` (which goes through the normal dedup → validate → reschedule → broadcast path).
  - "At most one pending wait" is the single bool; rapid network flapping can't queue multiple retries. Cleared by `cancelInFlightWorkAndClearCache()`, so both `clearTokenState()` (profile reset) and `registerProvider()` drop a pending retry.
  - A failed retry re-arms for the next restoration.
- No new protocol, no `connectivityWaitTask` field, no `Network.framework` import.

## Test Plan
Automated (deterministic, virtual-clock + gated reachability — no real network):

```
  -destination "platform=iOS Simulator,name=iPhone 17 Pro" \
  -only-testing:KlaviyoCoreTests
```

Five new tests in `AuthTokenManagerRefreshTests.swift`, all green (full target: 56 tests / 3 suites, no hangs):
1. Network-failed refresh retries when connectivity is restored (and `.notReachable` is ignored).
2. A non-network failure does **not** arm a connectivity retry.
3. `clearTokenState()` cancels a pending wait.
4. Rapid satisfied transitions yield a **single** retry.
5. A failed retry re-arms for the next restoration.

Tests await arming deterministically via a documented internal test-only accessor (`isAwaitingConnectivityRetryForTesting`) rather than a fixed yield count, so the reachable event can't race ahead of arming and flake under parallel load.

## Related Issues/Tickets
- [MAGE-683](https://linear.app/klaviyo/issue/MAGE-683/ios-authtokenmanager-connectivity-driven-refresh-retry)
- Stacked on `ab/MAGE-616/live-token-updates-via-refresh-stream` (PR #594).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved token refresh reliability: the auth system now defers a single retry for network-classified failures, resumes automatically when connectivity is restored, avoids duplicate retries, and clears pending retries on successful refresh or explicit clears. Added network-error classification to better detect connectivity cases.
* **Tests**
  * Added deterministic tests and helpers to simulate reachability transitions and verify connectivity-driven retry behavior.

### DIFF
--- a/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
+++ b/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
@@ -76,6 +76,20 @@ package actor AuthTokenManager {
     /// elapsed-sleep duration alone would fire late.
     private var refreshAtWallClock: Date?
 
+    /// `true` while a proactive refresh has failed for a network reason and the
+    /// manager is waiting for connectivity to be restored before retrying.
+    ///
+    /// This is a single boolean rather than a dedicated task because connectivity
+    /// is observed through the already-running ``lifecycleCancellable`` sink (the
+    /// SDK's existing reachability signal), not a separate per-wait subscription.
+    /// The flag is the "at most one pending wait" guarantee: a network-classified
+    /// refresh failure arms it (see ``performScheduledRefresh()``), the next
+    /// reachable ``Reachability/NetworkStatus`` consumes it (see
+    /// ``handleReachabilityChange(_:)``), and rapid network flapping cannot queue
+    /// multiple retries. Cleared by ``cancelInFlightWorkAndClearCache()`` so a
+    /// provider swap or profile reset drops any pending retry.
+    private var isAwaitingConnectivityRetry = false
+
     /// Long-lived Combine subscription that dispatches foreground transitions to
     /// ``handleForegroundTransition()``. Bound to the actor's lifetime (started in
     /// ``init`` and survives ``registerProvider(_:)``).
@@ -279,6 +293,7 @@ package actor AuthTokenManager {
         refreshTask?.cancel()
         refreshTask = nil
         refreshAtWallClock = nil
+        isAwaitingConnectivityRetry = false
         cachedToken = nil
     }
 
@@ -439,7 +454,10 @@ package actor AuthTokenManager {
     /// *and* scheduled the next refresh (via the wiring inside that method).
     /// On failure: leaves the cached token in place — the cache only goes
     /// stale at `exp - leeway`, so a foreground transition or user fetch
-    /// before then will retry. Connectivity-driven retry is owned by MAGE-683.
+    /// before then will retry. A *network-classified* failure additionally arms
+    /// ``isAwaitingConnectivityRetry`` so the next reachability restoration
+    /// re-fires this method (see ``handleReachabilityChange(_:)``); other
+    /// failures do not, since waiting for connectivity would not address them.
     private func performScheduledRefresh() async {
         guard provider != nil else { return }
         let task = inFlight?.task ?? startFetch()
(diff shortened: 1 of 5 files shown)


Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -76,6 +76,29 @@ package actor AuthTokenManager {
...
+    /// multiple retries. Cleared by ``cancelInFlightWorkAndClearCache()`` so a
+    /// provider swap or profile reset drops any pending retry.
+    private var isAwaitingConnectivityRetry = false
+
+    /// Test-only window onto ``isAwaitingConnectivityRetry`` so suites can
+    /// deterministically await the *arming* of the connectivity wait — which
+    /// lands asynchronously inside the refresh-failure path — instead of racing
+    /// it with a fixed number of yields. Reachable from the test target via
+    /// `@testable import`; not part of the package API.
+    var isAwaitingConnectivityRetryForTesting: Bool {
```
ndurell, line 91: Nit: Should we just make the other property more visible for testing? This seems like an odd way to add testability to me.

Sources/KlaviyoCore/Utils/URLError+Ext.swift
```diff
@@ -0,0 +1,29 @@
...
+import Foundation
+
+extension URLError {
+    /// `URLError` codes that indicate a genuine offline condition — the only
+    /// failures for which waiting on connectivity restoration is the right
+    /// remedy. HTTP error responses, validation rejections, and generic
+    /// timeouts are deliberately excluded: re-running the same request once a
+    /// path exists would not change their outcome.
+    private static let connectivityCodes: Set<URLError.Code> = [
+        .notConnectedToInternet,
```
ndurell, line 16: Probably out of scope but I wonder if these are the same codes we use for errors in the main sdk.

ndurell (approved):

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
ab1470: it's a little clunky I agree, but I wanted to avoid exposing `isAwaitingConnectivityRetry` publicly because it shouldn't be used outside of the `AuthTokenManager` _except_ for testing. I figured that having a separate test-only property would help to make this clear

evan-masseau: is `@spi` an option here to expose it, but ... secretly

evan-masseau (approved):

PR #596, opened 2026-06-04
feat(push): swizzle didRegisterForRemoteNotificationsWithDeviceToken for automatic token forwarding

glenn-klaviyo:
# Description

Adds `KlaviyoAppDelegateSwizzler` so hosts that opt in via `klaviyo_automatic_push_tracking = YES` no longer need to call `KlaviyoSDK().set(pushToken:)` manually — the SDK intercepts `application(_:didRegisterForRemoteNotificationsWithDeviceToken:)` at the Obj-C runtime layer and forwards the token automatically.

In addition to the code comments, I created a doc explaining how the swizzler works under the hood: https://docs.google.com/document/d/1GCZqC4Cag-yCResYonjLhTb2kLLwKOV-7QYvYGEI9zw/edit?usp=sharing

## Due Diligence

No public API changes. Behavior is gated by `klaviyo_automatic_push_tracking` Info.plist key; hosts without the key see zero behavior change.

## Changelog / Code Overview

**New: `Sources/KlaviyoSwift/KlaviyoAppDelegateSwizzler.swift`**

Donor class that grafts a Klaviyo IMP onto the host's `UIApplicationDelegate` class at SDK init time. Handles three host configurations:

1. **Standard UIKit AppDelegate that implements `didRegister`** — `method_exchangeImplementations` swaps IMPs so calls hit our donor first, then chain to the host's original.
2. **Standard UIKit AppDelegate that does not implement `didRegister`** — `class_addMethod` grafts our donor directly under the original selector.
3. **SwiftUI `@UIApplicationDelegateAdaptor` hosts** — `UIApplication.shared.delegate` is Apple's private wrapper that forwards selectors to the host's real AppDelegate via `-forwardingTargetForSelector:`. `resolveSwizzleTargetClass(for:)` walks the forwarding chain so we swizzle the real implementor, not the wrapper.

Also handles the deferred-install case where `UIApplication.shared.delegate` is nil at SDK init time — registers a one-shot observer for `UIApplication.didFinishLaunchingNotification` and swizzles then.

Idempotent across repeat `initialize()` calls via a one-way `didSwizzle` flag in the private `State` singleton (`NSLock`-protected for Swift 6 strict concurrency).

**Modified: `Sources/KlaviyoSwift/KlaviyoNotificationDelegate.swift`**

One-line addition to `injectIfEnabled()` to call `KlaviyoAppDelegateSwizzler.swizzleIfPossible()` after `inject(into:)`. Both share the same plist gate.

**New: `Tests/KlaviyoSwiftTests/KlaviyoAppDelegateSwizzlerTests.swift`**

Swift Testing coverage for `resolveSwizzleTargetClass(for:)` — the genuinely non-trivial logic in this file. Three scenarios:
- Direct implementor → returns initial class
- Forwarding proxy (SwiftUI case) → returns inner class
- Opaque delegate → falls back to initial class

`performSwizzle` and the donor IMP are intentionally not unit-tested — they mutate global Obj-C runtime state irreversibly per process and are better verified once manually in the example app. See "Out of scope" below.

## Test Plan

**Automated:**
- [x] `make test-library` passes (Swift Testing, iPhone 17 Pro / iOS 26)
- [x] `swiftlint --strict` passes
- [x] `swiftformat` passes via pre-commit
- [x] Builds clean on iOS 18.6 simulator (locally available)
- [ ] CI confirms builds and tests pass on iOS 15.2 and 16.4

**Manual (recommended once MAGE-657 is also merged to `feat/auto-push-tracking`):**
1. SPMExample: set `klaviyo_automatic_push_tracking = YES` in Info.plist
2. Remove any manual `KlaviyoSDK().set(pushToken:)` call
3. Launch app, grant notification permission, observe log: `Swizzled didRegisterForRemoteNotificationsWithDeviceToken on <AppDelegate>`
4. Confirm push token reaches Klaviyo (check profile in dashboard)
5. Repeat with a SwiftUI `@UIApplicationDelegateAdaptor` host — should see the forwarding log line as well

## Out of scope

Documented per the ticket:
- Other AppDelegate methods (`didReceiveRemoteNotification:fetchCompletionHandler:`, `didFailToRegisterForRemoteNotificationsWithError:`) — push opens are handled by the proxy in MAGE-657; silent push stays manual.
- Hosts using `-forwardInvocation:` without `-forwardingTargetForSelector:` — resolver returns the wrapper class; SDK still gets the token but host's method bypassed. SwiftUI uses `forwardingTarget` so the common case is covered.
- Test coverage for `performSwizzle` exchange/add paths and deferred install — intentional, see PR body above.

## Related Issues/Tickets

Resolves MAGE-659

## Summary by CodeRabbit

* **New Features**
  * Automatic push token registration: the SDK now seamlessly handles device token setup with Klaviyo without requiring manual configuration in your app delegate.

* **Tests**
  * Added test coverage for automatic push token registration across different app delegate implementations.

### DIFF
--- a/Sources/KlaviyoSwift/KlaviyoAppDelegateSwizzler.swift
+++ b/Sources/KlaviyoSwift/KlaviyoAppDelegateSwizzler.swift
@@ -1,0 +1,280 @@
+//
+//  KlaviyoAppDelegateSwizzler.swift
+//  klaviyo-swift-sdk
+//
+//  Swizzles `application(_:didRegisterForRemoteNotificationsWithDeviceToken:)` on the host
+//  app delegate so the SDK receives push tokens automatically.
+//
+//  Created by Glenn Brannelly on 5/15/26.
+//
+
+import Foundation
+import ObjectiveC.runtime
+import OSLog
+import UIKit
+
+/// Donor class whose `klaviyo_application(_:didRegisterForRemoteNotificationsWithDeviceToken:)`
+/// IMP is grafted onto the host app's `UIApplicationDelegate` class at runtime so the SDK
+/// receives device tokens without the host having to call `KlaviyoSDK().set(pushToken:)`.
+///
+/// ## Concurrency model
+///
+/// Swizzling entry points (`swizzleIfPossible()`, `performSwizzle(on:)`) are invoked from
+/// `KlaviyoNotificationDelegate.injectIfEnabled()`, which is `@MainActor`. The grafted IMP
+/// (`klaviyo_application(_:didRegisterForRemoteNotificationsWithDeviceToken:)`) is fired by
+/// iOS on the main thread per the `UIApplicationDelegate` contract. All state mutation is
+/// therefore main-thread-only, but we serialize it through `State`'s `NSLock` so the type
+/// stays safely usable from any thread — Swift 6 strict-concurrency-clean without sprinkling
+/// `MainActor.assumeIsolated` through the IMP that the Obj-C runtime invokes.
+final class KlaviyoAppDelegateSwizzler: NSObject, @unchecked Sendable {
+    /// All mutable state lives behind a lock so the compiler can see it as `Sendable`.
+    /// Strict-concurrency mode flags raw `static var` declarations as global mutable state —
+    /// this wrapper keeps that requirement satisfied without sprinkling `nonisolated(unsafe)`.
+    ///
+    /// Fields:
+    /// - `didSwizzle`: one-way idempotence flag; flipped true the moment swizzling is claimed.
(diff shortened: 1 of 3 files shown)


glenn-klaviyo:
bugbot run

glenn-klaviyo:
@coderabbitai review

ajaysubra:
> @coderabbitai review

Does it not happen on it's own when you commit? I also recently learnt that you can actually chat with it in any code comments it makes.

ajaysubra (approved):
LGTM

glenn-klaviyo:
@coderabbitai configuration

PR #600, opened 2026-06-05
fix(KlaviyoCore): retry failed/missed scheduled refresh on foreground (MAGE-735)

ab1470:
# Description
Retry a failed or missed scheduled proactive auth-token refresh on the next foreground transition, matching Android (MAGE-629). Fixes the divergence introduced by the proactive-refresh scheduling work (MAGE-625, #588).

> **Stacked on [`ab/MAGE-683/connectivity-driven-refresh-retry`](https://github.com/klaviyo/klaviyo-swift-sdk/tree/ab/MAGE-683/connectivity-driven-refresh-retry)** — review/merge that first. This PR's diff is scoped to the MAGE-735 commits below.

## Due Diligence

## Changelog / Code Overview
**The bug:** In `AuthTokenManager.sleepUntilAndRefresh(target:)`, `refreshAtWallClock` was nil'd *before* the refresh attempt ran. The foreground handler's case-2 ("missed refresh") guard (`if let scheduled = refreshAtWallClock, currentDate() >= scheduled`) could therefore never match after a *failed* fire, so a transient failure on a still-valid token got no proactive retry until the token actually crossed `exp − leeway`.

**The fix:**
- **Stop nil-ing `refreshAtWallClock` before the attempt.** It is now replaced by `scheduleRefresh(for:)` on a successful refresh, or cleared on reset / by the foreground handler. A failed fire leaves the past target in place, so the next foreground transition retries it exactly once (case 2).
- **Add `isPerformingScheduledRefresh` and gate the foreground case-2 re-drive on it.** Because the marker now survives the fire, a foreground transition can land while a scheduled refresh is still in flight. Rather than cancelling that refresh's task (which suppresses its `refreshes()` broadcast) and re-driving it (risking a duplicate), the foreground handler leaves the in-flight refresh to complete — it broadcasts once and reschedules. This also covers the connectivity-retry-vs-foreground race.

## Test Plan
Three tests added to `AuthTokenManagerRefreshTests`, mirroring Android's `foreground retries after scheduled timer refresh fails`:
- `foregroundRetriesAfterFailedScheduledRefresh` — a network-failed scheduled refresh is retried on foreground while the token is still valid (fails on pre-fix code).
- `concurrentForegroundDuringInFlightScheduledRefreshBroadcastsOnce` — a foreground transition during an in-flight refresh leaves it to complete and broadcast once, without re-invoking the provider.
- `foregroundAfterSuccessfulScheduledRefreshDoesNotRefire` — a succeeded refresh reschedules to a future target, so a later foreground is a no-op.

Run:
```
xcodebuild test -scheme klaviyo-swift-sdk-Package -destination "platform=iOS Simulator,name=iPhone 17 Pro" -only-testing:KlaviyoCoreTests
```
Result: **61 tests, 0 failures.** SwiftLint + SwiftFormat clean (pre-commit passed).

## Related Issues/Tickets
- MAGE-735 (this work) · MAGE-625 / #588 (introduced the divergence) · Android parity: klaviyo-android-sdk#469 (MAGE-629)

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevent duplicate scheduled refreshes during foreground transitions; allow in-flight refreshes to continue and avoid duplicate broadcasts.
  * Preserve scheduled refresh targets after failed attempts so retries on next app interaction can succeed.
  * Clear stale scheduled-refresh state when cancelling in-flight work.

* **Tests**
  * Added async tests covering scheduled refresh failure, concurrency, retry and non-refire scenarios.
  * Enhanced test helper to await a specified number of token emissions.

### DIFF
--- a/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
+++ b/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
@@ -76,10 +76,12 @@ package actor AuthTokenManager {
     /// elapsed-sleep duration alone would fire late.
     private var refreshAtWallClock: Date?
 
-    /// `true` while ``performScheduledRefresh()`` is mid-flight, so the foreground
-    /// "missed refresh" path (case 2 in ``handleForegroundTransition()``) can leave a
-    /// running refresh alone rather than cancelling and re-driving it. Cleared via `defer`.
-    private var isPerformingScheduledRefresh = false
+    /// Names the in-flight ``performScheduledRefresh()`` run, or `nil` when none is
+    /// mid-flight, so the foreground "missed refresh" path (case 2 in
+    /// ``handleForegroundTransition()``) can leave a running refresh alone. Keyed by a
+    /// per-run `id` (like ``inFlight``) so a stale run whose `task.value` never returns
+    /// can't leave the guard stuck or clear a newer generation's run.
+    private var activeScheduledRefreshID: UUID?
 
     /// `true` while a proactive refresh has failed for a network reason and the
     /// manager is awaiting connectivity before retrying. A single boolean (not a
@@ -309,6 +311,7 @@ package actor AuthTokenManager {
         refreshTask?.cancel()
         refreshTask = nil
         refreshAtWallClock = nil
+        activeScheduledRefreshID = nil
         isAwaitingConnectivityRetry = false
         cachedToken = nil
     }
@@ -483,12 +486,19 @@ package actor AuthTokenManager {
     /// ``isAwaitingConnectivityRetry`` so the next reachability restoration
     /// re-fires this method (``handleReachabilityChange()``); other failures don't.
     ///
-    /// Sets ``isPerformingScheduledRefresh`` for its duration so a concurrent
+    /// Sets ``activeScheduledRefreshID`` for its duration so a concurrent
     /// foreground transition leaves an in-flight refresh to complete (case 2).
     private func performScheduledRefresh() async {
         guard provider != nil else { return }
-        isPerformingScheduledRefresh = true
-        defer { isPerformingScheduledRefresh = false }
+        let refreshID = UUID()
+        activeScheduledRefreshID = refreshID
+        defer {
+            // Only clear if we still own the generation — a reset (or a newer run)
+            // may have rotated it while this run was suspended on `task.value`.
+            if activeScheduledRefreshID == refreshID {
+                activeScheduledRefreshID = nil
+            }
+        }
         let task = inFlight?.task ?? startFetch()
         do {
             let token = try await task.value
@@ -625,7 +635,7 @@ package actor AuthTokenManager {
         if let scheduled = refreshAtWallClock, currentDate() >= scheduled {
             // An in-flight refresh is left to finish rather than cancelled and
             // re-driven, which would drop its broadcast or duplicate it.
-            guard !isPerformingScheduledRefresh else {
+            guard activeScheduledRefreshID == nil else {
                 if #available(iOS 14.0, *) {
                     Logger.auth.info(
                         "AuthTokenManager: foreground transition (case=missed-refresh-already-running)"
--- a/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
+++ b/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
@@ -76,21 +76,9 @@ package actor AuthTokenManager {
     /// elapsed-sleep duration alone would fire late.
     private var refreshAtWallClock: Date?
 
-    /// `true` while ``performScheduledRefresh()`` is mid-flight. Read by the
-    /// foreground "missed refresh" path (case 2 in ``handleForegroundTransition()``)
-    /// so a transition that lands mid-refresh leaves the in-flight refresh to
-    /// finish instead of cancelling and re-driving it.
-    ///
-    /// Why this matters: ``refreshAtWallClock`` is no longer cleared before a
-    /// fired refresh runs, so a foreground transition can match the still-set
-    /// past target while a scheduled refresh is suspended on its fetch. Cancelling
-    /// that refresh's task (as case 2 otherwise does) makes its `await task.value`
-    /// throw and *suppresses its broadcast*; re-driving would then either
-    /// double-broadcast or, with the fetch already cleared, miss it. Letting the
-    /// in-flight refresh complete yields exactly one ``refreshes()`` emission and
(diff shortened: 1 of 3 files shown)


Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -450,14 +457,18 @@ package actor AuthTokenManager {
     /// duration, so in the happy path the body runs at most ~2–3 times across
     /// the entire scheduled window (one long sleep, then a final short
     /// iteration that observes `remaining <= 0` and breaks).
+    ///
+    /// ``refreshAtWallClock`` is deliberately *not* cleared before the attempt, so
+    /// a fired refresh that fails while the token is still valid leaves the target
+    /// in place for the next foreground transition to retry (case 2). Success
+    /// replaces it via ``scheduleRefresh(for:)``; resets clear it.
     private func sleepUntilAndRefresh(target: Date) async {
```
dan-peluso, line 466: kind of a general nit, but I'm not super into this file being almost completely comments. I get that having explanations for some things are nice, but I think this creeps into it being a bit much.

dan-peluso (approved):

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
ab1470: lol yeah, we discussed this last week [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/592#issuecomment-4625723841) and [here](https://klaviyo.slack.com/archives/C08GN96SF1B/p1780590029950959). Claude is pretty verbose with code comments and documentation. Since that chat I've been manually editing the comments a bit and/or asking Claude to pare things down

Tests/KlaviyoCoreTests/Auth/AuthTokenManagerRefreshTests.swift
```diff
@@ -247,6 +247,186 @@ struct AuthTokenManagerRefreshTests {
         #expect(resolved == freshToken)
     }
 
+    @Test
+    func foregroundRetriesAfterFailedScheduledRefresh() async throws {
+        // Android parity (MAGE-629): a scheduled refresh that fires and *fails*
```
evan-masseau: double whammy comment nit: name-checking the other repo, and a ticket number

ab1470: I think it did that because I asked Claude to compare one of my iOS PRs with the open Android PR for MAGE-629, and find any parity gaps. This ticket (MAGE-735) emerged out of that comparison, and it's specifically about bringing iOS into parity with MAGE-629

I'm gonna do a comment cleanup pass tho

evan-masseau: We should make like a comment manifesto for claude

evan-masseau (approved):

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -76,6 +76,11 @@ package actor AuthTokenManager {
     /// elapsed-sleep duration alone would fire late.
     private var refreshAtWallClock: Date?
 
+    /// `true` while ``performScheduledRefresh()`` is mid-flight, so the foreground
+    /// "missed refresh" path (case 2 in ``handleForegroundTransition()``) can leave a
+    /// running refresh alone rather than cancelling and re-driving it. Cleared via `defer`.
+    private var isPerformingScheduledRefresh = false
```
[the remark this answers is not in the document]
ab1470: Good catch — hardened in 8782da7b. Replaced the bare `isPerformingScheduledRefresh` Bool with a generation-keyed `activeScheduledRefreshID: UUID?`, mirroring the existing `inFlight` ownership pattern in this file. The run's `defer` now clears the flag only when it still owns the current generation, and `cancelInFlightWorkAndClearCache()` rotates the ID on teardown so a stale run hung on `task.value` can't leave the guard stuck `true` or clobber a newer generation's run.

Tests/KlaviyoCoreTests/Auth/AuthTokenManagerRefreshTests.swift
```diff
@@ -247,6 +247,174 @@ struct AuthTokenManagerRefreshTests {
...
+
+        await releaseRefresh.open()
+        let delivered = await firstElement(of: stream)
+        #expect(
+            delivered == secondToken,
+            "the in-flight refresh must complete and broadcast despite the concurrent foreground"
+        )
+        let invocations = await counter.value
+        #expect(invocations == 2, "the in-flight fetch must be shared, not re-invoked, saw \(invocations)")
+    }
```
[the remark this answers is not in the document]
ab1470: Good catch — fixed in cf7ee897. You're right that `invocations == 2` only proves the fetch was shared and `firstElement` hides a duplicate `send`. The test now subscribes via `TokenCollector` up front (so no broadcast is missed), yields generously to let any duplicate land, then asserts `delivered == [secondToken]` — exactly one emission — alongside the retained `invocations == 2` check.

Sources/KlaviyoCore/Auth/AuthTokenManager.swift
```diff
@@ -609,6 +633,16 @@ package actor AuthTokenManager {
             return
         }
         if let scheduled = refreshAtWallClock, currentDate() >= scheduled {
+            // An in-flight refresh is left to finish rather than cancelled and
```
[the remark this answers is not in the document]
ab1470: Confirmed real and fixed in 815ad77b6ef4e32713896d4af30f1ae4ac4d736c. You're right that `fireConnectivityRetryIfArmed` didn't honor the `activeScheduledRefreshID` guard: `isAwaitingConnectivityRetry` only clears once the shared fetch succeeds, so a reachability event arriving while a foreground-driven refresh was suspended on `task.value` started a second `performScheduledRefresh` that awaited the same fetch and broadcast twice.

Rather than duplicate the guard at each call site, I moved the single-flight check into `performScheduledRefresh` itself (`guard activeScheduledRefreshID == nil else { return }`), so the scheduled fire, foreground retry, and connectivity retry are all covered by one chokepoint. The foreground path keeps its own pre-check so it doesn't tear down a live schedule.

Added a regression test (`concurrentConnectivityRetryDuringForegroundRetryBroadcastsOnce`) that parks a foreground retry in-flight, fires a reachability event, and asserts a single broadcast — verified to fail with a duplicate `[secondToken, secondToken]` against the pre-fix code and pass after.

PR #603, opened 2026-06-05
feat(KlaviyoSwift): public unregisterAuthTokenProvider API (MAGE-685)

ab1470:
# Description
Adds a public `KlaviyoSDK.unregisterAuthTokenProvider()` so host apps can detach a previously registered auth token provider — e.g. on user logout — and tear down all associated token state.

## Due Diligence

## Changelog / Code Overview
- **`KlaviyoSDK.unregisterAuthTokenProvider()`** (`KlaviyoSwift`, public): mirrors `registerAuthTokenProvider`; wraps the async actor call in a fire-and-forget `Task`.
- **`AuthTokenManager.unregisterProvider()`** (`KlaviyoCore`, `package`): two-step teardown — clears the provider reference, then calls the existing `clearTokenState()` to cancel the scheduled refresh, cancel any in-flight fetch and pending connectivity retry, and drop the cached token. Provider is cleared **before** `clearTokenState()` so callbacks resolving on cancelled tasks find no provider to call back into.

Unlike `clearTokenState()` (used by `resetProfile`, which *retains* the provider so the next `currentToken()` re-acquires), unregister drops the provider entirely — token acquisition stops until a new provider is registered.

## Test Plan
Five new cases in `AuthTokenManagerRefreshTests` (all green; full `AuthTokenManager*` suites pass — 42 tests):
1. Idle unregister drops the provider (next fetch throws `.noProviderRegistered`).
2. Pending scheduled refresh is cancelled (no late provider invocation after clock advances past fire time).
3. Armed connectivity retry is cancelled (reachability restoration triggers no retry).
4. A caller dedup'd onto an in-flight fetch receives a failure, not a token (single provider invocation — dedup preserved).
5. Re-registration after unregister fetches via the new provider and schedules a fresh refresh.

Run:
```
  -destination "platform=iOS Simulator,name=iPhone 17 Pro" \
  -only-testing:KlaviyoCoreTests/AuthTokenManagerRefreshTests \
  -only-testing:KlaviyoCoreTests/AuthTokenManagerTests
```

## Related Issues/Tickets
- [MAGE-685](https://linear.app/klaviyo/issue/MAGE-685/ios-public-unregisterauthtokenprovider-api)
- Parent: MAGE-615 · Sibling: MAGE-623 (register API) · Builds on `clearTokenState()` (MAGE-626)
- Stacked on `ab/MAGE-735/foreground-retry-failed-scheduled-refresh` (this PR targets that branch).

## Summary by CodeRabbit

* **New Features**
  * New public method to unregister authentication token providers. This clears cached tokens, cancels any in-flight token requests, and stops scheduled token refreshes. Personalized in-app forms will lose access to token-based content until a new provider is registered. Ideal for logout flows and switching between user accounts.

### DIFF
--- a/Sources/KlaviyoSwift/Klaviyo+AuthToken.swift
+++ b/Sources/KlaviyoSwift/Klaviyo+AuthToken.swift
@@ -30,4 +30,18 @@ extension KlaviyoSDK {
             await AuthTokenManager.shared.registerProvider(provider)
         }
     }
+
+    /// Detaches a previously registered auth token provider — e.g. on user
+    /// logout.
+    ///
+    /// Clears the provider reference and tears down all associated token state:
+    /// the cached token is discarded and any scheduled proactive refresh or
+    /// in-flight fetch is cancelled. After this call, personalized in-app forms
+    /// have no token available until a new provider is registered via
+    /// ``registerAuthTokenProvider(_:)``.
+    public func unregisterAuthTokenProvider() {
+        Task {
+            await AuthTokenManager.shared.unregisterProvider()
+        }
+    }
 }
--- a/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
+++ b/Sources/KlaviyoCore/Auth/AuthTokenManager.swift
@@ -212,6 +212,24 @@ package actor AuthTokenManager {
         }
     }
 
+    /// Detaches the registered provider and tears down all associated token
+    /// state. Called from `KlaviyoSDK().unregisterAuthTokenProvider()` — e.g. on
+    /// user logout, where the host expects the full token lifecycle to end.
+    ///
+    /// Clears the provider reference *before* ``clearTokenState()`` so any
+    /// callbacks resolving on the cancelled tasks find no provider to call back
+    /// into. Unlike ``clearTokenState()`` — which retains the provider so the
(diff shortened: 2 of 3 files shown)


Sources/KlaviyoSwift/Klaviyo+AuthToken.swift
```diff
@@ -30,4 +30,18 @@ extension KlaviyoSDK {
             await AuthTokenManager.shared.registerProvider(provider)
         }
     }
+
+    /// Detaches a previously registered auth token provider — e.g. on user
+    /// logout.
+    ///
+    /// Clears the provider reference and tears down all associated token state:
+    /// the cached token is discarded and any scheduled proactive refresh or
```
ajaysubra, line 38: cached?

ajaysubra (approved):
LGTM just one quick question.

Sources/KlaviyoSwift/Klaviyo+AuthToken.swift
ab1470: Cached meaning just the JWT that's in memory. The AuthTokenManager keeps a fresh token in ephemeral memory so we're basically always ready to pass one into an IAF web view when needed. This keeps the token management independent of the IAF Web View's lifecycle

evan-masseau (approved):

PR #623, opened 2026-06-25
feat(PUSH-834): loosen open_url scheme gate to allowlist

morrilltim:
## Summary
Loosen the open_url action's http/https gate to an allowlist supporting mailto:, tel:, sms: in addition to web schemes. No dispatch changes needed (UIApplication.shared.open is already scheme-agnostic).

Note: smsto: is intentionally NOT on the iOS allowlist (it is on Android's). iOS Messages registers sms: but has no smsto: handler, so smsto: would no-op; it is rejected at the gate instead.

## Changes
- Loosen three input gates (DeepLinkHandler, UNNotificationResponse+Klaviyo, KlaviyoActionButtonParser) to accept allowlist
- Add shared URLSchemeAllowlist constant (package visibility in KlaviyoCore; internal mirror in KlaviyoSwiftExtension due to module dependencies); a parity test keeps the two copies in sync
- Allowlist: http, https, mailto, tel, sms (smsto excluded, Android-only)
- Block intent:, android-app:, javascript:, file:, content:, data: (scheme validation)
- On blocked scheme: drop silently (existing nil-return fall-through behavior)

## Test Plan
- Full test suite passes; new/updated tests: allowlisted schemes accepted, blocked schemes rejected, .openWebUrl dispatched for mailto:/tel:, smsto: rejected, KlaviyoCore/KlaviyoSwiftExtension allowlist parity

### Manual testing (real device)
Verified end to end by sending push notifications through the production web interface and tapping on a physical device:

| web_url value | Result |
|---|---|
| `tel:+18005551234` | Opens Phone (call prompt) ✅ |
| `mailto:support@example.com?subject=Hi&body=Hello` | Opens Mail composer (To/subject/body pre-filled) ✅ |
| `sms:+18005551234?body=Test%20message` | Opens Messages (body pre-filled) ✅ |
| `https://example.com` | Opens Safari ✅ |
| `smsto:+18005551234` | Dropped at gate + logged; no dead-end app open ✅ (Android-only scheme) |
| `javascript:` / `file://` / `intent://` / `geo:` | Dropped at gate, no crash ✅ |

`_openedPush` tracking fires on the allowed cases. Reject path logs `web_url '...' has a scheme not in the allowed list; ignoring.`

The smsto: case is what surfaced this: on iOS it originally passed the gate but then no-oped in UIApplication.shared.open, leaving the user in the app. Dropping it from the iOS allowlist makes the rejection explicit and logged.

## Related
Part of PUSH-834 (multi-repo). Also see Android and fender PRs, plus app backend PR klaviyo/app#123425 (loosens the server-side web_url validation that would otherwise reject these schemes at save time).

https://linear.app/klaviyo/issue/PUSH-834

## Summary by CodeRabbit

* **New Features**
  * Expanded supported `open_url` and notification `web_url` handling to include allowlisted non-web schemes: `mailto:`, `tel:`, and `sms:`.

* **Bug Fixes**
  * Added URL scheme validation for `open_url` action buttons and notification `web_url`, silently ignoring blocked schemes (e.g., `javascript:`, `intent:`, `file:`) while preserving tracking.
  * On iOS 14+, rejected schemes now log a warning.

* **Tests**
  * Expanded test coverage for allowlisted and blocked schemes, including allowlist parity checks between the two SDK components.

### DIFF
--- a/Tests/KlaviyoSwiftTests/DeepLinkHandlingTests.swift
+++ b/Tests/KlaviyoSwiftTests/DeepLinkHandlingTests.swift
@@ -171,6 +171,110 @@ final class DeepLinkHandlingTests: XCTestCase {
         klaviyoSwiftEnvironment.send = originalSend
     }
 
+    // MARK: - open_url with allowlisted non-web schemes (PUSH-834)
+
+    @MainActor
+    func testHandleNotificationResponseDispatchesOpenWebUrlForMailtoWebUrl() async throws {
+        let urlString = "mailto:support@example.com"
+        let expectedURL = try XCTUnwrap(URL(string: urlString))
+        let userInfo: [AnyHashable: Any] = [
+            "body": ["_k": "1"],
+            "web_url": urlString
+        ]
+        let response = try UNNotificationResponse.with(userInfo: userInfo)
+
+        environment.linkHandler.unregisterCustomHandler()
+
+        let completionCalled = expectation(description: "completion handler called")
+        let actionReceived = expectation(description: "openWebUrl action received for mailto:")
+        let originalSend = klaviyoSwiftEnvironment.send
+        klaviyoSwiftEnvironment.send = { action in
+            if case let .openWebUrl(url) = action {
+                XCTAssertEqual(url, expectedURL, "Should dispatch openWebUrl with mailto: URL")
+                actionReceived.fulfill()
+                return nil
+            }
+            return originalSend(action)
+        }
+
+        let sdk = KlaviyoSDK()
+        let result = sdk.handle(notificationResponse: response, withCompletionHandler: {
+            completionCalled.fulfill()
+        })
+
+        XCTAssertTrue(result, "SDK should return true for Klaviyo notifications with web_url")
+        await fulfillment(of: [completionCalled, actionReceived], timeout: 1.0)
+
+        klaviyoSwiftEnvironment.send = originalSend
+    }
+
+    @MainActor
+    func testHandleNotificationResponseDispatchesOpenWebUrlForTelWebUrl() async throws {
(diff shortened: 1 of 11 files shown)


josephzhou-klaviyo (approved):
Approving but frankly i have negligible swift background

jasper-hsu-kyvo (approved):

Tests/KlaviyoSwiftTests/DeepLinkHandlingTests.swift
```diff
@@ -171,6 +171,110 @@ final class DeepLinkHandlingTests: XCTestCase {
...
+            if case .openWebUrl = action {
+                openWebUrlDispatched = true
+            }
+            return originalSend(action)
+        }
+
+        _ = sdk.handle(notificationResponse: response, withCompletionHandler: {})
+
+        // Give time for any async dispatch to land
+        try await Task.sleep(nanoseconds: 100_000_000)
```
glenn-klaviyo: is there any way to avoid the `Task.sleep` here? I'm personally a little hesitant about `sleeps` in tests as they can be flaky. Okay if not since I see this has been done in the past but wanted to raise it.

belleklaviyo (approved):
neatoooo looks good to me and looks well tested

glenn-klaviyo (approved):
LGTM. wish we didn't have to duplicate the url schemes but I understand with the current set up.

Tests/KlaviyoSwiftTests/DeepLinkHandlingTests.swift
morrilltim: Yea, I agree with trying to not use a sleep. I'll try and rework these tests.

PR #626, opened 2026-06-30
feat(KlaviyoCore): retry on greater of Retry-After and exponential backoff (MAGE-500)

evan-masseau:
## What
Implements MAGE-500 proposal #2: for retryable responses (429 + 5xx), wait the GREATER of the SDK's exponential backoff and the server's `Retry-After` header.

## Why
The retryable-status handler in `KlaviyoAPI` previously overwrote the exponential backoff with the `Retry-After` value whenever it parsed. A request deep in a rate-limit storm could therefore retry too soon just because the server's rate-limit window reset to a short `Retry-After`.

## Change
- `nextBackoff = max(exponentialBackOff, retryAfterSeconds)` when the `Retry-After` header parses.
- Falls back to exponential backoff when the header is missing or unparseable.
- Jitter behavior unchanged.

## Tests
Added 4 `KlaviyoAPITests` cases: Retry-After wins, backoff wins, 5xx greater-of, and no-header fallback.

## CI
Verified green via `workflow_dispatch` on the branch — all 6 matrix jobs (Xcode 15.2/15.4/16.4 × debug+release) passed.

## Note
Per repo policy, `feat/*` cannot target `master` directly — please retarget this PR to the active `rel/*` branch.

## Scope
Greater-of change only; other MAGE-500 proposals left as follow-ups.

🔗 [View this job in Reca Studio](https://kiosk.klaviyo-dev.com/remote-coding-agent/flow/546ee672-2202-4b28-a0b8-93a2b7f0f3b0)
🤖 Generated with Reca

Reca job ID: 546ee672-2202-4b28-a0b8-93a2b7f0f3b0

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry/backoff behavior for retryable HTTP responses by capping exponential backoff and prioritizing the larger value of server `Retry-After` vs capped exponential backoff; falls back cleanly when `Retry-After` is missing or invalid.
  * Updated retry limits to be endpoint-specific for more appropriate retry behavior across different operations.
* **Tests**
  * Added async test coverage for `Retry-After` parsing, backoff selection, precedence rules, and the exponential backoff cap.

### DIFF
--- a/Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
+++ b/Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
@@ -19,13 +19,22 @@ public enum HTTPStatusCode {
     public static let retryableServerErrorRange = 500...599
 }
 
+@usableFromInline
 enum RetryBackoffConstants {
     /// Ceiling on the SDK's exponential backoff interval, in seconds (5 minutes).
     ///
     /// Bounds our own backoff so it can't grow unbounded across a long rate-limit storm. Aligns with
     /// comparable SDKs (Segment caps at 300s) per MAGE-500. A server-provided `Retry-After` may still
     /// exceed this ceiling — only the SDK-computed backoff is capped.
+    @usableFromInline
     static let maxBackoffSeconds = 300
+
+    /// The response header the server uses to request a specific retry delay.
+    ///
+    /// Klaviyo's API sends this as delay-seconds (a non-negative integer); the HTTP-date form is not
+    /// emitted, so a non-integer value falls back to exponential backoff.
+    @usableFromInline
+    static let retryAfterHeader = "Retry-After"
 }
 
 public struct KlaviyoAPI {
@@ -82,9 +91,15 @@ public struct KlaviyoAPI {
             // (Retry-After expected for 429, future-proofing for 5xx). Taking the greater of the two
             // keeps a request deep in a rate-limit storm backing off rather than retrying too soon
             // just because the server's rate-limit window reset to a short Retry-After.
+            //
+            // Klaviyo sends Retry-After as delay-seconds; the HTTP-date form is not used, so a
+            // non-integer value falls back to exponential backoff. value(forHTTPHeaderField:) is
+            // case-insensitive, so HTTP/2 lowercase header names are handled.
             var nextBackoff: Int = exponentialBackOff
-            if let retryAfter = httpResponse.value(forHTTPHeaderField: "Retry-After"),
-               let retryAfterSeconds = Int(retryAfter) {
+            let retryAfterValue = httpResponse.value(
+                forHTTPHeaderField: RetryBackoffConstants.retryAfterHeader
+            )
+            if let retryAfterValue, let retryAfterSeconds = Int(retryAfterValue) {
                 nextBackoff = max(exponentialBackOff, retryAfterSeconds)
             }
             let jitter = environment.randomInt()
--- a/Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
+++ b/Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
@@ -19,22 +19,21 @@ public enum HTTPStatusCode {
     public static let retryableServerErrorRange = 500...599
 }
 
-@usableFromInline
-enum RetryBackoffConstants {
+/// Public because it is referenced from `KlaviyoAPI.init`'s public default-argument closure;
+/// Swift 5.9 (Xcode 15.2/15.4) requires such symbols to be `public` (not just `@usableFromInline`).
+public enum RetryBackoffConstants {
     /// Ceiling on the SDK's exponential backoff interval, in seconds (5 minutes).
     ///
     /// Bounds our own backoff so it can't grow unbounded across a long rate-limit storm. Aligns with
     /// comparable SDKs (Segment caps at 300s) per MAGE-500. A server-provided `Retry-After` may still
     /// exceed this ceiling — only the SDK-computed backoff is capped.
-    @usableFromInline
-    static let maxBackoffSeconds = 300
+    public static let maxBackoffSeconds = 300
 
     /// The response header the server uses to request a specific retry delay.
     ///
     /// Klaviyo's API sends this as delay-seconds (a non-negative integer); the HTTP-date form is not
     /// emitted, so a non-integer value falls back to exponential backoff.
-    @usableFromInline
-    static let retryAfterHeader = "Retry-After"
+    public static let retryAfterHeader = "Retry-After"
 }
 
 public struct KlaviyoAPI {
--- a/Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
+++ b/Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
@@ -19,6 +19,15 @@ public enum HTTPStatusCode {
     public static let retryableServerErrorRange = 500...599
 }
 
+enum RetryBackoffConstants {
+    /// Ceiling on the SDK's exponential backoff interval, in seconds (5 minutes).
+    ///
+    /// Bounds our own backoff so it can't grow unbounded across a long rate-limit storm. Aligns with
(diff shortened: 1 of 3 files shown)


sre-tokens:
bugbot run

Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
```diff
@@ -43,10 +43,14 @@ public struct KlaviyoAPI {
...
+            // (Retry-After expected for 429, future-proofing for 5xx). Taking the greater of the two
+            // keeps a request deep in a rate-limit storm backing off rather than retrying too soon
+            // just because the server's rate-limit window reset to a short Retry-After.
             var nextBackoff: Int = exponentialBackOff
-            // Check Retry-After header for any retryable error (expected for 429, future-proofing for 5xx)
-            if let retryAfter = httpResponse.value(forHTTPHeaderField: "Retry-After") {
-                nextBackoff = Int(retryAfter) ?? exponentialBackOff
+            if let retryAfter = httpResponse.value(forHTTPHeaderField: "Retry-After"),
+               let retryAfterSeconds = Int(retryAfter) {
+                nextBackoff = max(exponentialBackOff, retryAfterSeconds)
```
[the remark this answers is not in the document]
evan-masseau: Retry after header is only expected to be an int in seconds, not a date. This is over engineering

Tests/KlaviyoCoreTests/KlaviyoAPITests.swift
```diff
@@ -134,6 +134,88 @@ final class KlaviyoAPITests: XCTestCase {
...
+        if let retryAfter {
+            headers["Retry-After"] = retryAfter
+        }
+        return HTTPURLResponse(
+            url: TEST_URL,
+            statusCode: statusCode,
+            httpVersion: nil,
+            headerFields: headers
+        )!
+    }
```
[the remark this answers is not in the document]
evan-masseau: Generally true, but not for test code

Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
```diff
@@ -8,6 +8,15 @@
 import AnyCodable
 import Foundation
 
+enum RetryBackoffConstants {
+    /// Ceiling on the SDK's exponential backoff interval, in seconds (5 minutes).
+    ///
+    /// Bounds our own backoff so it can't grow unbounded across a long rate-limit storm. Aligns with
+    /// comparable SDKs (Segment caps at 300s) per MAGE-500. A server-provided `Retry-After` may still
```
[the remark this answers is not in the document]
evan-masseau: Good catch, ticket references don't belong in source code, even comments

sre-tokens:
bugbot run

Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
```diff
@@ -66,11 +83,26 @@ public struct KlaviyoAPI {
...
             && code != HTTPStatusCode.httpVersionNotSupported
         if code == HTTPStatusCode.rateLimited || isRetryableServerError {
-            let exponentialBackOff = Int(pow(2.0, Double(requestAttemptInfo.attemptNumber)))
+            // Cap our exponential backoff at the max retry interval so it can't grow unbounded
+            // across a long rate-limit storm. A server-provided Retry-After may still exceed this.
+            let exponentialBackOff = min(
+                Int(pow(2.0, Double(requestAttemptInfo.attemptNumber))),
+                RetryBackoffConstants.maxBackoffSeconds
+            )
+            // Wait the GREATER of the server-provided Retry-After and our exponential backoff
```
ajaysubra, line 87: Do we know how this impacts time requests spend in queue? The solution solves for when there is an active rate limiting of endpoint going but once say we bump up the rate limit and we are in the situation where at a company level certain calls from various devices are getting through and certain aren't does this increase the time a request lives in the SDK queue? 

Curious to hear what you think?

ajaysubra (commented):
LGTM, had a question about queue time of requests.

morrilltim (approved):

Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
```diff
@@ -21,6 +21,23 @@ public enum HTTPStatusCode {
     public static let retryableServerErrorRange = 500...599
 }
 
+/// Public because it is referenced from `KlaviyoAPI.init`'s public default-argument closure;
+/// Swift 5.9 (Xcode 15.2/15.4) requires such symbols to be `public` (not just `@usableFromInline`).
```
ab1470:
```suggestion
```
meh, not a helpful comment. I don't think it would be useful even as a code comment (`//`), but definitely not helpful as a documentation comment (`///`)

ab1470 (approved):

Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
evan-masseau: Done in `7499658` — dropped the redundant "why public" comment on RetryBackoffConstants. The same public-default-argument rationale already lives on the HTTPStatusCode enum directly above, so nothing is lost. -Claude

Tests/KlaviyoCoreTests/KlaviyoAPITests.swift
```diff
@@ -166,6 +166,140 @@ final class KlaviyoAPITests: XCTestCase {
...
+        let attemptInfo = try XCTUnwrap(RequestAttemptInfo(attemptNumber: 3, maxAttempts: 50))
+
+        let result = await KlaviyoAPI().send(request, attemptInfo)
+
+        guard case let .failure(.rateLimitError(backOff)) = result else {
+            XCTFail("Expected rateLimitError, got \(result)")
+            return
+        }
+        XCTAssertEqual(backOff, 8)
+    }
```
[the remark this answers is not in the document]
evan-masseau: Leaving the six explicit tests as-is for now: the table-driven version trades the per-scenario comments (429 vs 5xx, header-wins vs exp-wins, cap, unparseable) for tuple labels plus a shared extraction of the .rateLimitError/.serverError backOff, which reads less clearly for a modest line saving and risks a refactor bug in currently-passing tests. Leaving open in case you would prefer the consolidation. -Claude

Sources/KlaviyoCore/Networking/KlaviyoAPI.swift
evan-masseau: Good question — this deliberately trades a little per-device latency for meaningfully less fleet-wide load.

**Per device:** yes, a request can dwell a bit longer. Before, when a `Retry-After` was present we retried at ~that value, flat (no escalation). Now we take `max(Retry-After, exponentialBackoff)`, so after ~6 attempts the backoff overtakes and retries space out, capping at one retry / 5 min — worst case a request waits up to ~5 min longer to land after the limit lifts. No data cost, though: events are timestamped at capture, so this is latency-to-server, not lost or misdated data.

**At fleet scale (the win):** and it is biggest exactly where it hurts most. The pathological case is `Retry-After=0` (burst limit): today every device retries every few seconds, sustained, which keeps the account pinned at its limit and inflates the 429 count. `max(exp, 0) = exp`, so that storm decays to ~1 retry / 5 min per device instead of ~12/min. Per-device jitter also desyncs the fleet so retries stop arriving in synchronized waves. Net: far fewer retries hitting the server, and the underlying limit recovers faster.

**Is this a standard pattern?** Yes — we are not inventing anything. RFC 9110 defines `Retry-After` as a *minimum* wait (retry no sooner; may wait longer), so `max(backoff, Retry-After)` is spec-aligned and strictly server-safe — it can only add delay, never retry sooner than the server asked. Exponential backoff + jitter while honoring `Retry-After` (usually via `max`) is the norm in the AWS/GCP/Azure SDKs and Segment.

-Claude

evan-masseau: Well, we already had jitter, so claude is inventing the “arriving in waves” thing. But otherwise yes, Claude conveyed what I was trying to get at with this PR.

PR #627, opened 2026-07-01
fix(state): evict oldest queued request at capacity (MAGE-496)

evan-masseau:
Part of the **SDK Network Behavior Upgrades** stacked chain. **Base: `feat/mage-694-retry-all-5xx`.**

MAGE-496 — align queue-depth behavior across the Android & iOS SDKs.

## Change
`KlaviyoState.enqueueRequest` now evicts the **oldest** queued request at capacity (`removeFirst()` + developer warning) instead of silently dropping the newest. Also corrects an off-by-one: the old `queue.count + 1 < maxQueueSize` guard effectively capped at 199; it now holds a true 200.

## Tests
Added `KlaviyoStateTests` for evict-oldest-at-capacity and append-below-capacity. Verified locally: full suite green on Xcode 26 (0 failures), lint clean on pinned toolchain.

🔗 Chain: retry-all-5xx (#619) → **align** → backoff (#626)
🤖 Generated with Reca

## Summary by CodeRabbit

* **New Features**
  * Queue items now carry an “enqueued at” timestamp.
  * Added support for priority queueing, placing priority events at the front.
  * When the queue reaches its maximum size, new items are accepted by evicting the oldest entry based on the enqueued timestamp.
  * Backward compatibility: older saved queue items without the timestamp get a safe fallback value.
* **Bug Fixes**
  * Priority events now follow the same bounded-queue and timestamp-based eviction rules as normal events.
* **Tests**
  * Expanded coverage for queue capacity, eviction ordering, and timestamp serialization behavior.


Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -73,8 +73,11 @@ struct KlaviyoState: Equatable, Codable {
     }
 
     mutating func enqueueRequest(request: KlaviyoRequest) {
-        guard queue.count + 1 < StateManagementConstants.maxQueueSize else {
-            return
+        if queue.count >= StateManagementConstants.maxQueueSize {
+            let maxSize = StateManagementConstants.maxQueueSize
+            environment.emitDeveloperWarning(
+                "Request queue at capacity (\(maxSize)); evicting oldest request to make room.")
+            queue.removeFirst()
```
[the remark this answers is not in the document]
evan-masseau: yeah.. thats a problem. My next idea would be to include an "enqueue time" on requests, and evict by actual enqueue date/time, not just place in queue.

Sources/KlaviyoCore/Networking/KlaviyoRequest.swift
```diff
@@ -18,17 +18,56 @@ public struct KlaviyoRequest: Identifiable, Equatable, Codable {
     /// The API endpoint this request targets.
     public let endpoint: KlaviyoEndpoint
 
+    /// The time at which this request was added to the send queue.
+    ///
+    /// Used to evict the oldest request when the queue reaches capacity. This mirrors
+    /// the Android SDK's `queuedTime`. Defaults to the current time when the request
+    /// is created.
```
evan-masseau:
```suggestion
    /// The time at which this request was added to the send queue.
    /// Used to evict the oldest request when the queue reaches capacity.
    /// Defaults to the current time when the request is created.
```

Sources/KlaviyoCore/Networking/KlaviyoRequest.swift
```diff
@@ -18,17 +18,56 @@ public struct KlaviyoRequest: Identifiable, Equatable, Codable {
...
+        case enqueuedAt
+    }
+
+    /// Custom decoding that stays backward compatible with queues persisted before
+    /// `enqueuedAt` existed (e.g. requests carried across an app upgrade). A missing
+    /// timestamp defaults to `Date.distantPast` so these legacy requests sort as the
+    /// oldest and are evicted first under overflow. Because `min(by:)` returns the first
+    /// minimal element, legacy entries tie-break by queue position (front-most first),
+    /// which preserves their original age order.
+    public init(from decoder: Decoder) throws {
```
evan-masseau:
```suggestion
    /// Custom decoding for backward compatibility with queues persisted before
    /// `enqueuedAt` existed (e.g. requests carried across an app upgrade). 
    ///
    /// A missing timestamp defaults to `Date.distantPast` so legacy requests sort as the
    /// oldest and are evicted first under overflow. 
    public init(from decoder: Decoder) throws {
```

Sources/KlaviyoCore/Networking/KlaviyoRequest.swift
```diff
@@ -18,17 +18,56 @@ public struct KlaviyoRequest: Identifiable, Equatable, Codable {
...
+        enqueuedAt = try container.decodeIfPresent(Date.self, forKey: .enqueuedAt) ?? .distantPast
+    }
+
+    /// Equality intentionally ignores `enqueuedAt`, keeping dedup and TCA state-diffing keyed on
+    /// `id` + `endpoint` exactly as before this field existed. This is safe rather than a footgun
+    /// for value-equality consumers: `id` is a unique per-request identifier and `enqueuedAt` is
+    /// set once at creation and never mutated, so any two requests that compare equal already
+    /// carry the same `enqueuedAt` — there is no reachable state where excluding it hides a real
+    /// difference (e.g. a diff that would otherwise trigger a view update).
+    public static func ==(lhs: KlaviyoRequest, rhs: KlaviyoRequest) -> Bool {
```
evan-masseau, line 65:
```suggestion
    /// Equality intentionally ignores `enqueuedAt`, keeping dedup and TCA state-diffing keyed on
    /// `id` + `endpoint` exactly as before this field existed. This is safe rather than a footgun
    /// for value-equality consumers: `id` is a unique per-request identifier and `enqueuedAt` is
    /// set once at creation and never mutated, so any two requests that compare equal already
    /// carry the same `enqueuedAt`
    public static func ==(lhs: KlaviyoRequest, rhs: KlaviyoRequest) -> Bool {
```

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -73,12 +73,39 @@ struct KlaviyoState: Equatable, Codable {
...
-            return
-        }
+        evictOldestIfAtCapacity()
         queue.append(request)
     }
 
+    /// Enqueues a high-priority request at the front of the queue (e.g. opened-push or
+    /// geofence events that should flush immediately), enforcing the same capacity cap as
+    /// ``enqueueRequest(request:)``. The oldest request is evicted first (when full) so this
+    /// prioritized request — which carries the newest timestamp — is never the one dropped.
```
evan-masseau:
```suggestion
    /// Enqueues a high-priority request at the front of the queue (e.g. opened-push or
    /// geofence events that should flush immediately), enforcing the same capacity cap as
    /// ``enqueueRequest(request:)``. 
```

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -73,12 +73,39 @@ struct KlaviyoState: Equatable, Codable {
...
+    mutating func enqueuePriorityRequest(request: KlaviyoRequest) {
+        evictOldestIfAtCapacity()
+        queue.insert(request, at: 0)
+    }
+
+    /// Evicts the oldest queued request (by `enqueuedAt`) when the queue is at or above
+    /// capacity, making room for one more. Call this *before* adding a new request so the
+    /// request being added is never the one evicted. Prioritized events are inserted at the
+    /// front but carry the newest timestamp, so they are never selected as the oldest.
```
evan-masseau:
```suggestion
    /// Evicts the oldest queued request (by `enqueuedAt`) when the queue is at or above
    /// capacity, making room for one more.
```

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -73,12 +73,39 @@ struct KlaviyoState: Equatable, Codable {
...
+    mutating func evictOldestIfAtCapacity() {
```
evan-masseau: Pretty sure this can be private

```suggestion
    private mutating func evictOldestIfAtCapacity() {
```

jasper-hsu-kyvo (approved):

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -73,12 +73,36 @@ struct KlaviyoState: Equatable, Codable {
...
+            "Request queue at capacity (\(maxSize)); evicting oldest request to make room.")
+        // Evict the oldest by wall-clock enqueue time. `min(by:)` returns the first minimal
+        // element, so ties — and legacy `.distantPast` entries carried across an app upgrade —
+        // break by front-most queue position (age order). A backwards clock correction (e.g. an
+        // NTP step) could momentarily mis-order two closely-spaced timestamps, but that only
+        // changes *which* near-oldest request is dropped at overflow; we accept wall-clock here
+        // rather than thread a monotonic sequence counter through every enqueue path.
+        if let oldestIndex = queue.indices.min(by: { queue[$0].enqueuedAt < queue[$1].enqueuedAt }) {
+            queue.remove(at: oldestIndex)
+        }
```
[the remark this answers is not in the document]
evan-masseau: Fixed in `52ad9fa1` — `evictOldestIfAtCapacity()` now loops (`while queue.count >= maxSize`) instead of removing a single entry, so a queue that is already over the bound (init-time queue merging or in-flight reinsertion) drains all the way to `maxSize - 1` and the caller's append/insert lands it at exactly the cap. Added a regression test that seeds the queue above capacity and asserts a single enqueue drains it back down. -Claude

Tests/KlaviyoSwiftTests/KlaviyoStateTests.swift
```diff
@@ -193,4 +193,265 @@ final class KlaviyoStateTests: XCTestCase {
...
+        }
+        var state = KlaviyoState(apiKey: TEST_API_KEY, anonymousId: "anon", queue: requests)
+
+        // Act: one enqueue.
+        let newRequest = makeTokenRequest(id: "new", enqueuedAt: base.addingTimeInterval(999_999))
+        state.enqueueRequest(request: newRequest)
+
+        // Assert: drained to exactly the cap in a single call, dropping the oldest requests
+        // (req-0 through req-25) and keeping the rest plus the newcomer at the tail.
+        XCTAssertEqual(state.queue.count, maxSize, "A single enqueue must drain an over-capacity queue back to the cap")
```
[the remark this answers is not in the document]
evan-masseau: Fixed in `1260d8fa` — the assertion is now wrapped across multiple lines (under the 110-char limit). It was folded into the exact-contents comparison below, so the long single-line form is gone.

Tests/KlaviyoSwiftTests/KlaviyoStateTests.swift
```diff
@@ -193,4 +193,265 @@ final class KlaviyoStateTests: XCTestCase {
...
+        XCTAssertFalse(state.queue.contains { $0.id == "req-0" }, "Oldest requests must be evicted first")
+        XCTAssertFalse(state.queue.contains { $0.id == "req-25" }, "Eviction must drain enough of the oldest to reach the cap")
+        XCTAssertTrue(state.queue.contains { $0.id == "req-26" }, "Requests newer than the drained window must survive")
+        XCTAssertEqual(state.queue.last?.id, "new", "New request must be appended at the tail")
```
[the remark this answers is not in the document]
evan-masseau: Fixed in `1260d8fa` — replaced the spot-checks with an exact-contents comparison: `XCTAssertEqual(state.queue.map(\.id), (26..<overCapacity).map { "req-\($0)" } + ["new"])`. That proves the full eviction window (req-0…req-25 evicted, the rest surviving in order, newcomer at the tail) rather than allowing arbitrary removals to pass. -Claude

glenn-klaviyo (approved):
LGTM

ab1470 (approved):

belleklaviyo (approved):

PR #633, opened 2026-07-07
feat(forms): handle IAF openExternalUrl bridge event (PUSH-732)

morrilltim:
## Summary

Handles the IAF `openExternalUrl` native bridge event in the iOS SDK so that in-app form CTAs configured with external/web URLs open in Safari (bypassing any registered deep link handler) rather than routing through the deep link path. Adds a new public `formExternalUrlClicked` lifecycle event so customers can differentiate external-browser CTAs from in-app deep-link CTAs in their analytics integrations.

Changes:
- `IAFNativeBridgeEvent`: new `openExternalUrl` case (decode, handshake manifest v1). Decodes via a dedicated `ExternalUrlEventPayload` that reads the flat `url` field onsite emits — **not** the platform-split `ios`/`android` shape used by `openDeepLink` (see Manual QA below)
- `FormLifecycleEvent`: new public `formExternalUrlClicked(formId:formName:buttonLabel:url:)` case
- `IAFWebViewModel`: handles the event — guards nil/empty URL, validates scheme against `openUrlAllowedSchemes`, opens via `DeepLinkHandler.openExternalURL`, fires the lifecycle event when form metadata is present; extracted a shared `invokeCtaLifecycleHandler` helper used by both CTA paths
- `KlaviyoSDK+Forms`: doc comment + example updated for the new case
- Example app: switch updated for the new case

Note: `formExternalUrlClicked` is a source-breaking addition to the public `FormLifecycleEvent` enum — consumers with exhaustive switches need a new case on upgrade. Worth a changelog callout.

## Test Plan

- Unit tests: `xcodebuild test -only-testing:KlaviyoFormsTests` — all passing
  - Decoding: new event with/without buttonLabel, without form context, empty-URL normalization — fixtures assert the real onsite `url` payload
  - Handshake: both hardcoded expectations updated to include `openExternalUrl` v1
  - Handler: lifecycle fires with metadata; skipped without metadata, with nil URL, and with a blocked scheme (`javascript://`)
  - Lifecycle: handler invocation + computed properties for the new case

- **Manual QA (completed ✅)** — on-device via local Fender:
  - **Setup:** iOS Simulator (iPhone 16 Pro, Xcode 26.5); example `SPMExample` app pointed at a local Fender **master** onsite dev server (`assetSource: development`); fender master onsite is the source of the `openExternalUrl` bridge event (`openWebUrlAction.ts`)
  - **Form:** "Tim Test Open Web URL" with an external-URL CTA
  - **Result:** tapping the external-URL CTA **opened the correct webpage in Safari**, bypassing the deep-link handler ✅
  - **Bug found & fixed during QA:** the decode originally reused `DeepLinkEventPayload` (reads the `ios` key), but onsite emits the web URL as a flat `url` field, so the URL always decoded to `nil` — the CTA guard tripped and the form dismissed without opening Safari. Fixed by adding `ExternalUrlEventPayload` (reads `url`) and correcting the test fixtures/helper, which had encoded the same wrong key (so unit tests were passing vacuously).

> CI note: the repo's `swift.yml` test job runs on `pull_request` with a `branches: ['*']` filter, which does not match this PR's slashed base branch (`feat/push-638-open-url`), so the unit-test workflow does not execute in CI for this stacked PR. Tests were verified locally (`KlaviyoFormsTests`, all green) and will run in CI once the stack lands on `master`.

## Acceptance Criteria

| ✅ Human | 🤖 AI | Criterion |
|---------|-------|-----------|
| - [x] | ✅ | IAF form CTA that emits `openExternalUrl` opens the URL in Safari — **verified on-device** (opened correct webpage; also unit-tested) |
| - [x] | ✅ | `formExternalUrlClicked` lifecycle event fires with `formId`, `formName`, `buttonLabel`, `url` when metadata is present (unit-tested against the real `url` payload) |
| - [ ] | ✅ | Existing `openDeepLink` IAF behavior is unchanged (unit-tested; not separately re-run on-device) |
| - [x] | ✅ | Unit tests: `IAFNativeBridgeEventTests` (decode of new event), `FormLifecycleHandlerTests` (lifecycle event invocation), `IAFWebViewModelTests` (handler + handshake) |
| - [x] | ✅ | Manual QA via Fender form configured with an external URL CTA |

[PUSH-732](https://linear.app/klaviyo/issue/PUSH-732/ios-sdk-handle-iaf-openexternalurl-message-bus-event)

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle tracking for in-app form CTAs that open external URLs, including the form, button label, and destination URL.
  * Added support for validating allowed URL schemes before opening external links.
  * Added external URL CTA handling to the in-app form event bridge.

* **Documentation**
  * Updated lifecycle handler guidance with deep-link and external URL CTA examples.

* **Tests**
  * Added coverage for external URL events, decoding, validation, and lifecycle callbacks.

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/Models/FormLifecycleEvent.swift
+++ b/Sources/KlaviyoForms/InAppForms/Models/FormLifecycleEvent.swift
@@ -56,12 +56,23 @@ public enum FormLifecycleEvent: Equatable, Sendable {
     /// - `deepLinkUrl`: The deep link URL associated with the CTA.
     case formCtaClicked(formId: String, formName: String, buttonLabel: String, deepLinkUrl: URL)
 
+    /// Triggered when a user taps a call-to-action (CTA) button in a form
+    /// that has an external (web) URL configured.
+    ///
+    /// Fired after the SDK has initiated opening the URL in Safari. Not emitted
+    /// if no external URL is configured for the CTA.
+    ///
+    /// - `buttonLabel`: The label text of the tapped button.
+    /// - `url`: The external URL associated with the CTA.
+    case formExternalUrlClicked(formId: String, formName: String, buttonLabel: String, url: URL)
+
     /// The unique identifier of the form that triggered this event.
     public var formId: String {
         switch self {
         case let .formShown(formId, _),
              let .formDismissed(formId, _),
-             let .formCtaClicked(formId, _, _, _):
+             let .formCtaClicked(formId, _, _, _),
+             let .formExternalUrlClicked(formId, _, _, _):
             return formId
         }
     }
@@ -71,7 +82,8 @@ public enum FormLifecycleEvent: Equatable, Sendable {
         switch self {
         case let .formShown(_, formName),
              let .formDismissed(_, formName),
-             let .formCtaClicked(_, formName, _, _):
+             let .formCtaClicked(_, formName, _, _),
+             let .formExternalUrlClicked(_, formName, _, _):
             return formName
         }
     }
@@ -82,6 +94,7 @@ public enum FormLifecycleEvent: Equatable, Sendable {
         case .formShown: return "formShown"
         case .formDismissed: return "formDismissed"
         case .formCtaClicked: return "formCtaClicked"
+        case .formExternalUrlClicked: return "formExternalUrlClicked"
         }
     }
 }
--- a/Tests/KlaviyoFormsTests/IAFWebViewModelTests.swift
+++ b/Tests/KlaviyoFormsTests/IAFWebViewModelTests.swift
@@ -353,13 +353,13 @@ final class IAFWebViewModelTests: XCTestCase {
     // MARK: - openExternalUrl Tests
 
     private func makeOpenExternalUrlMessage(
-        ios: String? = "https://example.com",
+        url: String? = "https://example.com",
         formId: String? = "form123",
         formName: String? = "Newsletter",
         buttonLabel: String? = "Learn More"
     ) -> MockWKScriptMessage {
         var data: [String: String] = [:]
-        data["ios"] = ios
+        data["url"] = url
         data["formId"] = formId
         data["formName"] = formName
         data["buttonLabel"] = buttonLabel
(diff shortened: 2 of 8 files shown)


Tests/KlaviyoFormsTests/IAFWebViewModelTests.swift
```diff
@@ -349,6 +349,97 @@ final class IAFWebViewModelTests: XCTestCase {
...
+        data["ios"] = ios
+        data["formId"] = formId
+        data["formName"] = formName
+        data["buttonLabel"] = buttonLabel
+        let dataJson = data.map { "\"\($0.key)\": \"\($0.value)\"" }.joined(separator: ", ")
+        return MockWKScriptMessage(
+            name: "KlaviyoNativeBridge",
+            body: "{ \"type\": \"openExternalUrl\", \"data\": { \(dataJson) } }"
+        )
+    }
```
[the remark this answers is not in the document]
morrilltim: Good catch — confirmed and fixed in 085d5cc.

`makeOpenExternalUrlMessage` was building the payload with the `ios` key while the decoder reads the flat `url` field (`ExternalUrlEventPayload.url`), so every bridge message carried a nil URL. `testHandleOpenExternalUrlFiresLifecycleEvent` was actually failing, and the metadata / nil-url / disallowed-scheme tests were passing vacuously via the nil-URL early return. Helper now sends `url`, so all four exercise their intended paths. Verified: full `KlaviyoFormsTests` green, and the external-URL CTA confirmed opening in Safari on-device.

Sources/KlaviyoForms/InAppForms/Models/FormLifecycleEvent.swift
```diff
@@ -56,12 +56,23 @@ public enum FormLifecycleEvent: Equatable, Sendable {
...
 
+    /// Triggered when a user taps a call-to-action (CTA) button in a form
+    /// that has an external (web) URL configured.
+    ///
+    /// Fired after the SDK has initiated opening the URL in Safari. Not emitted
+    /// if no external URL is configured for the CTA.
+    ///
+    /// - `buttonLabel`: The label text of the tapped button.
+    /// - `url`: The external URL associated with the CTA.
+    case formExternalUrlClicked(formId: String, formName: String, buttonLabel: String, url: URL)
```
amber-klaviyo, line 67: You mention in the PR description "Note: formExternalUrlClicked is a source-breaking addition to the public FormLifecycleEvent enum — consumers with exhaustive switches need a new case on upgrade. Worth a changelog callout." but I don't see a changelog update. Will there be a different PR for that? should it be included in this PR?

amber-klaviyo (approved):
No blocking comments, looks good!

Sources/KlaviyoForms/InAppForms/Models/FormLifecycleEvent.swift
morrilltim: I think when this change ships in the SDK it should be added to the release notes. I will make sure that gets in there.

PR #639, opened 2026-07-08
refactor: fold reset semantics into Core primitives + decouple Forms teardown (MAGE-834)

belleklaviyo:
# Description
Gives the `KlaviyoCore` singletons a uniform `reset()` and removes the last teardown coupling from `KlaviyoForms` to `KlaviyoInternal`. IAF teardown now cancels only its own observers and touches no global SDK state. Part of Phase 2 (eliminate `KlaviyoInternal`); feeds MAGE-835.

> **Stacked PR:** targets `isobellelim/mage-833-relocate-geofence-dispatch` and auto-retargets to `feat/modularization` as the earlier Phase-2 PRs merge.

**Reframing note:** the ticket's original premise (Forms teardown "clears global identity/config + tears down the mirror") was already stale — MAGE-832 stopped `resetProfileDataSubject` from touching the stores/mirror, and no production code observes the KlaviyoInternal subjects anymore. The four `KlaviyoInternal.reset*` calls in teardown were **vestigial** (they reset private Combine relays nothing in Forms reads). This PR deletes them and formalizes a real Core reset API.

## Due Diligence

## Changelog / Code Overview
**`feat(KlaviyoCore)` — uniform `reset()`:**
- `EventBus.reset()` (renamed from `clearBuffer()`, `package`→`public`) — clears the replay buffer.
- `EventDispatcher.reset()` — `target = nil` under the existing lock (unregister).
- `IdentityStore.reset()` / `SDKConfigStore.reset()` — restore the default (`ProfileData()` / `KlaviyoConfig()`), equal to the existing `update(default)` idiom.
- Repointed `KlaviyoInternal.clearEventBuffer` wrapper + `EventPublishTests` to `reset()`; zero `clearBuffer` references remain.

**`refactor(KlaviyoForms)` — decouple IAF teardown:**
- Removed the 4 vestigial `KlaviyoInternal.reset*` calls from `IAFPresentationManager` (`destroyWebviewAndListeners` ×3, `handleAPIKeyChange` ×1). Existing observer-nil + task-cancel cleanup is self-sufficient — the observers cancel their own EventBus/store subscriptions on deinit.
- New `IAFPresentationManagerTeardownTests` regression guard: identity/config survive teardown.
- Migrated reset-to-default test isolation to `reset()` in `IAFWebViewModelTests`, `KlaviyoWebViewControllerTests`, `KlaviyoLocationManagerTests`.

KlaviyoInternal's subjects/publishers/`reset*Subject`/`fetch*` + `KlaviyoInternalTests` are intentionally left for **MAGE-835** (facade deletion). `EventDispatcher.reset()` is added for API completeness but not wired into any shared teardown.

## Test Plan
- Full suite green: **444 tests, 0 failures** across all 5 bundles (KlaviyoCore 120 / KlaviyoSwift 218 / KlaviyoForms 60 / KlaviyoLocation 37 / KlaviyoSwiftExtension 9).
- New: 4 Core `reset()` unit tests + the teardown regression guard (passes before and after removal).
- Grep gates: no `KlaviyoInternal.reset*Subject` in `Sources`; no `clearBuffer` in `Sources`/`Tests`.
- `xcodebuild build` + `test` on `iPhone 17 Pro`.

## Related Issues/Tickets
Related to MAGE-834
Part of MAGE-828

## Summary by CodeRabbit

- **New Features**
  - Added test-support `reset()` APIs for the event bus buffer, event dispatcher registration, identity data, and SDK configuration (including a renamed event-bus reset).
- **Bug Fixes**
  - Adjusted in-app form teardown/re-initialization so it no longer clears shared identity/config or resets internal event subjects during the webview/listener destruction path.
- **Tests**
  - Added and updated tests to verify replay buffer clearing, dispatcher reset behavior, deterministic event timing, and safer teardown/reset of shared singleton stores.

### DIFF
--- a/Sources/KlaviyoCore/EventBus.swift
+++ b/Sources/KlaviyoCore/EventBus.swift
@@ -46,7 +46,7 @@ public final class EventBus: EventPublishing, EventBroadcasting {
     }
 
     /// Test-support: clears the replay buffer for isolation between tests and for the
-    /// Core reset surface (MAGE-834). Subscribers are held by consumers, not the bus,
+    /// Core reset surface. Subscribers are held by consumers, not the bus,
     /// so a reset only clears the bounded replay buffer.
     public func reset() {
         buffer.clear()
--- a/Sources/KlaviyoCore/EventBus.swift
+++ b/Sources/KlaviyoCore/EventBus.swift
@@ -48,7 +48,7 @@ public final class EventBus: EventPublishing, EventBroadcasting {
     /// Test-support: clears the replay buffer for isolation between tests and for the
     /// Core reset surface. Subscribers are held by consumers, not the bus,
     /// so a reset only clears the bounded replay buffer.
-    public func reset() {
+    package func reset() {
         buffer.clear()
     }
 }
--- a/Sources/KlaviyoCore/EventBus.swift
+++ b/Sources/KlaviyoCore/EventBus.swift
@@ -45,9 +45,10 @@ public final class EventBus: EventPublishing, EventBroadcasting {
         .eraseToAnyPublisher()
     }
 
-    /// Test-support: clears the replay buffer for isolation between tests.
-    /// Not the Forms reset mechanism (that is MAGE-834).
-    package func clearBuffer() {
+    /// Test-support: clears the replay buffer for isolation between tests and for the
+    /// Core reset surface (MAGE-834). Subscribers are held by consumers, not the bus,
+    /// so a reset only clears the bounded replay buffer.
+    public func reset() {
         buffer.clear()
     }
 }
--- a/Sources/KlaviyoSwift/KlaviyoInternal.swift
+++ b/Sources/KlaviyoSwift/KlaviyoInternal.swift
@@ -160,8 +160,8 @@ package enum KlaviyoInternal {
     /// No-op. The event-reset mechanism moves onto KlaviyoCore in MAGE-834.
     package static func resetEventSubject() {}
 
-    /// Clears the event bus replay buffer to ensure clean state between tests.
-    package static func clearEventBuffer() {
+    /// Resets the event bus (clears its replay buffer) to ensure clean state between tests.
+    package static func resetEventBus() {
         EventBus.shared.reset()
     }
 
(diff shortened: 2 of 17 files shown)


belleklaviyo:
bugbot run

jasper-hsu-kyvo (approved):

Sources/KlaviyoSwift/KlaviyoInternal.swift
```diff
@@ -162,7 +162,7 @@ package enum KlaviyoInternal {
 
     /// Clears the event bus replay buffer to ensure clean state between tests.
     package static func clearEventBuffer() {
```
jasper-hsu-kyvo: q: should we rename `clearEventBuffer` to something like `resetEventBus`

Sources/KlaviyoCore/EventBus.swift
```diff
@@ -45,9 +45,10 @@ public final class EventBus: EventPublishing, EventBroadcasting {
         .eraseToAnyPublisher()
     }
 
-    /// Test-support: clears the replay buffer for isolation between tests.
-    /// Not the Forms reset mechanism (that is MAGE-834).
-    package func clearBuffer() {
+    /// Test-support: clears the replay buffer for isolation between tests and for the
+    /// Core reset surface. Subscribers are held by consumers, not the bus,
+    /// so a reset only clears the bounded replay buffer.
+    package func reset() {
```
glenn-klaviyo, line 51: since this function and others below are used for test support, can we/ should we wrap these in `#if DEBUG` compiler flags?

glenn-klaviyo (approved):
looks good from my perspective. one question

glenn-klaviyo (approved):

Sources/KlaviyoCore/EventBus.swift
belleklaviyo: hmm good point, apparently though the Makefile that runs tests would not be able to run in it `DEBUG` mode, so it would fail to compile in that case. but still good call out and I could at least scope down the reset funcs below to `package`

PR #648, opened 2026-07-14
feat: add create client subscription API

aahiltn:
# Description
Adds a public `KlaviyoSDK.create(subscription:)` API that enqueues `POST /client/subscriptions` so integrators can subscribe the current profile to a list with optional per-channel consent (email / SMS / WhatsApp) and an optional `customSource`.

A subscription that's missing identifiers at replay time is dropped with a developer warning rather than re-buffered.

Implementation of BSUID for the whatsapp channel is not yet implemented, waiting on that to be added to the current API revision. 

See API docs: 
https://developers.klaviyo.com/en/reference/create_client_subscription

## Due Diligence

## Changelog / Code Overview
- Public `Subscription` model + `KlaviyoSDK.create(subscription:)`
- `KlaviyoEndpoint.createSubscription` → `/client/subscriptions` with JSON:API list relationship and nested profile `subscriptions` consent
- Reuses `ProfilePayload` for the embedded profile; channel consent lives in `SubscriptionChannels`
- Validates identifiers against requested channels; buffers requests while initializing (same pattern as events/profiles)
- Warns and drops empty explicit channel configs so they are not collapsed into API default MARKETING consent
- Push marketing via `subscriptions.push` intentionally out of scope (existing `set(pushToken:)` path)

## Test Plan
- [x] `KlaviyoEndpointTests` — URL, revision, wire shape with/without channels (nil `subscriptions` omitted)
- [x] `StateManagementTests` — enqueue happy path, channels, init buffering, missing identifiers, empty channels
- [x] Manual: initialize SDK, set email/phone, call `create(subscription:)`, confirm 202 from `/client/subscriptions` in proxy/logs

## Related Issues/Tickets
Related to MAGE-854

## Summary by CodeRabbit

* **New Features**
  * Added `create(subscription:)` to create list subscription/consent records for email, SMS, and WhatsApp.
  * Supports channel-based consent selection (email marketing + open-tracking; SMS/WhatsApp marketing + transactional) with optional `customSource`.
  * Updated subscription creation to validate required identifiers and only enqueue after initialization.

* **Tests**
  * Added coverage for request URL/body/headers and conditional payload fields, plus state-management enqueueing, replay, and validation behaviors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@
   - [Unregistering from Geofencing](#unregistering-from-geofencing)
 - [Additional Details](#additional-details)
   - [Sandbox Support](#sandbox-support)
+  - [Logging](#logging)
   - [SDK Data Transfer](#sdk-data-transfer)
   - [Retries](#retries)
 - [Contributing](#contributing)
@@ -935,6 +936,20 @@ Klaviyo's SDK will determine and store the environment that your push token belo
 allowing your tokens to route sends to the correct environments. There is no additional setup needed.
 As long as you have deployed your application to Sandbox with our SDK employed to transmit push tokens to our backend,
 the ability to send and receive push on these Sandbox applications should work out-of-the-box.
+
+### Logging
+The SDK logs diagnostic information to the system console via `os.Logger`. Logging is enabled by default
+and can be toggled at runtime:
+
+```swift
+KlaviyoSDK().setLoggingEnabled(false) // silence all SDK logging
+KlaviyoSDK().setLoggingEnabled(true)  // re-enable logging
+KlaviyoSDK().isLoggingEnabled         // check the current setting
+```
+
+Disabling logging silences all log output from the `KlaviyoSwift`, `KlaviyoCore`, `KlaviyoForms`,
+and `KlaviyoLocation` modules. It does not affect `KlaviyoSwiftExtension`, which runs in a separate
+app-extension process.
 
 ### SDK Data Transfer
 Starting with version 1.7.0, the SDK will cache incoming data and flush it back to the Klaviyo API on an interval.
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@
   - [Anonymous Tracking Notice](#anonymous-tracking-notice)
 - [Event tracking](#event-tracking)
   - [Arguments](#arguments)
+- [List Subscriptions](#list-subscriptions)
+  - [Arguments](#arguments-1)
 - [Push Notifications](#push-notifications)
   - [Prerequisites](#prerequisites)
   - [Collecting Push Tokens](#collecting-push-tokens)
@@ -232,6 +234,51 @@ The `create` method takes an event object as an argument. The event can be const
 - [required] `name`: The name of the event you want to track, as a `EventName` enum. A list of common Klaviyo defined event metrics can be found in `Event.EventName`. You can also create custom events by using the `CustomEvent` enum case of `Event.EventName`
 - `properties`: A dictionary of properties that are specific to the event. This argument is optional.
 - `value`: A numeric value (`Double`) to associate with this event. For example, the dollar amount of a purchase.
+
+## List Subscriptions
+
+The SDK can create a subscription and consent record for email, SMS, and Whatsapp channels based on the profile's email and phone number attributes — via the [Create Client Subscription API](https://developers.klaviyo.com/en/reference/create_client_subscription). One of either email or phone number must be provided. WhatsApp BSUID support is coming soon.
+
+Set the profile's email and/or phone number **before** subscribing. A subscribe call that requests a channel whose identifier has not been set is not enqueued (a developer warning is logged).
+
+Profiles can be opted into multiple channels — for example email marketing, SMS marketing, and SMS transactional. Specify the channel(s) to subscribe by providing a `channels` value on `Subscription`. If you include `channels`, only those channels will be subscribed. If you use `Subscription.allAvailableMarketing` (no `subscriptions` object is sent), subscriptions are defaulted to `MARKETING`.
+
+Subscriptions to push notifications are not created through this API — use the existing [`set(pushToken:)`](#collecting-push-tokens) registration path instead.
+
+```swift
+KlaviyoSDK().set(email: "johnpork@demo.com")
+KlaviyoSDK().set(phoneNumber: "+15005550006")
+
+// Subscribe to a list, defaulting to MARKETING consent for every channel
+// available from the identifiers you've set:
+KlaviyoSDK().create(subscription: .allAvailableMarketing(listId: "XXXXXX"))
+
+// Or request specific channels and consent types:
+let subscription = Subscription(
+    listId: "XXXXXX",
+    channels: .init(
+        email: .marketing,
+        sms: [.marketing, .transactional],
+        whatsapp: [.marketing, .transactional]
+    ),
+    customSource: "iOS in-app form"
+)
+KlaviyoSDK().create(subscription: subscription)
+```
+### Arguments
+
+The `create(subscription:)` method takes a `Subscription`. Construct one with either:
+
+- `Subscription.allAvailableMarketing(listId:customSource:)` — omits the `subscriptions` object so the API defaults to `MARKETING` consent for every channel available from the profile's identifiers.
+- `Subscription(listId:channels:customSource:)` — sends a `subscriptions` object for only the channels you name. `channels` is required on this initializer so a broad grant is never the result of an omitted argument.
+
+`Subscription.Channels` takes optional `email`, `sms`, and `whatsapp` values. Each channel exposes only the consent sub-types the API supports for it — pass one value (e.g. `.marketing`) or multiple (e.g. `[.marketing, .transactional]`):
+
+The consent types per channel include:
+- `email`: `.marketing`, `.openTracking`
+- `sms` / `whatsapp`: `.marketing`, `.transactional`
+
+`customSource` is an optional free-text label describing where the signup originated (e.g. a form or screen name), stored as the consent record's source.
 
 ## Push Notifications
 
--- a/Sources/KlaviyoCore/Models/APIModels/CreateSubscriptionPayload.swift
+++ b/Sources/KlaviyoCore/Models/APIModels/CreateSubscriptionPayload.swift
@@ -1,0 +1,54 @@
+//
+//  CreateSubscriptionPayload.swift
+//
+//
+//  Created by Aahil Nishad on 7/7/26.
+//
+
+import Foundation
+
+public struct CreateSubscriptionPayload: Equatable, Codable {
+    public let data: SubscriptionData
+
+    public init(listId: String, profile: ProfilePayload, customSource: String? = nil) {
+        data = SubscriptionData(listId: listId, profile: profile, customSource: customSource)
+    }
+
+    public struct SubscriptionData: Equatable, Codable {
+        public var type = "subscription"
+        public let attributes: Attributes
+        public let relationships: Relationships
+
+        public init(listId: String, profile: ProfilePayload, customSource: String?) {
+            attributes = Attributes(customSource: customSource, profile: .init(data: profile))
+            relationships = Relationships(list: .init(data: .init(id: listId)))
+        }
+
+        public struct Attributes: Equatable, Codable {
+            public let customSource: String?
+            public let profile: Profile
+
+            enum CodingKeys: String, CodingKey {
+                case customSource = "custom_source"
+                case profile
+            }
+
+            public struct Profile: Equatable, Codable {
+                public let data: ProfilePayload
+            }
+        }
+
+        public struct Relationships: Equatable, Codable {
+            public let list: ListRelationship
+
+            public struct ListRelationship: Equatable, Codable {
+                public let data: ListData
+
+                public struct ListData: Equatable, Codable {
+                    public var type = "list"
+                    public let id: String
+                }
+            }
+        }
+    }
+}
--- a/Sources/KlaviyoCore/Models/APIModels/CreateSubscriptionPayload.swift
+++ b/Sources/KlaviyoCore/Models/APIModels/CreateSubscriptionPayload.swift
@@ -15,7 +15,7 @@ public struct CreateSubscriptionPayload: Equatable, Codable {
     }
 
     public struct SubscriptionData: Equatable, Codable {
-        public var type = "subscription"
+        public let type = "subscription"
         public let attributes: Attributes
         public let relationships: Relationships
 
@@ -45,7 +45,7 @@ public struct CreateSubscriptionPayload: Equatable, Codable {
                 public let data: ListData
 
                 public struct ListData: Equatable, Codable {
-                    public var type = "list"
+                    public let type = "list"
                     public let id: String
                 }
             }
--- a/Sources/KlaviyoSwift/Models/Subscription.swift
+++ b/Sources/KlaviyoSwift/Models/Subscription.swift
@@ -8,7 +8,7 @@
 import Foundation
 
 /// Represents a subscription request to subscribe a profile to a Klaviyo list.
-public struct Subscription: Equatable {
+public struct Subscription: Equatable, Sendable {
     /// The ID of the Klaviyo list to subscribe the profile to.
     public let listId: String
 
@@ -26,7 +26,7 @@ public struct Subscription: Equatable {
     /// exposes only the consent sub-types it supports, so invalid combinations (transactional email,
     /// open-tracking SMS) don't compile. Combine sub-types with array-literal syntax, e.g.
     /// `.init(sms: [.marketing, .transactional])`.
-    public struct Channels: Equatable {
+    public struct Channels: Equatable, Sendable {
         /// Consent sub-types to request on the EMAIL channel.
         public let email: Email?
         /// Consent sub-types to request on the SMS channel.
@@ -41,7 +41,7 @@ public struct Subscription: Equatable {
         }
 
         /// Consent sub-types supported on the EMAIL channel.
-        public struct Email: OptionSet, Equatable {
+        public struct Email: OptionSet, Equatable, Sendable {
             public let rawValue: Int
             public init(rawValue: Int) { self.rawValue = rawValue }
 
@@ -52,7 +52,7 @@ public struct Subscription: Equatable {
         }
 
         /// Consent sub-types supported on the SMS and WhatsApp channels.
-        public struct Messaging: OptionSet, Equatable {
+        public struct Messaging: OptionSet, Equatable, Sendable {
             public let rawValue: Int
             public init(rawValue: Int) { self.rawValue = rawValue }
 
--- a/Sources/KlaviyoSwift/Models/Subscription.swift
+++ b/Sources/KlaviyoSwift/Models/Subscription.swift
@@ -1,0 +1,91 @@
+//
+//  Subscription.swift
+//
+//
+//  Created by Aahil Nishad on 7/7/26.
+//
+
+import Foundation
+
+/// Represents a subscription request to subscribe a profile to a Klaviyo list.
+public struct Subscription: Equatable {
+    /// The ID of the Klaviyo list to subscribe the profile to.
+    public let listId: String
+
+    /// The channels to request consent for, or `nil` to defer to the server's default of MARKETING
+    /// consent on every identified channel. `nil` is only reachable through
+    /// ``allAvailableMarketing(listId:customSource:)`` — the public initializer requires channels so a
+    /// broad grant is never the result of an omitted argument.
+    public let channels: Channels?
+
+    /// Free-text label describing where this signup originated (e.g. a form or screen name).
+    /// Stored as the consent record's `$source`, pass an explicit value to override, or `nil` to omit it entirely.
+    public let customSource: String?
+
+    /// The channels to request consent for. Mirrors the API's `subscriptions` object: each channel
+    /// exposes only the consent sub-types it supports, so invalid combinations (transactional email,
+    /// open-tracking SMS) don't compile. Combine sub-types with array-literal syntax, e.g.
+    /// `.init(sms: [.marketing, .transactional])`.
+    public struct Channels: Equatable {
+        /// Consent sub-types to request on the EMAIL channel.
+        public let email: Email?
+        /// Consent sub-types to request on the SMS channel.
+        public let sms: Messaging?
+        /// Consent sub-types to request on the WhatsApp channel.
+        public let whatsapp: Messaging?
+
+        public init(email: Email? = nil, sms: Messaging? = nil, whatsapp: Messaging? = nil) {
+            self.email = email
+            self.sms = sms
+            self.whatsapp = whatsapp
+        }
+
+        /// Consent sub-types supported on the EMAIL channel.
+        public struct Email: OptionSet, Equatable {
+            public let rawValue: Int
+            public init(rawValue: Int) { self.rawValue = rawValue }
+
+            /// Email marketing consent.
+            public static let marketing = Email(rawValue: 1 << 0)
+            /// Email open-tracking consent.
+            public static let openTracking = Email(rawValue: 1 << 1)
+        }
+
+        /// Consent sub-types supported on the SMS and WhatsApp channels.
+        public struct Messaging: OptionSet, Equatable {
+            public let rawValue: Int
+            public init(rawValue: Int) { self.rawValue = rawValue }
+
+            /// Marketing consent.
+            public static let marketing = Messaging(rawValue: 1 << 0)
+            /// Transactional messaging consent.
+            public static let transactional = Messaging(rawValue: 1 << 1)
+        }
+    }
+
+    /// Creates a subscription request for the given channels.
+    /// - Parameters:
+    ///   - listId: The ID of the Klaviyo list to subscribe the profile to.
+    ///   - channels: The channels and sub-types to request consent for.
+    ///   - customSource: Optional signup-source label stored as the consent record's `$source`.
+    ///     Omitted from the request when `nil` (the default).
+    public init(listId: String, channels: Channels, customSource: String? = nil) {
+        self.init(listId: listId, channels: .some(channels), customSource: customSource)
+    }
+
+    private init(listId: String, channels: Channels?, customSource: String?) {
+        self.listId = listId
+        self.channels = channels
+        self.customSource = customSource
+    }
+
+    /// Creates a subscription that grants MARKETING consent on every channel the profile has an
+    /// identifier for (email → email marketing, phone → SMS marketing). Mirrors the server's default
+    /// behavior when no `subscriptions` object is sent, but requesting it is a deliberate call.
+    /// - Parameters:
+    ///   - listId: The ID of the Klaviyo list to subscribe the profile to.
+    ///   - customSource: Optional signup-source label stored as the consent record's `$source`.
+    public static func allAvailableMarketing(listId: String, customSource: String? = nil) -> Subscription {
+        Subscription(listId: listId, channels: nil, customSource: customSource)
+    }
+}
--- a/Sources/KlaviyoSwift/Models/SubscriptionAPIExtension.swift
+++ b/Sources/KlaviyoSwift/Models/SubscriptionAPIExtension.swift
@@ -1,0 +1,46 @@
+//
+//  SubscriptionAPIExtension.swift
+//
+//
+//  Created by Aahil Nishad on 7/14/26.
+//
+
+import Foundation
+import KlaviyoCore
+
+extension Subscription.Channels {
+    /// Whether the requested channels need an email identifier on the profile.
+    var needsEmail: Bool {
+        email?.isEmpty == false
+    }
+
+    /// Whether the requested channels need a phone number identifier on the profile.
+    var needsPhone: Bool {
+        sms?.isEmpty == false || whatsapp?.isEmpty == false
+    }
+
+    /// Maps the requested channels to the API's `subscriptions` object, or `nil` when no sub-types
+    /// were named (nothing to send).
+    var toAPIModel: SubscriptionChannels? {
+        let emailConsent = email.flatMap { set -> EmailConsent? in
+            guard !set.isEmpty else { return nil }
+            return EmailConsent(
+                marketing: set.contains(.marketing) ? .subscribed : nil,
+                openTracking: set.contains(.openTracking) ? .subscribed : nil
+            )
+        }
+        let smsConsent = messagingConsent(sms)
+        let whatsappConsent = messagingConsent(whatsapp)
+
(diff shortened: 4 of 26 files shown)


aahiltn:
bugbot run

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -601,6 +631,70 @@ struct KlaviyoReducer: ReducerProtocol {
...
+        case let .enqueueSubscription(subscription):
+            guard case .initialized = state.initalizationState,
+                  let apiKey = state.apiKey,
+                  let anonymousId = state.anonymousId
+            else {
+                environment.emitDeveloperWarning(
+                    "SDK must be initialized before creating a subscription; request was not enqueued."
+                )
+                return .none
+            }
```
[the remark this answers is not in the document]
aahiltn: This is intended behavior. Subscriptions can only be queued if email or phone num have been set on a profile. We could potentially have a two pass queue for processing events to ensure that email comes first, but (open for discussion) I believe it makes logical sense to just reject subscriptions if so, from a product standpoint.

Sources/KlaviyoSwift/Models/Subscription.swift
```diff
@@ -0,0 +1,75 @@
...
+//  Subscription.swift
+//
+//
+//  Created by Aahil Nishad on 7/7/26.
+//
+
+import Foundation
+
+/// Represents a subscription request to subscribe a profile to a Klaviyo list.
+public struct Subscription: Equatable {
```
ndurell: Did we discuss these apis with the team?

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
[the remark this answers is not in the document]
ndurell: I thought normally if things aren't initialized yet we can enqueue them for later processing after initialization completes?

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -601,6 +632,70 @@ struct KlaviyoReducer: ReducerProtocol {
...
+                profile: profile,
+                customSource: subscription.customSource
+            )
+
+            let endpoint = KlaviyoEndpoint.createSubscription(apiKey, payload)
+            let request = KlaviyoRequest(endpoint: endpoint)
+            state.enqueueRequest(request: request)
+
+            return .none
+
```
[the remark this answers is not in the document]
ndurell: +1

Sources/KlaviyoSwift/Models/Subscription.swift
ndurell: @aahiltn following up on my question above with a concrete concern: I'd like us to rethink the `Channels` shape before this ships, since it's frozen public API once released. Can you bring this to the team? We should also loop in the Android folks — we'll likely want the same contract there, and renaming after release is a breaking change on both platforms.

Issues with the six flat `Bool`s:
- `Channels()` with all flags false is a representable-but-meaningless value, which is why the reducer needs the empty-channels warning/drop branch at all.
- The wire format is nested (email: `marketing`/`open_tracking`; SMS/WhatsApp: `marketing`/`transactional`) but the public API flattens it, so the compiler can't catch a transposed flag, and every new channel/sub-type means lockstep edits across `Channels`, `SubscriptionChannels.subscribed(...)`, and `toAPIModel`.
- Bools can only ever express "subscribed" — no room to grow if the endpoint ever accepts other consent values.

Proposal — mirror the wire shape with one optional field per channel plus small per-channel OptionSets:

```swift
public struct Channels: Equatable {
    public let email: EmailConsent?       // OptionSet: .marketing, .openTracking
    public let sms: MessagingConsent?     // OptionSet: .marketing, .transactional
    public let whatsapp: MessagingConsent?
}

// call sites:
Subscription(listId: "L123", channels: .init(email: .marketing))
Subscription(listId: "L123",
             channels: .init(email: [.marketing, .openTracking],
                             sms: [.marketing, .transactional]))
```

- Invalid combinations (transactional email, open-tracking SMS) don't compile; `needsEmail`/`needsPhone` become simple nil-checks; normalizing `[]` → `nil` in the init removes the empty-channels warning path.
- Maps 1:1 to `SubscriptionChannels`, so the 6-param `subscribed(...)` factory and most of the translation layer disappear.
- Extensible without breaking source: new channel = new optional field, new sub-type = new static member.

Related: `channels: nil` being the *default* while producing the broadest grant (server defaults to MARKETING on every identified channel) worries me from a consent/compliance angle — e.g. a profile with a phone number collected for transactional SMS gets SMS marketing consent from a call that meant "subscribe to this email list." I'd make that explicit rather than the omitted-parameter behavior, e.g. a required `channels` with an explicit `.allAvailableMarketing` option.

---
_Generated by [Claude Code](https://claude.ai/code)_

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -778,4 +778,94 @@ class StateManagementTests: XCTestCase {
...
+        let initialState = INITILIZING_TEST_STATE()
+        let store = TestStore(initialState: initialState, reducer: KlaviyoReducer())
+
+        await store.send(.enqueueSubscription(Subscription(listId: "list-123")))
+        XCTAssertTrue(store.state.pendingRequests.isEmpty)
+        XCTAssertTrue(store.state.queue.isEmpty)
+    }
+
+    @MainActor
+    func testEnqueueSubscriptionMissingIdentifiersDoesNotEnqueue() async throws {
```
ndurell, line 892: The two most bug-prone validation branches in `enqueueSubscription` have no coverage: "email channels requested but no email set" and "phone channels requested but no phone set" (the `needsEmail` / `needsPhone` guards). Since all the "does not enqueue" tests produce the same observable result (empty queue), a regression that swapped or deleted those checks would pass the suite.

Could we add two focused tests — e.g. `channels: .init(emailMarketing: true)` with only a phone set, and `channels: .init(smsMarketing: true)` with only an email set — each asserting the queue stays empty?

---
_Generated by [Claude Code](https://claude.ai/code)_

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -778,4 +778,94 @@ class StateManagementTests: XCTestCase {
...
+                            )
+                        )
+                    )
+                )
+            )
+        }
+    }
+
+    @MainActor
+    func testEnqueueSubscriptionBeforeInitializedDoesNotEnqueue() async throws {
```
ndurell: These three "does not enqueue" tests assert only that the queue stays empty, but never that the expected developer warning fired. Since several distinct guards in `enqueueSubscription` all produce the same observable result, the tests can't distinguish "the right guard fired" from "some unrelated early return."

There's an existing pattern for exactly this in `StateManagementEdgeCaseTests.swift` — override `environment.emitDeveloperWarning` with an `XCTestExpectation` (optionally asserting on the message) to pin down which branch was taken.

---
_Generated by [Claude Code](https://claude.ai/code)_

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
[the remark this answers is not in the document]
ndurell: Adding some context from the git history: an earlier version of this branch *did* buffer subscriptions, and commit `4fa237b` ("fix: remove subscription buffering") deliberately removed it — but the PR description still says subscriptions are "buffer[ed]... while initializing (same pattern as events/profiles)", so the description and code currently contradict each other.

Two things worth separating in the "intended behavior" rationale:

1. Rejecting when **no identifiers are set** is a defensible product rule. But the init guard fires before any identifier validation — during the `.initializing` window *everything* is dropped, including calls from users whose email/phone are already persisted from a previous session and just haven't loaded from disk yet. `initialize()` loads state asynchronously, so `create(subscription:)` called right after it can silently lose the subscription even though the profile has valid identifiers. That's a timing race, not a product decision, and whether it fails depends on disk-load latency.

2. If we do restore buffering, note the ordering bug the earlier (now resolved) Bugbot thread found: `pendingRequests` replays FIFO, so a subscription buffered before `setEmail`/`setPhoneNumber` would replay first and fail identifier validation against not-yet-updated state. Buffering needs to either order subscriptions after identifier setters or defer validation to replay time.

Either way, if the drop behavior stays, the PR description and the `create(subscription:)` doc comment should state it explicitly so integrators don't rely on the event/profile pattern.

---
_Generated by [Claude Code](https://claude.ai/code)_

Tests/KlaviyoCoreTests/KlaviyoEndpointTests.swift
```diff
@@ -283,4 +283,108 @@ final class KlaviyoEndpointTests: XCTestCase {
...
+        let profile = try XCTUnwrap(attributes["profile"] as? [String: Any])
+        let profileData = try XCTUnwrap(profile["data"] as? [String: Any])
+        let profileAttrs = try XCTUnwrap(profileData["attributes"] as? [String: Any])
+        XCTAssertNil(
+            profileAttrs["subscriptions"],
+            "omitting channels must omit subscriptions so the API can default to MARKETING"
+        )
+    }
+
+    func testCreateSubscriptionEndpointWithChannels() throws {
```
ndurell, line 320: Channel serialization is only exercised at the extremes — all six flags on (here) or no channels at all (`testCreateSubscriptionEndpointUrlRequest`). No test covers a partial combination, so a bug where e.g. `emailOpenTracking` alone fails to build an `EmailConsent`, or a disabled channel's key leaks into the JSON, would pass undetected.

Could we add a partial case — e.g. only `smsMarketing: true` — asserting `sms.marketing` is present while `sms.transactional`, `email`, and `whatsapp` are entirely absent from the payload?

---
_Generated by [Claude Code](https://claude.ai/code)_

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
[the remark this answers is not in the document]
aahiltn: ~In this case, the difference is subscriptions have a hard dependency on email/phone that events/profiles don't, since those can use anonymousIds.~

~With the existing approach, buffering them alongside everything else in one FIFO doesn't actually work: if a dev calls create(subscription:) before set(email:), the replay runs subscription-first and fails. An option would be to switch out our existing FIFO queue to include a two pass approach that always sets emails/phone numbers first, but I felt that was unnecessary complexity, and it was better off just documenting / erroring if email or sms hadn't been set yet. (spoke with @ab1470 about this)~

~From a product standpoint as well, I'd also say it is reasonable for a developer to collect these fields first before subscribing a user to a list.~

Conflated two different issues. Went ahead and implemented buffering, where after the initialize stage, create before setEmail in the pending batch → warn + drop on replay (same as after init).

Sources/KlaviyoSwift/Models/Subscription.swift
aahiltn: Ya, the team is aware that we're adding this API to the SDK. 

I'm most concerned about points 2 and 4, and I can address those by removing the factory and adding an `.allAvailableMarketing` option. With these changes, the callsites would now look like 

```
Subscription(listId: "L123", channels: .init(email: .marketing))
Subscription(listId: "L123", channels: .init(email: [.marketing, .openTracking], sms: .marketing))
Subscription.allAvailableMarketing(listId: "L123")
```

Note that channels is now a required parameter. allAvailableMarketing is also the server default, just now documented explicitly. This approach also does keep the guard on sending `channels: init()`.

Tests/KlaviyoSwiftTests/StateManagementTests.swift
aahiltn: addressed

Tests/KlaviyoSwiftTests/StateManagementTests.swift
aahiltn: addressed

Tests/KlaviyoCoreTests/KlaviyoEndpointTests.swift
aahiltn: addressed testCreateSubscriptionEndpointWithPartialChannels()

aahiltn:
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > Sources/KlaviyoSwift/StateManagement/StateManagement.swift (1)> `637-656`: _🎯 Functional Correctness_ | _🟠 Major_ | _🏗️ Heavy lift_
> > > **Buffered subscriptions can still be skipped on replay**
> > > `pendingRequests` replays in FIFO order, so a buffered `.subscription` can run before a later `.setEmail`/`.setPhoneNumber` and fail identifier validation at replay time. That drops the subscription behind a console warning. Consider snapshotting the needed identifiers with the subscription or deferring subscription replay until identity actions have been applied.
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against current code. Fix only still-valid issues, skip the
> > > rest with a brief reason, keep changes minimal, and validate.
> > > 
> > > In `@Sources/KlaviyoSwift/StateManagement/StateManagement.swift` around lines 637
> > > - 656, Update the enqueueSubscription and pendingRequests replay flow so
> > > buffered subscriptions are not discarded when replayed before later identity
> > > actions. Preserve the subscription’s required identifier state by snapshotting
> > > the needed identifiers when buffering, or defer its replay until
> > > setEmail/setPhoneNumber actions have been applied; ensure
> > > buildSubscriptionRequest is retried successfully instead of returning after
> > > validation failure.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 
> Outside diff comments:
> In `@Sources/KlaviyoSwift/StateManagement/StateManagement.swift`:
> - Around line 637-656: Update the enqueueSubscription and pendingRequests replay
> flow so buffered subscriptions are not discarded when replayed before later
> identity actions. Preserve the subscription’s required identifier state by
> snapshotting the needed identifiers when buffering, or defer its replay until
> setEmail/setPhoneNumber actions have been applied; ensure
> buildSubscriptionRequest is retried successfully instead of returning after
> validation failure.
> ```
> 
> ℹ️ Review info

We can drop subscriptions if email or phone num has not yet been set.

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -778,4 +778,296 @@ class StateManagementTests: XCTestCase {
...
+                                    email: EmailConsent(marketing: .subscribed),
+                                    sms: MarketingTransactionalConsent(marketing: .subscribed)
+                                ),
+                                anonymousId: XCTUnwrap(initialState.anonymousId)
+                            )
+                        )
+                    )
+                )
+            )
+        }
```
glenn-klaviyo: in swift, generally it's a good idea to avoid large nesting calls like this as it gets harder to read the more you add. you could clean this up by pulling out the request into individual properties. 

would end up looking something more like this:
```
    @MainActor
    func testEnqueueSubscriptionWithChannels() async throws {
        var initialState = INITIALIZED_TEST_STATE()
        initialState.email = "test@example.com"
        initialState.phoneNumber = "+15005550006"

        let apiKey = try XCTUnwrap(initialState.apiKey)
        let anonymousId = try XCTUnwrap(initialState.anonymousId)

        let subscription = Subscription(
            listId: "list-123",
            channels: .init(email: .marketing, sms: .marketing)
        )
        let payload = CreateSubscriptionPayload(
            listId: "list-123",
            profile: ProfilePayload(
                email: "test@example.com",
                phoneNumber: "+15005550006",
                subscriptions: SubscriptionChannels(
                    email: EmailConsent(marketing: .subscribed),
                    sms: MarketingTransactionalConsent(marketing: .subscribed)
                ),
                anonymousId: anonymousId
            )
        )
        let request = KlaviyoRequest(endpoint: .createSubscription(apiKey, payload))

        await store.send(.enqueueSubscription(subscription)) {
            $0.enqueueRequest(request: request)
        }
    }
```

aahiltn: ah, cleaner! appreciate it and addressed

Sources/KlaviyoSwift/Models/SubscriptionAPIExtension.swift
```diff
@@ -0,0 +1,46 @@
...
+//  Created by Aahil Nishad on 7/14/26.
+//
+
+import Foundation
+import KlaviyoCore
+
+extension Subscription.Channels {
+    /// Whether the requested channels need an email identifier on the profile.
+    var needsEmail: Bool {
+        !(email?.isEmpty ?? true)
```
belleklaviyo: just to avoid a double negative could we do `{ email?.isEmpty == false }` instead?

Sources/KlaviyoSwift/Models/SubscriptionAPIExtension.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /// Whether the requested channels need a phone number identifier on the profile.
+    var needsPhone: Bool {
+        !(sms?.isEmpty ?? true) || !(whatsapp?.isEmpty ?? true)
```
belleklaviyo: same here

belleklaviyo:
are we planning any README updates with this?

README.md
```diff
@@ -233,6 +235,53 @@ The `create` method takes an event object as an argument. The event can be const
...
 
+## List Subscriptions
+
+The SDK can create a subscription and consent record for email, SMS, and Whatsapp channels based on the profile's email and phone number attributes — via the [Create Client Subscription API](https://developers.klaviyo.com/en/reference/create_client_subscription). One of either email or phone number must be provided. WhatsApp BSUID support is coming soon.
+
+Set the profile's email and/or phone number **before** subscribing. A subscribe call that requests a channel whose identifier has not been set is not enqueued (a developer warning is logged).
+
+Profiles can be opted into multiple channels — for example email marketing, SMS marketing, and SMS transactional. Specify the channel(s) to subscribe by providing a `channels` value on `Subscription`. If you include `channels`, only those channels will be subscribed. If you use `Subscription.allAvailableMarketing` (no `subscriptions` object is sent), subscriptions are defaulted to `MARKETING`.
+
+Subscriptions to push are not created through this API — use the existing [`set(pushToken:)`](#collecting-push-tokens) registration path instead.
```
belleklaviyo: nit: push notifications

README.md
```diff
@@ -233,6 +235,53 @@ The `create` method takes an event object as an argument. The event can be const
...
+
+```swift
+KlaviyoSDK().set(email: "johnpork@demo.com")
```
belleklaviyo, line 249: 👨 🐷 nice

README.md
```diff
@@ -233,6 +235,53 @@ The `create` method takes an event object as an argument. The event can be const
...
+KlaviyoSDK().set(phoneNumber: "+15005550006")
+
+// Subscribe to a list, defaulting to MARKETING consent for every channel
+// available from the identifiers you've set:
+KlaviyoSDK().create(subscription: .allAvailableMarketing(listId: "Y6nRLr"))
```
belleklaviyo: is this a real list id, just check it's not

README.md
```diff
@@ -233,6 +235,53 @@ The `create` method takes an event object as an argument. The event can be const
...
+        email: .marketing,
+        sms: [.marketing, .transactional],
+        whatsapp: [.marketing, .transactional]
+    ),
+    customSource: "iOS in-app form"
+)
+KlaviyoSDK().create(subscription: subscription)
+```
+
+Like other SDK calls, subscription requests are queued and sent in batches; the API responds asynchronously (`202 Accepted`).
```
belleklaviyo: feel like we don't need this

README.md
aahiltn: replaced w a placeholder `XXXXXX`

README.md
```diff
@@ -233,6 +235,53 @@ The `create` method takes an event object as an argument. The event can be const
...
+
+### Arguments
+
+The `create(subscription:)` method takes a `Subscription`. Construct one with either:
+
+- `Subscription.allAvailableMarketing(listId:customSource:)` — omits the `subscriptions` object so the API defaults to `MARKETING` consent for every channel available from the profile's identifiers. Requires at least one of email or phone number to be set.
```
belleklaviyo: think you can omit `Requires at least one of email or phone number to be set.` already mentioned above

README.md
```diff
@@ -233,6 +235,53 @@ The `create` method takes an event object as an argument. The event can be const
...
+- `Subscription(listId:channels:customSource:)` — sends a `subscriptions` object for only the channels you name. `channels` is required on this initializer so a broad grant is never the result of an omitted argument.
+
+`Subscription.Channels` takes optional `email`, `sms`, and `whatsapp` values. Each channel exposes only the consent sub-types the API supports for it, as `OptionSet`s you can combine with array-literal syntax:
+
+- `email`: `.marketing`, `.openTracking`
+- `sms` / `whatsapp`: `.marketing`, `.transactional`
```
belleklaviyo, line 279: can you just show it as a set like `[.email, .sms]` instead of saying `OptionSet`

README.md
```diff
@@ -233,6 +235,53 @@ The `create` method takes an event object as an argument. The event can be const
...
+
+`customSource` is an optional free-text label describing where the signup originated (e.g. a form or screen name), stored as the consent record's source. Omitted from the request when `nil`.
```
belleklaviyo: can omit `Omitted from the request when nil`

belleklaviyo (approved):
thanks for addressing all my nit comments, looks good! looking forward to testing it with the test app updates too

Sources/KlaviyoSwift/Models/SubscriptionAPIExtension.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /// Whether the requested channels need a phone number identifier on the profile.
+    var needsPhone: Bool {
+        sms?.isEmpty == false || whatsapp?.isEmpty == false
+    }
+
+    /// Maps the requested channels to the API's `subscriptions` object, or `nil` when no sub-types
+    /// were named (nothing to send).
+    var toAPIModel: SubscriptionChannels? {
```
glenn-klaviyo, line 24: in general this is fine but I've personally never been a huge fan of this `toAPIModel` convention. the alternative to this is to use "failable" optional conversion initializers on the type itself. 

So this file would become
```
import Foundation
import KlaviyoCore

extension Subscription.Channels {
    /// Whether the requested channels need an email identifier on the profile.
    var needsEmail: Bool {
        email?.isEmpty == false
    }

    /// Whether the requested channels need a phone number identifier on the profile.
    var needsPhone: Bool {
        sms?.isEmpty == false || whatsapp?.isEmpty == false
    }
}

extension EmailConsent {
    /// Maps the requested EMAIL sub-types to the API consent object, or `nil` when none were named.
    init?(_ options: Subscription.Channels.Email) {
        guard !options.isEmpty else { return nil }
        self.init(
            marketing: options.contains(.marketing) ? .subscribed : nil,
            openTracking: options.contains(.openTracking) ? .subscribed : nil
        )
    }
}

extension MarketingTransactionalConsent {
    /// Maps the requested SMS/WhatsApp sub-types to the API consent object, or `nil` when none were named.
    init?(_ options: Subscription.Channels.Messaging) {
        guard !options.isEmpty else { return nil }
        self.init(
            marketing: options.contains(.marketing) ? .subscribed : nil,
            transactional: options.contains(.transactional) ? .subscribed : nil
        )
    }
}

extension SubscriptionChannels {
    /// Maps the requested channels to the API's `subscriptions` object, or `nil` when no sub-types
    /// were named (nothing to send).
    init?(_ channels: Subscription.Channels) {
        let email = channels.email.flatMap(EmailConsent.init)
        let sms = channels.sms.flatMap(MarketingTransactionalConsent.init)
        let whatsapp = channels.whatsapp.flatMap(MarketingTransactionalConsent.init)

        guard email != nil || sms != nil || whatsapp != nil else { return nil }
        self.init(email: email, sms: sms, whatsapp: whatsapp)
    }
}
```

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -314,6 +315,63 @@ struct KlaviyoState: Equatable, Codable {
...
+            }
+            if requestedChannels.needsPhone, phoneNumber == nil {
+                environment.emitDeveloperWarning(
+                    "Subscription requires a phone number for the requested channels, " +
+                        "but phone number is not set."
+                )
+                return nil
+            }
+
+            guard let mappedChannels = requestedChannels.toAPIModel else {
```
glenn-klaviyo, line 338: and with the optional initializers, this then becomes
```
guard let mappedChannels = SubscriptionChannels(requestedChannels) {
```

glenn-klaviyo (approved):
LGTM

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
aahiltn: ooh I like this. I'm happy to make these changes, I'll scope them in a separate PR.

Sources/KlaviyoSwift/Models/SubscriptionAPIExtension.swift
ndurell: @aahiltn did you see this?

ndurell: This seems nicer to me!

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
ndurell: These are internal changes right? If so then it's fine to address later. If not I think we need to do it now.

aahiltn: Internal changes, so we can address later. Seems like a great idea!

ndurell (approved):

Sources/KlaviyoCore/Models/APIModels/CreateSubscriptionPayload.swift
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        public struct Relationships: Equatable, Codable {
+            public let list: ListRelationship
+
+            public struct ListRelationship: Equatable, Codable {
+                public let data: ListData
+
+                public struct ListData: Equatable, Codable {
+                    public var type = "list"
```
ab1470: nit: can this be a `let` because it'll never change?

Sources/KlaviyoCore/Models/APIModels/CreateSubscriptionPayload.swift
```diff
@@ -0,0 +1,54 @@
...
+
+public struct CreateSubscriptionPayload: Equatable, Codable {
+    public let data: SubscriptionData
+
+    public init(listId: String, profile: ProfilePayload, customSource: String? = nil) {
+        data = SubscriptionData(listId: listId, profile: profile, customSource: customSource)
+    }
+
+    public struct SubscriptionData: Equatable, Codable {
+        public var type = "subscription"
```
ab1470: same comment as [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/648/changes#r3615434082)

Atharva8168 (approved):
LGTM overall, no changes needed to ship this. Just a non-blocking notes for future reference:

1. The PR description says subscription requests are "buffer[ed]... while initializing (same pattern as events/profiles)", but per the thread on `StateManagement.swift` (commit `4fa237b` deliberately removed subscription buffering), a subscription that's missing identifiers at replay time is now dropped with a developer warning rather than re-buffered.

Sources/KlaviyoSwift/Models/Subscription.swift
```diff
@@ -0,0 +1,91 @@
...
+        public let whatsapp: Messaging?
+
+        public init(email: Email? = nil, sms: Messaging? = nil, whatsapp: Messaging? = nil) {
+            self.email = email
+            self.sms = sms
+            self.whatsapp = whatsapp
+        }
+
+        /// Consent sub-types supported on the EMAIL channel.
+        public struct Email: OptionSet, Equatable {
```
ab1470:
```suggestion
        public struct Email: OptionSet, Equatable, Sendable {
```
we can add some thread safety and minimize compiler warnings by conforming to `Sendable`
https://www.avanderlee.com/swift/sendable-protocol-closures/

Sources/KlaviyoSwift/Models/Subscription.swift
```diff
@@ -0,0 +1,91 @@
...
+            public init(rawValue: Int) { self.rawValue = rawValue }
+
+            /// Email marketing consent.
+            public static let marketing = Email(rawValue: 1 << 0)
+            /// Email open-tracking consent.
+            public static let openTracking = Email(rawValue: 1 << 1)
+        }
+
+        /// Consent sub-types supported on the SMS and WhatsApp channels.
+        public struct Messaging: OptionSet, Equatable {
```
ab1470:
```suggestion
        public struct Messaging: OptionSet, Equatable, Sendable {
```
same as [above](https://github.com/klaviyo/klaviyo-swift-sdk/pull/648/changes#r3615635263)

Sources/KlaviyoSwift/Models/Subscription.swift
```diff
@@ -0,0 +1,91 @@
...
+//  Subscription.swift
+//
+//
+//  Created by Aahil Nishad on 7/7/26.
+//
+
+import Foundation
+
+/// Represents a subscription request to subscribe a profile to a Klaviyo list.
+public struct Subscription: Equatable {
```
ab1470:
```suggestion
public struct Subscription: Equatable, Sendable {
```
same as [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/648/changes#r3615635263)

Sources/KlaviyoSwift/Models/Subscription.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    /// Free-text label describing where this signup originated (e.g. a form or screen name).
+    /// Stored as the consent record's `$source`, pass an explicit value to override, or `nil` to omit it entirely.
+    public let customSource: String?
+
+    /// The channels to request consent for. Mirrors the API's `subscriptions` object: each channel
+    /// exposes only the consent sub-types it supports, so invalid combinations (transactional email,
+    /// open-tracking SMS) don't compile. Combine sub-types with array-literal syntax, e.g.
+    /// `.init(sms: [.marketing, .transactional])`.
+    public struct Channels: Equatable {
```
ab1470:
```suggestion
    public struct Channels: Equatable, Sendable {
```
same as [here](https://github.com/klaviyo/klaviyo-swift-sdk/pull/648/changes#r3615635263)

ab1470 (approved):
nice! Just some small comments

PR #649, opened 2026-07-14
refactor: cut KlaviyoSwift from Forms & Location public entry points (MAGE-893)

belleklaviyo:
# Description
Removes `import KlaviyoSwift` from the Forms & Location public entry-point files via a new Core-owned marker protocol, and drops KlaviyoForms' module-level KlaviyoSwift dependency — fully source-compatible.

## Due Diligence
  - Pure refactor with no behavior change; the existing `KlaviyoFormsTests`/`KlaviyoLocationTests` (which call the public `KlaviyoSDK()` entry points) serve as the source-compatibility regression and pass.
- [x] I have adjusted or added new test cases to team test docs, if applicable.

  - Adds `public protocol KlaviyoSDKExtending {}` to KlaviyoCore. Existing call sites (`KlaviyoSDK().registerForInAppForms()`, `.registerGeofencing()`, etc.) are unchanged and remain source-compatible.
  - Part of the Modularize iOS SDK effort; targets the `feat/modularization` feature branch.

## Changelog / Code Overview
**Approach:** Option 3b — marker protocol in KlaviyoCore (non-breaking).

- **KlaviyoCore:** new `KlaviyoSDKExtending` marker protocol (no requirements).
- **KlaviyoSwift:** `extension KlaviyoSDK: KlaviyoSDKExtending {}` (KlaviyoSwift owns `KlaviyoSDK`, so this is not a retroactive conformance).
- **KlaviyoForms / KlaviyoLocation entry points:** rewritten from `extension KlaviyoSDK` to `extension KlaviyoSDKExtending`, importing only KlaviyoCore, with chaining methods returning `-> Self`. At a concrete call site the protocol-extension member resolves and `Self` binds to `KlaviyoSDK`, so the public API and chaining are preserved.
- **Dependency graph:** KlaviyoForms' product target + podspec switch `KlaviyoSwift` → `KlaviyoCore` (Forms-only integrators stop pulling KlaviyoSwift). The podspec's `OTHER_SWIFT_FLAGS` drops `-package-name KlaviyoSwift`.

**Scope notes:**
- KlaviyoLocation **keeps** its KlaviyoSwift module dependency — `KlaviyoLocationManager+CLLocationManagerDelegate` still uses the `package` symbol `GeofenceEventDispatch`. Only Location's entry-point *file* is realigned, to keep the Forms and Location public entry points consistent.
- DocC `` `KlaviyoSDK` `` symbol links in the doc comments may not resolve without the import (cosmetic, non-functional).

## Test Plan
1. `xcodebuild build -scheme KlaviyoForms -destination "platform=iOS Simulator,name=iPhone 17 Pro"` → builds with **no** KlaviyoSwift dependency (proves the dep drop is safe).
2. `xcodebuild build -scheme klaviyo-swift-sdk-Package ...` → BUILD SUCCEEDED.
3. `xcodebuild test ... -only-testing:KlaviyoFormsTests -only-testing:KlaviyoLocationTests` → passing (source-compat regression: existing tests exercise the public `KlaviyoSDK()` entry points).

_Note: `swiftformat` was not available in the authoring environment; files are verbatim from the reviewed plan. Run `swiftformat .` if pre-commit/CI flags formatting._

## Related Issues/Tickets
Resolves MAGE-893
Part of MAGE-892

## Summary by CodeRabbit

* **New Features**
  * Added a shared SDK module interface for Forms and Location features.
  * Forms and geofencing APIs now preserve the specific SDK type when chaining methods.
* **Improvements**
  * Updated package integration so Forms and Location rely on the core SDK module.
  * Existing registration, lifecycle, and geofencing behavior remains unchanged.


belleklaviyo:
bugbot run

Package.swift
```diff
@@ -71,7 +71,7 @@ let package = Package(
         ),
         .target(
             name: "KlaviyoForms",
-            dependencies: ["KlaviyoSwift"],
+            dependencies: ["KlaviyoCore"],
```
belleklaviyo, line 74: 🪓 ❗ ❗

timstephens-k:
why are we codeowners here?

Sources/KlaviyoCore/KlaviyoSDKExtending.swift
```diff
@@ -0,0 +1,15 @@
...
+//
+
+/// Marker protocol that anchors public `KlaviyoSDK` API extensions declared in feature modules
+/// (Forms, Location) without those modules importing `KlaviyoSwift`.
+///
+/// `KlaviyoSwift` conforms `KlaviyoSDK` to this protocol. Feature modules extend
+/// `KlaviyoSDKExtending` instead of `KlaviyoSDK` directly, depending only on `KlaviyoCore` for the
+/// type anchor. At a concrete call site the extension member resolves and `Self` binds to
+/// `KlaviyoSDK`, so the public API and method chaining are preserved.
+public protocol KlaviyoSDKExtending {}
```
ajaysubra: I'm still wrapping my head around this pattern but is this protocol/interface essentially something that blueprints a feature module? If so can we call it something like `KlaviyoSDKModule` or something instead of extending?

Sources/KlaviyoSwift/KlaviyoSDK+Extending.swift
```diff
@@ -0,0 +1,10 @@
+//
+//  KlaviyoSDK+Extending.swift
+//  klaviyo-swift-sdk
+//
+//  Created by Isobelle Lim on 7/14/26.
+//
+
+import KlaviyoCore
+
+extension KlaviyoSDK: KlaviyoSDKExtending {}
```
ajaysubra: Any reason for this to be it's own file?

ajaysubra (approved):
LGTM with just nit-picks because why not!

Sources/KlaviyoCore/KlaviyoSDKExtending.swift
belleklaviyo: fair, yeah it's kinda abstracting the KlaviyoSDK object to a protocol defined in KlaviyoCore, so the feature modules can extend this protocol rather than the object itself. this pattern makes it non breaking to existing implementations. but yeah can rename to something better

Sources/KlaviyoSwift/KlaviyoSDK+Extending.swift
belleklaviyo: fair, will condense

Sources/KlaviyoCore/KlaviyoSDKExtending.swift
glenn-klaviyo: I like thea idea of calling this `KlaviyoSDKModule`

Sources/KlaviyoCore/KlaviyoSDKModule.swift
```diff
@@ -0,0 +1,15 @@
+//
```
glenn-klaviyo, line 1: we should probs rename this file as well to `KlaviyoSDKModule`

glenn-klaviyo (approved):
LGTM, small comment

Sources/KlaviyoCore/KlaviyoSDKModule.swift
belleklaviyo: oops good catch

PR #666, opened 2026-07-21
feat(push): open web URL support for notifications, push actions and IAF

morrilltim:
# Description
Ships the "open web URL" push/IAF work for the 5.4.0 release: `open_url` support for push body taps and action buttons, non-web scheme allowlisting, `deep_link`-over-`web_url` precedence, and the in-app form `openExternalUrl` bridge event.

## Due Diligence
- [x] I have tested this on a simulator or a physical device.
- [x] I have added sufficient unit/integration tests of my changes.
- [ ] I have adjusted or added new test cases to team test docs, if applicable.
- [x] I am confident these changes are compatible with all iOS and XCode versions the SDK currently supports.

## Release/Versioning Considerations
- [ ] `Patch` Contains internal changes or backwards-compatible bug fixes.
- [x] `Minor` Contains changes to the public API.
- [ ] `Major` Contains **breaking** changes.
- [ ] Contains readme or migration guide changes.
- [x] This is planned work for an upcoming release.

Note: adds public `FormLifecycleEvent.formExternalUrlClicked` case (PUSH-732) — source-breaking for any exhaustive switch over `FormLifecycleEvent`; call out in changelog.

## Changelog / Code Overview
Consolidates the following merged PRs from the `feat/push-638-open-url` integration branch:
- #580 — handle `open_url` push tap and action button (PUSH-638)
- #615 — prefer `deep_link` over `web_url` when both are present (PUSH-638)
- #623 — loosen `open_url` scheme gate to an allowlist: http/https/mailto/tel/sms (PUSH-834)
- #633 — handle IAF `openExternalUrl` bridge event (PUSH-732)
- #654 — consolidate `formExternalUrlClicked` into `formCtaClicked`

## Test Plan
Each commit landed via its own reviewed/tested PR (see linked PRs above for individual test plans). This PR merges the integration branch as-is into `rel/5.4.0`.

## Related Issues/Tickets
PUSH-637, PUSH-638, PUSH-732, PUSH-834 — part of the "open web URL" project.

## Summary by CodeRabbit

* **New Features**
  * Added support for opening external URLs from push notification body taps, action buttons, and in-app form CTAs.
  * Supports approved schemes including HTTP(S), email, phone, and SMS links.
  * Deep-link actions continue to take precedence when both deep-link and external URL targets are provided.

* **Bug Fixes**
  * Unsafe or invalid URL schemes are now rejected before opening.
  * Improved handling of missing URLs and incomplete form CTA metadata.

* **Documentation**
[...]

PR #669, opened 2026-07-22
test(push): fix racy web_url/deep_link precedence tests

morrilltim:
# Description
Fixes an intermittent CI failure on the `web_url`/`deep_link` precedence tests, caused by a race between two independently-scheduled Tasks in the test, not a bug in the SDK's dispatch behavior.

## Due Diligence

## Changelog / Code Overview
PR #666 (`feat/push-638-open-url` → `rel/5.4.0`) hit a real but intermittent CI failure on `16.4, release`: `testHandleBodyTap_DeepLinkUnchangedWhenWebUrlAbsent` and `testHandleBodyTap_WebUrlDispatchesOpenWebUrl` failed on two separate reruns (once with 1 test failing, once with 2).

Root cause: `dispatchOnMainThread()` schedules the actual action dispatch (`klaviyoSwiftEnvironment.send(action)`) in its own `Task { await MainActor.run { ... } }`, while `handle(notificationResponse:withCompletionHandler:)` schedules the completion callback in a *separate* `Task { @MainActor in completionHandler() }`. Swift Concurrency gives no ordering guarantee between independently-created Tasks, so `wait(for: [callback], timeout: 1.0)` can return before the actual dispatch(es) have landed — especially under CI load, which is exactly what surfaced this.

Fix (test-only, no production code changed): each affected test now also waits on an `XCTestExpectation` with `expectedFulfillmentCount` set to the number of `send()` calls the scenario actually produces (verified against `handleActionButtonTap`/`resolveOpenAction` in `Klaviyo.swift`), fulfilled from inside the `send` closure itself. This makes the test wait for the actual side effects it asserts on, rather than an unrelated callback.

Affected tests: `testHandleBodyTap_WebUrlDispatchesOpenWebUrl`, `testHandleBodyTap_DeepLinkUnchangedWhenWebUrlAbsent`, `testHandleBodyTap_DeepLinkTakesPrecedenceOverWebUrl`, `testHandleActionButtonTap_OpenUrlButton`, `testHandleActionButtonTap_OpenUrlButtonWithBlockedSchemeDoesNotDispatch`.

## Test Plan
`make CONFIG=release test-library` — full suite (251 tests) passes, including all five updated tests. `swiftlint --strict` on the changed file shows only pre-existing violations (identical before/after this diff, confirmed via `git stash`), nothing newly introduced.

## Related Issues/Tickets
PUSH-638 — part of the "open web URL" project. Surfaced via flaky CI on #666.

### DIFF
--- a/Tests/KlaviyoSwiftTests/KlaviyoSDKTests.swift
+++ b/Tests/KlaviyoSwiftTests/KlaviyoSDKTests.swift
@@ -487,6 +487,11 @@ class KlaviyoSDKTests: XCTestCase {
 
     func testHandleBodyTap_WebUrlDispatchesOpenWebUrl() throws {
         let callback = XCTestExpectation(description: "callback is made")
+        // Each dispatched action (event track + openWebUrl) is scheduled on its own
+        // Task by `dispatchOnMainThread`, independently of the completion handler's Task,
+        // so waiting on `callback` alone races with those dispatches landing.
+        let dispatched = XCTestExpectation(description: "actions dispatched")
+        dispatched.expectedFulfillmentCount = 2
         let webURL = try XCTUnwrap(URL(string: "https://example.com/sale"))
 
         let userInfo: [AnyHashable: Any] = [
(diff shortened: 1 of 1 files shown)


glenn-klaviyo (approved):
LGTM

PR #675, opened 2026-07-24
fix: replace AsyncStream-based handshake wait with a resumed continuation

morrilltim:
## Summary
- Fixes a real hang in `IAFWebViewModel.establishHandshake(timeout:)`, which is what's been causing the "16.4, release" CI job to hang for 1-2 hours (twice, on both PR #673 and PR #674) instead of failing fast.
- Replaces the `AsyncStream<Void>`-backed handshake signal with a single one-shot `CheckedContinuation<Void, Error>`.

## Why
`establishHandshake` raced two tasks inside `withThrowingTaskGroup`: the real wait (`handshakeStream.first { _ in true }`) against a `Task.sleep`-based timeout. Structured concurrency guarantees a task group can't return until *every* child task finishes — including a "losing", cancelled one. Cancelling the operation task only requests cancellation; it never actually caused the suspended `AsyncStream.first(where:)` call to resolve, since nothing ever yielded into or finished that stream. Once the timeout fired with no handshake having happened, the whole call — and the test/host process running it — hung forever instead of throwing `TimeoutError.timeout`.

This is exactly the scenario `IAFWebViewModelPreloadingTests.testPreloadWebsiteNoActionTimeout` exercises (a mock delegate that never fires a handshake event), and it's what's been intermittently hanging CI:
- Confirmed via direct log inspection of the cancelled CI job: the last thing printed before the ~1h50m stall was `Test Case '...testPreloadWebsiteNoActionTimeout' started.`
- Confirmed this is pre-existing/untouched by any in-flight PR: `git diff origin/rel/5.4.0 -- Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift Tests/KlaviyoFormsTests/IAFWebViewModelPreloadingTests.swift` is empty against the base this branches from.
- Reproduced the underlying mechanism standalone (plain Swift script + a `@MainActor`-isolated variant) with an unyielded `AsyncStream` raced the same way — cancellation resolved it fine on the local/current Swift runtime, meaning this is likely a runtime-version-specific gap in how reliably an in-flight `AsyncStream.next()` responds to cancellation (matches the observed pattern of only `16.4` ever hanging, never `15.2`/`15.4`). Rather than depend on that behavior at all, the fix removes the dependency entirely.

## Fix
Both the success path (`.handShook` bridge event) and the timeout path now explicitly resume the *same* stored continuation, whichever happens first; the loser is a guarded no-op. No cancellation propagation is required for either side to unblock the other.

## Test plan
- [x] `IAFWebViewModelPreloadingTests` (all 3) and the full `KlaviyoFormsTests` suite (61 tests) pass locally
- [x] Ran `testPreloadWebsiteNoActionTimeout` 11 times back-to-back — consistently resolves in ~0.1-0.4s, no hangs
- [x] `swiftlint --strict` violation count/set on the changed file matches base exactly (no new violations)
- [ ] CI green (this is what should stop the "16.4, release" job from hanging)

### DIFF
--- a/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
+++ b/Sources/KlaviyoForms/InAppForms/IAFWebViewModel.swift
@@ -39,16 +39,50 @@ class IAFWebViewModel: KlaviyoWebViewModeling {
     /// continuation. Deliberately doesn't require `@MainActor` to touch: the timeout side resumes
     /// it from a plain, non-actor-isolated `Task`, so unblocking on timeout never needs the main
     /// actor to schedule more work onto itself while `establishHandshake` is already suspended on it.
+    ///
+    /// Tracks a generation token so a stale timeout task from a superseded call can never resume
+    /// a newer one's continuation -- it only fires if its own token still matches the box's
+    /// current generation.
     private final class HandshakeContinuationBox: @unchecked Sendable {
         private let lock = NSLock()
         private var continuation: HandshakeContinuation?
+        private var generation = 0
 
-        func exchange(_ newValue: HandshakeContinuation?) -> HandshakeContinuation? {
+        /// Stores `newValue`, returning a token for the timeout task to check back in with.
+        /// Any still-pending prior continuation is resumed with `error` rather than dropped.
+        func store(_ newValue: HandshakeContinuation, supersedingWith error: Error) -> Int {
             lock.lock()
-            defer { lock.unlock() }
             let previous = continuation
             continuation = newValue
-            return previous
+            generation += 1
+            let token = generation
+            lock.unlock()
+            previous?.resume(throwing: error)
+            return token
+        }
+
+        /// Resumes the pending continuation with `error`, but only if `token` still matches the
+        /// most recently stored continuation's generation -- otherwise a newer call has already
+        /// superseded it and this is a no-op.
+        func resumeIfCurrent(token: Int, throwing error: Error) {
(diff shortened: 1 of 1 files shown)


morrilltim:
Update: the first version of this fix still hung in CI at the exact same spot (\`testPreloadWebsiteNoActionTimeout\`, confirmed via job log before cancelling). That ruled out AsyncStream specifics as the sole cause -- the common thread between the old and first-attempt code was that both created work needing to hop onto \`@MainActor\` while \`establishHandshake\` (already on the main actor) was suspended waiting on it, which looks like a main-actor reentrancy issue on top of a suspected iOS 16.4 runtime quirk (I don't have a 16.4 simulator locally to confirm directly -- only 18.6/26.5 are available).

Pushed a second revision that removes main-actor involvement from the race entirely: the pending continuation now lives in a lock-guarded, non-actor-isolated box, and both the timeout path and the \`.handShook\` success path resume it directly without needing to schedule anything back onto the main actor. Rebuilt, all 61 \`KlaviyoFormsTests\` pass, and \`testPreloadWebsiteNoActionTimeout\` ran cleanly 10x back-to-back locally. Watching CI on this revision now.

morrilltim:
CI is green across the board on this revision, including `16.4, release` (6m30s, no hang). This confirms the actor-agnostic redesign fixed it.

morrilltim:
Closing -- someone else is already working on this fix.

PR #688, opened 2026-07-30
fix(KlaviyoSwift): build Store.production on main during initialize

belleklaviyo:
# Description
Fixes `MAGE-1007`: `KlaviyoSDK.initialize(with:)` built the main-thread-only `Store.production` off-main when a host initialized from a background queue or TCA reducer effect, emitting the TCA "store initialized on a non-main thread" runtime warning.

## Due Diligence

## Changelog / Code Overview
`initialize(with:)` ran `SharedStoreMirror.setup()` synchronously on the caller's thread. `setup()` touches `klaviyoSwiftEnvironment.statePublisher()`, forcing the lazy global `Store.production` into existence — and installing the mirror's Combine sink — on whatever thread got there first. Off-main callers tripped `Store`'s main-thread-only invariant.

- **`Klaviyo.swift`** — wrap only `SharedStoreMirror.setup()` in `Task { @MainActor in … }`. The `injectNotificationDelegate()` call site is intentionally left untouched so the notification delegate still installs synchronously before `initialize` returns on the main path (it self-manages its own threading). The mirror is order-insensitive: `statePublisher` is a current-value publisher gated by `.filter(initialized)`, so it observes the initialized state whether it subscribes before or after the `.initialize` action lands.
- **`KlaviyoSDKInitializeThreadingTests.swift`** — dropped the previous `injectOnMain` assertion (it stubbed away the very threading logic it claimed to test, so it could never pass against the correct fix); retitled the background-thread test around the store invariant it actually guards; added a main-thread test that exercises the **real** production injection closure and asserts the delegate is installed synchronously before `initialize` returns.

### Known residual (out of scope for this ticket)
The reviewer flagged that on a true cold start from a background thread, the synchronous `injectNotificationDelegate()` read of the `klaviyoSwiftEnvironment` global could itself be the first touch that lazily builds `Store.production` off-main. This is **pre-existing** (not introduced or worsened here) and is a broader lazy-global-initialization concern. `MAGE-894` (promote KlaviyoCore stores to canonical source of truth; remove `SharedStoreMirror`) is likely to delete this path entirely — worth sequencing the two rather than layering another workaround here.

## Test Plan
- `xcodebuild test -scheme klaviyo-swift-sdk-Package -destination "platform=iOS Simulator,name=iPhone 17 Pro" -only-testing:KlaviyoSwiftTests/KlaviyoSDKInitializeThreadingTests` → both tests pass.
- Manual repro (per ticket): `klaviyo-swift-test-app` initializes from a TCA `Effect.run` (background cooperative thread). Before: TCA warning fires during `didFinishLaunching`. After: no warning.

## Related Issues/Tickets
Resolves MAGE-1007
Related to MAGE-894

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SDK initialization reliability by ensuring shared store mirror setup runs on the main thread even when initialization starts from a background thread.
  * Ensured initialization dispatches consistently on the main thread for correct store behavior.
  * Improved state handling for late subscriptions so existing initialization state is reflected immediately.
* **Tests**
  * Added regression tests covering main-thread correctness during initialization.
  * Added tests verifying notification delegate installation is synchronous on supported iOS versions.
  * Added coverage for late subscribers picking up already-initialized state.

### DIFF
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -92,7 +92,8 @@ public struct KlaviyoSDK {
     /// - Returns: a KlaviyoSDK instance
     @discardableResult
     public func initialize(with apiKey: String) -> KlaviyoSDK {
-        SharedStoreMirror.setup()
+        // Hop to main: `setup()` touches `Store.production`, which is main-thread-only.
+        Task { @MainActor in SharedStoreMirror.setup() }
         dispatchOnMainThread(action: .initialize(apiKey))
         klaviyoSwiftEnvironment.injectNotificationDelegate()
         return self
--- a/Sources/KlaviyoSwift/Klaviyo.swift
+++ b/Sources/KlaviyoSwift/Klaviyo.swift
@@ -92,9 +92,10 @@ public struct KlaviyoSDK {
     /// - Returns: a KlaviyoSDK instance
     @discardableResult
     public func initialize(with apiKey: String) -> KlaviyoSDK {
-        // Hop to main: `setup()` touches `Store.production`, which is main-thread-only.
-        Task { @MainActor in SharedStoreMirror.setup() }
-        dispatchOnMainThread(action: .initialize(apiKey))
+        Task {
+            @MainActor in SharedStoreMirror.setup()
+            _ = klaviyoSwiftEnvironment.send(.initialize(apiKey))
+        }
         klaviyoSwiftEnvironment.injectNotificationDelegate()
         return self
     }
--- a/Tests/KlaviyoSwiftTests/KlaviyoSDKInitializeThreadingTests.swift
+++ b/Tests/KlaviyoSwiftTests/KlaviyoSDKInitializeThreadingTests.swift
@@ -16,10 +16,12 @@ import XCTest
 /// during `initialize` must be routed to the main thread regardless of the caller's thread.
 @MainActor
 final class KlaviyoSDKInitializeThreadingTests: XCTestCase {
+    private var savedCoreEnvironment: KlaviyoEnvironment!
     private var savedEnvironment: KlaviyoSwiftEnvironment!
 
     override func setUp() {
         super.setUp()
+        savedCoreEnvironment = environment
         savedEnvironment = klaviyoSwiftEnvironment
         environment = KlaviyoEnvironment.test()
         klaviyoSwiftEnvironment = KlaviyoSwiftEnvironment.test()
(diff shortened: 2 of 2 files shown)


belleklaviyo:
bugbot run

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -92,7 +92,8 @@ public struct KlaviyoSDK {
     /// - Returns: a KlaviyoSDK instance
     @discardableResult
     public func initialize(with apiKey: String) -> KlaviyoSDK {
-        SharedStoreMirror.setup()
+        // Hop to main: `setup()` touches `Store.production`, which is main-thread-only.
+        Task { @MainActor in SharedStoreMirror.setup() }
```
ab1470: are we certain we're not introducing a timing issue by dispatching the `setup()` to a task? With this change, lines 97 & 98 could complete before the `setup()` task on line 96 completes

aahiltn (commented):
Thanks for the fix, runs without the warning this time! Good to me, but dependent on the iOS devs ofc

Sources/KlaviyoSwift/Klaviyo.swift
ab1470: Ok I asked Claude to evaluate my own comment lol, here's what it said:

> What the reviewer flagged: with Task { @MainActor in SharedStoreMirror.setup() } on line 96, lines 97 (dispatchOnMainThread(action: .initialize(apiKey))) and 98 (injectNotificationDelegate()) could run/complete before that task finishes, so the mirror's Combine subscription might not be attached before the .initialize action fires.
>
> Why it doesn't matter:
>
>1. Line 97 isn't actually sequential either. dispatchOnMainThread (Klaviyo.swift:14-18) wraps klaviyoSwiftEnvironment.send(action) in its own separate Task { await MainActor.run { ... } }. So both line 96 and line 97 are independently-scheduled hops to the main actor — there's no guarantee that one runs before the other regardless of this diff.
>2. Line 98 is irrelevant to the race. injectNotificationDelegate() just sets a UNUserNotificationCenter delegate synchronously — it never touches KlaviyoState or the stores SharedStoreMirror mirrors into, so its ordering relative to setup() has no effect.
>3. The actual safety net: statePublisher() is backed by a CurrentValueSubject, not a PassthroughSubject. Store.state is declared as CurrentValueSubject<State, Never> (Vendor/ComposableArchitecture/Store.swift:163), and KlaviyoSwiftEnvironment.statePublisher just erases that. CurrentValueSubject replays its current value to any new subscriber at subscribe time. So no matter which task wins the race:
>  - If setup() subscribes first, it's live and will receive the .initialized state whenever .initialize processes.
>  - If .initialize processes first and the state is already .initialized by the time setup() subscribes, the subscription immediately receives that current value on attach.
>
>Either ordering results in the mirror observing the initialized state exactly once. The .filter { $0.initalizationState == .initialized } + .removeDuplicates combo makes this ordering-independent by design.
>
>So the comment is asking the right question, but the answer is "no, because CurrentValueSubject semantics make this immune to that race" — worth replying with that explanation rather than serializing the two tasks. One gap: the new KlaviyoSDKInitializeThreadingTests only assert both hops land on main thread — neither actually exercises the "setup() subscribes late" ordering to lock in this guarantee as a regression test. Might be worth a follow-up test that fires .initialize and lets it fully process before SharedStoreMirror.setup() runs, then asserts the mirror still picks up the state.

ab1470 (approved):

PR #692, opened 2026-08-03
fix(state): avoid Int(infinity) trap in .flushQueue while offline [MAGE-1023]

evan-masseau:
# Description

`Int(Double.infinity)` traps in Swift. `.flushQueue` could reach that conversion while the SDK was backing off with no connectivity, hard-crashing the host app (a `fatalError`-class precondition failure, not a catchable error). Guards `.flushQueue` on a finite flush interval. Closes MAGE-1023.

## Due Diligence

No public API change; crash fix only. Targeting `rel/5.4.0`.

## Changelog / Code Overview

Three pre-existing pieces combined into a crash:

1. `.networkConnectivityChanged(.notReachable)` parks `state.flushInterval` at `Double.infinity` and cancels the flush timer.
2. `.flushQueue`'s backoff countdown computes `max(backOff - Int(state.flushInterval), 0)` — `Int(.infinity)` traps.
3. The priority path (opened push / geofence) dispatches `.flushQueue` **unconditionally**, so the cancelled timer doesn't save us. `.flushQueue`'s existing guards don't stop it either: `.initialized` passes, and `state.flushing` is `false` because `.requestFailed` cleared it.

The fix bails out of `.flushQueue` when `flushInterval` isn't finite:

```swift
guard state.flushInterval.isFinite else {
    return .none
}
```

Chosen over `Int(clamping:)` because it's the more correct semantic — there's no point draining a queue with no network, and clamping would let the retry proceed offline, burning attempts on requests that cannot succeed. It also keeps any *future* arithmetic on `flushInterval` safe, and `isFinite` covers NaN as well as infinity. The backoff is left untouched, so it resumes when connectivity returns.

**Not a 5.4.0 regression** — all three pieces exist unchanged on `master`. What this release changes is how *reachable* precondition 1 is: the retryable set went from `[429, 500, 502, 503, 504]` to 429 plus all of 500–599, and edge failures during an incident correlate strongly with degraded client connectivity. So the scenario the RCA work targeted also arms this crash. `master` picks this up automatically via the standard release merge-forward when `rel/5.4.0` lands — `rel/*` is always merged back, so no separate port is needed.

### Two deliberate side effects of the guard's placement

It sits above the `retryWithBackoff` block, so it stops more than the crashing path. Both are intended and both are now documented:

1. **The non-backoff `.retry` path also stops draining while offline.** That path never crashed, but draining with no network only burns attempts and inflates `X-Klaviyo-Attempt-Count` on requests that cannot succeed. Pinned by `testFlushQueueWhileOfflineWithoutBackoffDoesNotDrainQueue` — without it, someone could later move the guard down inside the backoff block, still fix the crash, still pass the suite, and silently revert this half.
2. **A pending profile is deferred, not lost, while offline.** The guard also precedes `if state.pendingProfile != nil { state.enqueueProfileOrTokenRequest() }`, so a pending profile no longer converts into a queued request until the next online `.flushQueue`. Verified non-destructive: `pendingProfile` is a `[Profile.ProfileKey: AnyEncodable]` dictionary that `setProfileProperty` merges into per key, so successive offline updates accumulate (last write wins per property) rather than clobbering each other.

## Test Plan

New reducer test, `testFlushQueueWhileOfflineDuringBackoffDoesNotTrap`, following the documented repro: `retryState` in `.retryWithBackoff` -> `.networkConnectivityChanged(.notReachable)` -> `.flushQueue`.

**Verified the test is not vacuous** — with the guard stashed it reproduces the exact reported trap and the test process dies:

```
Fatal error: Double value cannot be converted to Int because it is either infinite or NaN
** TEST FAILED **
```

With the guard: full package suite green, **296 tests, 0 failures**. SwiftLint/SwiftFormat clean.

One ordering subtlety worth noting for reviewers: `INITIALIZED_TEST_STATE()` has `flushing: true`, which would make `.flushQueue` bail at the earlier `if state.flushing` check and render the test vacuous. The test asserts through `.cancelInFlightRequests` (which clears `flushing`) first, so it genuinely reaches the conversion.

Not manually re-repro'd on a device — the reporter confirmed the device repro steps, and the reducer test covers the same path deterministically.

*Unchecked box:* team test docs. The Q2 manual UAT plan already exercises offline-during-backoff generally; a targeted "offline + backoff + tap a push" row is worth adding to the test sheet, but I didn't want to bundle a shared-doc edit into this fix.

## Related Issues/Tickets

- Closes MAGE-1023
- Context: project "SDK Networking RCA Items"; exposure increased by the widened 5xx retry range in the 5.4.0 rollup (#667).
- The absolute-deadline refactor tracked in the related parity ticket removes this `Int(Double)` conversion entirely and would supersede this guard.

-- Claude

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when connectivity is lost during retry backoff.
  - Queue flushing now safely pauses offline instead of triggering errors or draining queued requests.
  - Preserved retry timing and state while the device remains offline.

- **Tests**
  - Added coverage for offline queue flushing with and without retry backoff.
  - Verified queued requests remain intact until connectivity is restored.

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -285,8 +285,8 @@ struct KlaviyoReducer: ReducerProtocol {
             if state.flushing {
                 return .none
             }
-            // Offline parks `flushInterval` at `.infinity`, which the backoff below would trap
-            // converting to `Int`. No point draining the queue with no network anyway.
+            // The priority path can dispatch `.flushQueue` while offline, where `flushInterval` is
+            // `.infinity` — the backoff below would trap on `Int()`, and draining is pointless.
             guard state.flushInterval.isFinite else {
                 return .none
             }
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -285,11 +285,8 @@ struct KlaviyoReducer: ReducerProtocol {
             if state.flushing {
                 return .none
             }
-            // Offline parks `flushInterval` at `.infinity` (see `.networkConnectivityChanged`),
-            // which also cancels the flush timer — but the priority path (opened push, geofence)
-            // dispatches `.flushQueue` unconditionally, so we can still land here while offline.
-            // Bail out: there is nothing to gain from draining the queue with no network, and it
-            // keeps a non-finite interval out of the `Int` conversion below, which would trap.
+            // Offline parks `flushInterval` at `.infinity`, which the backoff below would trap
+            // converting to `Int`. No point draining the queue with no network anyway.
             guard state.flushInterval.isFinite else {
                 return .none
             }
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -285,6 +285,14 @@ struct KlaviyoReducer: ReducerProtocol {
             if state.flushing {
                 return .none
             }
+            // Offline parks `flushInterval` at `.infinity` (see `.networkConnectivityChanged`),
+            // which also cancels the flush timer — but the priority path (opened push, geofence)
+            // dispatches `.flushQueue` unconditionally, so we can still land here while offline.
+            // Bail out: there is nothing to gain from draining the queue with no network, and it
+            // keeps a non-finite interval out of the `Int` conversion below, which would trap.
+            guard state.flushInterval.isFinite else {
+                return .none
+            }
             if case let .retryWithBackoff(requestCount, totalCount, backOff) = state.retryState {
                 let newBackOff = max(backOff - Int(state.flushInterval), 0)
                 if newBackOff > 0 {
--- a/Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
+++ b/Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
@@ -500,6 +500,32 @@ class StateManagementEdgeCaseTests: XCTestCase {
         _ = await store.send(.flushQueue)
     }
 
+    @MainActor
+    func testFlushQueueWhileOfflineWithoutBackoffDoesNotDrainQueue() async {
+        // The guard sits above the backoff block, so it also stops the non-backoff `.retry` path —
+        // which never crashed, but draining while offline only burns attempts. Pins that half:
+        // the queue must stay put rather than moving into requestsInFlight.
+        var initialState = INITIALIZED_TEST_STATE()
+        initialState.retryState = .retry(1)
+        let request = initialState.buildProfileRequest(
+            apiKey: initialState.apiKey!,
+            anonymousId: initialState.anonymousId!
+        )
+        initialState.queue = [request]
+        let store = TestStore(initialState: initialState, reducer: KlaviyoReducer())
+
+        _ = await store.send(.networkConnectivityChanged(.notReachable)) {
+            $0.flushInterval = Double.infinity
+        }
+        _ = await store.receive(.cancelInFlightRequests) {
+            $0.flushing = false
+        }
+
+        // Queue is non-empty on purpose — with an empty queue `.flushQueue` returns early anyway
+        // and this test would pass with or without the guard.
+        _ = await store.send(.flushQueue)
+    }
+
     // MARK: - Missing api key for token request
 
     @MainActor
--- a/Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
+++ b/Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
@@ -474,6 +474,37 @@ class StateManagementEdgeCaseTests: XCTestCase {
         _ = await store.send(.networkConnectivityChanged(.reachableViaWWAN))
     }
 
+    // MARK: - Flush queue while offline during retry backoff
+
+    @MainActor
+    func testFlushQueueWhileOfflineDuringBackoffDoesNotTrap() async {
+        // Going offline parks `flushInterval` at `.infinity` and cancels the flush timer — but the
+        // priority path (opened push / geofence) dispatches `.flushQueue` unconditionally, so we can
+        // still land in `.flushQueue` while offline. With `retryState` in backoff, the countdown
+        // computes `backOff - Int(state.flushInterval)`, and `Int(Double.infinity)` traps in Swift,
+        // taking down the host app. `.flushQueue` must no-op instead, leaving the backoff intact so
+        // it resumes when connectivity returns.
+        var initialState = INITIALIZED_TEST_STATE()
+        initialState.retryState = .retryWithBackoff(
+            requestCount: 1,
+            totalRetryCount: 1,
+            currentBackoff: 30
+        )
+        let store = TestStore(initialState: initialState, reducer: KlaviyoReducer())
+
+        _ = await store.send(.networkConnectivityChanged(.notReachable)) {
+            $0.flushInterval = Double.infinity
+        }
+        // Also clears `flushing`, so the `if state.flushing` bail-out below does NOT mask the
+        // conversion — without the guard this test reaches the trap.
+        _ = await store.receive(.cancelInFlightRequests) {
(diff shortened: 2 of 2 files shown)


Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -285,6 +285,14 @@ struct KlaviyoReducer: ReducerProtocol {
             if state.flushing {
                 return .none
             }
+            // Offline parks `flushInterval` at `.infinity` (see `.networkConnectivityChanged`),
+            // which also cancels the flush timer — but the priority path (opened push, geofence)
+            // dispatches `.flushQueue` unconditionally, so we can still land here while offline.
+            // Bail out: there is nothing to gain from draining the queue with no network, and it
+            // keeps a non-finite interval out of the `Int` conversion below, which would trap.
+            guard state.flushInterval.isFinite else {
```
ndurell, line 290: The flush interval will get reset when the device comes back online right? Also do we need these giant comments in here?

Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
```diff
@@ -474,6 +474,37 @@ class StateManagementEdgeCaseTests: XCTestCase {
...
     }
 
+    // MARK: - Flush queue while offline during retry backoff
+
+    @MainActor
+    func testFlushQueueWhileOfflineDuringBackoffDoesNotTrap() async {
+        // Going offline parks `flushInterval` at `.infinity` and cancels the flush timer — but the
+        // priority path (opened push / geofence) dispatches `.flushQueue` unconditionally, so we can
+        // still land in `.flushQueue` while offline. With `retryState` in backoff, the countdown
+        // computes `backOff - Int(state.flushInterval)`, and `Int(Double.infinity)` traps in Swift,
```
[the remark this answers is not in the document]
evan-masseau: Declining — the measurement is off. That line is **100 characters**, not 111:

```
$ awk "NR==484 {print length(\$0)}" Tests/KlaviyoSwiftTests/StateManagementEdgeCaseTests.swift
100
```

`.swiftlint.yml` sets `line_length: 110`, and the longest line this PR adds is 101 chars — which is why SwiftLint passes both in the pre-commit hook and in CI. No violation to fix.

The comment did get shortened in 822f5518, but for an unrelated reason (reviewer asked for less verbosity), not because it breached the limit.

-Claude

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
evan-masseau: Yes on both counts.

**Flush interval reset:** confirmed — when connectivity returns, `.networkConnectivityChanged` sets `flushInterval` back to a finite constant and immediately schedules a flush:

```swift
case .reachableViaWiFi:
    state.flushInterval = StateManagementConstants.wifiFlushInterval   // line 442
case .reachableViaWWAN:
    state.flushInterval = StateManagementConstants.cellularFlushInterval // line 444
}
return environment.timer(state.flushInterval)      // schedules .flushQueue
```

So the guard only holds while genuinely offline, and `retryState` is deliberately left untouched so the backoff resumes rather than restarting. That is also why I went with the guard instead of `Int(clamping:)` — clamping computes `max(backOff - Int.max, 0) == 0`, which would drop the backoff and let the retry fire while still offline, burning attempts on requests that cannot succeed.

**Comments:** fair — trimmed in 822f5518. The source comment went 5 lines to 2, and the test comment 6 to 2. I kept one short line noting that `.cancelInFlightRequests` clears `flushing`, because without it `.flushQueue` bails at the earlier `if state.flushing` check and the test silently passes with or without the fix. Happy to cut that one too if you would rather it just be terse.

-Claude

evan-masseau:
Re-reviewed at `822f5518`. The comment trim reads well, and challenging the line-length finding was the right call — CodeRabbit withdrew it.

> **Edited after verification.** Items 1 and 3 held up and are addressed. Items 2 and 4 were **wrong** on the facts and are struck through below with corrections, so the record isn't misleading.

### 1. Add the non-backoff offline case (medium) — ✅ done

The guard sits above the `retryWithBackoff` block, so `.flushQueue` now no-ops for *any* `retryState` when offline — including `.retry`, a path that never crashed. Previously it drained the queue and attempted `.sendRequest`; now it bails.

That change is good (no burnt attempts, no spurious `X-Klaviyo-Attempt-Count` increments on requests that cannot succeed) and it matches the stated "no point draining with no network" rationale. But it's a deliberate behavior change in a path the crash never touched, and nothing pins it.

Addressed in `bcebcb18` via `testFlushQueueWhileOfflineWithoutBackoffDoesNotDrainQueue`. The queue is seeded non-empty on purpose — with an empty queue `.flushQueue` returns early anyway and the test would pass with or without the guard. Verified it fails with the guard removed ("State was not expected to change" + 2 unexpected actions).

### 2. ~~Track the `master` fix (medium)~~ — ❌ retracted, premise was wrong

~~The body says `master` "needs the same fix". There's no ticket for it. Until 5.4.0 merges, `master` carries a live crash, and that window widens if the release slips.~~

**Correction:** this repo always merges `rel/*` back to `master` after release, so the fix reaches `master` through the normal merge-forward — there is no separate port to track and no ticket needed. The PR body now states that definitively instead of hedging with a parenthetical. (I filed MAGE-1024 for this and have since cancelled it as unnecessary.)

### 3. Document the `pendingProfile` deferral (low) — ✅ done

The guard also precedes:

```swift
if state.pendingProfile != nil {
    state.enqueueProfileOrTokenRequest()
}
```

So while offline, a pending profile no longer converts into a queued request. Documented in the PR body, along with the confirmation that was asked for: `pendingProfile` is a `[Profile.ProfileKey: AnyEncodable]` dictionary and `setProfileProperty` merges into it per key, so successive offline updates accumulate (last write wins per property) rather than clobbering each other.

### 4. ~~Optional: make the sentinel self-healing (low)~~ — ❌ mostly wrong, declined

~~If `startNotifier()` throws after a `.notReachable`, the recovery signal is lost and the guard turns the crash into a queue that stalls silently and **permanently** — which then hits the 200-cap eviction and discards data invisibly. Arguably worse than the crash. `environment.reachabilityStatus()` already exists, so foreground could re-derive `flushInterval`.~~

**Corrections, all verified in code:**

* **"Permanently" is false.** `flushInterval` is not in `KlaviyoState.CodingKeys` — it is **not persisted** and defaults to the finite `wifiFlushInterval`. Any stall is bounded to the app session and self-heals on the next cold launch. The queue *is* persisted, so events survive to that launch and drain then.
* **The self-healing already exists.** `startNotifier()` performs an explicit initial check (`ReachabilitySwift.swift:181-184` → `reachabilityChanged()`), and `stopNotifier()` (189-195) deliberately does not touch `previousFlags`. So after `.notReachable` parked `.infinity`, `previousFlags` still reads unreachable — on the next foreground with network back, flags differ, the notification fires, and `flushInterval` resets. Every foreground already re-derives status; adding a manual `reachabilityStatus()` read would be redundant.
* **The residual hole is pathological.** `startNotifier()` can only throw from `SCNetworkReachabilitySetCallback` / `SCNetworkReachabilitySetDispatchQueue` returning false (lines 171-179) — SystemConfiguration-level failures that would break reachability SDK-wide regardless of this guard. Losing data needs that *and* a preceding offline event *and* >200 events in the same session *and* no relaunch.
* **The guard didn't create this.** The stuck sentinel is pre-existing; it previously produced a crash (with backoff) or futile sends into no network (without). The symptom changed, not the condition.

Not worth fixing now, and pointedly not on a release branch, where touching reachability/foreground plumbing carries more risk than the failure it would prevent. The absolute-deadline refactor removes the `Int(Double)` conversion entirely and is the durable fix.

---

For the record on the reset question, since it's the obvious worry with this guard: it holds, for the `previousFlags` reason in item 4 above.

One nit, take it or leave it: the trimmed source comment dropped the reason we can reach `.flushQueue` at all while offline (the priority path bypassing the cancelled timer). That's the non-obvious half, and the source-side reader is the one who'd otherwise delete the guard. — restored in `bcebcb18`, still within two lines.

-- Claude

evan-masseau:
bugbot run

evan-masseau:
@coderabbitai review

belleklaviyo (approved):
thanks for catching this and finding a simple fix, lgtm

PR #707, opened 2026-08-27
fix(KlaviyoCore): restore automatic push-token dedup in the durable pre-init buffer (MAGE-1137)

glenn-klaviyo:
# Description
Restores dedup for repeated pre-init push-token buffering, lost when MAGE-952 replaced the old in-memory `pendingRequests` (which deduped automatic-only fires) with the durable `UnattributedBuffer` (a plain FIFO append with no dedup for any request type). `setAutomaticPushToken` can fire before `initialize()`, and each fire was buffering a separate, redundant `.pushToken` register request.

## Due Diligence

## Changelog / Code Overview
`UnattributedBuffer.append` now coalesces any repeated buffered `.pushToken` entry to the latest, rather than accumulating one per pre-init fire. Deliberately **not** provenance-aware (manual vs. automatic) — register-push-token is a last-write-wins upsert, and nothing else in the SDK distinguishes manual from automatic tokens post-init either, so coalescing applies uniformly. This is a narrow, intentional divergence from the pre-MAGE-952 behavior (which only deduped automatic fires); see the design discussion on the linked ticket.

A second commit closes a gap CodeRabbit caught in review: `hydrateIfNeeded()` only inherited whatever was on disk without normalizing it, so a buffer file written by a pre-fix build could still resurrect duplicate tokens on next launch. Both `append` and hydration replay now share the same coalesce+cap logic (`addEntry`).

## Test Plan
New coverage in `UnattributedBufferTests` and `RequestEnqueuerTests`:
- repeated `.pushToken` appends coalesce to the latest
- coalescing doesn't affect other buffered request types or trigger spurious cap-eviction
- a coalesce landing between `drainSnapshot()`/`removeDrained()` survives (doesn't get silently dropped)
- a persisted file with duplicate tokens is normalized on hydration
- full-path `RequestEnqueuer.enqueuePushToken` coalescing

Full `KlaviyoCoreTests` bundle (276 tests) green; SwiftLint/SwiftFormat clean.

## Related Issues/Tickets
MAGE-1137

## Summary by CodeRabbit

* **Bug Fixes**
  * Repeated push-token updates are now consolidated, retaining only the latest token.
  * Buffered requests preserve unrelated entries and avoid unnecessary eviction when capacity is reached.
  * Persisted duplicate entries are normalized when restored.
  * Requests added while buffered items are being processed are now retained correctly.
  * Buffer behavior is now applied consistently to restored and newly added requests.

### DIFF
--- a/Sources/KlaviyoCore/UnattributedBuffer.swift
+++ b/Sources/KlaviyoCore/UnattributedBuffer.swift
@@ -81,11 +81,9 @@ final class UnattributedBuffer {
         }
     }
 
-    /// A push-token registration is a snapshot of "the current token state" — a repeated pre-init
-    /// fire (e.g. multiple automatic APNs callbacks before `initialize()`) makes any earlier buffered
-    /// `.pushToken` redundant, so it's dropped in favor of the newest one rather than both being
-    /// drained and sent. Not provenance-aware (manual vs. automatic) — coalescing applies regardless
-    /// of which caller produced the request, since both build the same idempotent payload shape.
+    /// Repeated pre-init token buffering (e.g. multiple automatic APNs callbacks before
+    /// `initialize()`) drops any earlier buffered token — only the latest is kept, whether it came
+    /// from a manual or automatic call.
     func append(_ request: UnattributedRequest) {
         lock.withLock {
             hydrateIfNeeded()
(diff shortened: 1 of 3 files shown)


belleklaviyo (approved):
thanks for restoring this in the new structure! looks good to me

PR #714, opened 2026-09-02
fix(KlaviyoSwift): rebind push token on pre-init identity change (MAGE-1165)

belleklaviyo:
Note: this is all edge case handling to protect against scenarios where devs have integrated our SDK in some unadvised way (ordering initialize, any calls before initialize, etc.)

# Description

**The problem it fixes.** When an app identifies a *different* user via `set(profile:)` **before** calling `initialize()` — account-switch apps that initialize the SDK late (login gate, consent gate, wrapper-SDK timing) rather than in `didFinishLaunching` — the device's APNs push token stayed registered to the **previous** user. That device kept receiving the old user's notifications and the new user got none on it, until some unrelated change (reinstall, OS/app update, permission toggle) happened to force a re-register. It's also a regression versus pre-modularization behavior.

**Why it happened.** Pre-init API intents are held in the durable `UnattributedBuffer` so nothing is lost if the app is killed before init — but they're stored as **pre-built payloads**, which freeze the identity at enqueue time. So a token registration and a profile switch became two independent frozen payloads, and the token was frozen to the old identity. The **initialized** path avoids this by re-registering the existing token against the new identity; the pre-init path had lost that rebind. (Same seam as the broader `pendingRequests` → `UnattributedBuffer` migration: the buffer freezes payloads instead of replaying intents through the reducer.)

## Due Diligence

- [x] This is planned work for an upcoming release. (Part of the iOS modularization line; targets `feat/modularization-2`.)

## Changelog / Code Overview

On a pre-init identity change where a push token is registered, re-register the token against the new identity so it doesn't strand on the old one.

1. **Reducer** (`enqueueProfile`, pre-init branch) — reads the persisted token from `IdentityStore` **before** the identity reset, then buffers **two separate entries**: a bare `.profile` (carrying the new identity + structured attributes) and, if a token exists, an identity-only `.pushToken` registration. Because the identity was just updated, the token registers to the **new** identity; a later token callback coalesces to that same identity.

   Why separate rather than one combined token+profile payload: a combined payload would live in a `.pushToken` buffer entry, and `.pushToken` entries are coalesced last-write-wins (token dedup). A later pre-init token registration — e.g. an **async automatic APNs callback**, which commonly fires around `set(profile:)` — would then replace the combined entry and silently drop the profile's name/location/custom properties. Keeping the profile in its own `.profile` entry makes it immune to that coalescing.

2. **Canonical token state** — pre-init `setPushToken` now persists the token to `IdentityStore` (canonical device state), so a token set **same-session, before** the switch is visible to the rebind above. Without this the token would live only as a buffered registration and the rebind couldn't see it.

3. **Supporting Core change** — `RequestFactory.tokenPayload(pushToken:enablement:background:profile:)` (a token payload embedding a full profile); the identity-only `tokenPayload(identity:…)` overload delegates to it.

The pre-init `resetProfile` path already rebinds correctly and is intentionally untouched.

## Test Plan

Regression + unit tests (all green; `xcodebuild test … -only-testing:KlaviyoCoreTests -only-testing:KlaviyoSwiftTests`):
- `RequestFactoryTests` — `tokenPayload(…profile:)` embeds the given profile.
- `StateManagementTests`:
  - `testPreInitProfileSwitchRebindsTokenToNewIdentity` — token persisted from a prior session + pre-init identity switch → a `.profile` and an identity-only `.pushToken`, both for the new identity; drain → one register bound to the new identity.
  - `testPreInitTokenSetBeforeProfileSwitchRebindsToNewIdentity` — token set pre-init this session (buffered, not yet in `IdentityStore`) + switch → the stale standalone token coalesces away; profile + rebound token land for the new identity.
  - `testPreInitProfileAttributesSurviveLaterTokenRegistration` — profile with `firstName` set pre-init, then a `setPushToken` → the profile's attributes survive (not clobbered by token coalescing).
  - `testPreInitProfileWithoutTokenStillBuffersBareProfile` — no token → still a bare `.profile`.

Full suite: KlaviyoSwiftTests 339, KlaviyoCoreTests green.

## Related Issues/Tickets

- [MAGE-1165](https://linear.app/klaviyo/issue/MAGE-1165)
- Targets `feat/modularization-2` (rebased onto it after MAGE-1136 merged).

## Summary by CodeRabbit

* **New Features**
  * Added support for building push-token payloads with caller-provided profile information.
  * Centralized profile payload construction using the current email, phone number, and external ID.

* **Bug Fixes**
  * Improved push-token handling before SDK initialization, including persistence and reliable registration.
  * Ensured profile changes correctly associate buffered push-token registrations with the active identity.
  * Preserved profile attributes when push-token registration follows profile updates.
  * Added support for buffering profiles without a push token and processing queued requests once during initialization.

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -590,7 +590,8 @@ struct KlaviyoReducer: ReducerProtocol {
                 // stored profile, run the SAME identifier-change reset, then push synchronously
                 // so the merged identity is durable before the profile sync is buffered.
                 state.identity = IdentityStore.shared.current
-                let tokenData = IdentityStore.shared.pushToken // token lives in IdentityStore; read before reset
+                // Read before the reset below — the token lives in IdentityStore, not pre-init state.
+                let tokenData = IdentityStore.shared.pushToken
                 let preInitCurrentIds = [state.email, state.phoneNumber, state.externalId]
                 let preInitIncomingIds = [profile.email, profile.phoneNumber, profile.externalId].map {
                     $0?.trimWhiteSpaceOrReturnNilIfEmpty()
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -302,8 +302,8 @@ struct KlaviyoReducer: ReducerProtocol {
                   let anonymousId = state.anonymousId
             else {
                 // Persist the token as canonical device state so a later pre-init identity change
-                // reads it and rebinds; the profile switch's combined token+profile then coalesces
-                // this standalone registration away.
+                // reads it and rebinds; that rebind buffers a fresh `.pushToken` registration, which
+                // coalesces this standalone one away.
                 IdentityStore.shared.updatePushToken(PushTokenData(
                     pushToken: pushToken,
                     pushEnablement: enablement,
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -619,18 +619,14 @@ struct KlaviyoReducer: ReducerProtocol {
                 // attributes (name/title/organization/image/location) survive the pre-init buffer
                 // and sync completely after initialize() (MAGE-1141).
                 let profilePayload = state.profilePayload(from: profile, anonymousId: anonymousId)
+                RequestEnqueuer.enqueueProfile(payload: CreateProfilePayload(data: profilePayload))
                 if let tokenData {
-                    // A push token is registered: rebind it to the new identity in one combined
-                    // token+profile request, mirroring the initialized branch (MAGE-1165).
-                    let payload = RequestFactory.tokenPayload(
-                        pushToken: tokenData.pushToken,
-                        enablement: tokenData.pushEnablement,
-                        background: tokenData.pushBackground,
-                        profile: profilePayload
-                    )
-                    RequestEnqueuer.enqueuePushToken(payload: payload)
-                } else {
-                    RequestEnqueuer.enqueueProfile(payload: CreateProfilePayload(data: profilePayload))
+                    // Re-register the token against the new identity as a SEPARATE call. Keeping the
+                    // profile in its own `.profile` buffer entry means a later token callback (e.g. an
+                    // async automatic APNs fire) can't clobber the profile's structured attributes via
+                    // push-token coalescing. The identity-only registration coalesces to the current
+                    // identity, which the update above just set to the new one.
+                    RequestEnqueuer.enqueuePushToken(tokenData.pushToken, enablement: tokenData.pushEnablement)
                 }
                 return .none
             }
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -590,6 +590,7 @@ struct KlaviyoReducer: ReducerProtocol {
                 // stored profile, run the SAME identifier-change reset, then push synchronously
                 // so the merged identity is durable before the profile sync is buffered.
                 state.identity = IdentityStore.shared.current
+                let tokenData = IdentityStore.shared.pushToken // token lives in IdentityStore; read before reset
                 let preInitCurrentIds = [state.email, state.phoneNumber, state.externalId]
                 let preInitIncomingIds = [profile.email, profile.phoneNumber, profile.externalId].map {
                     $0?.trimWhiteSpaceOrReturnNilIfEmpty()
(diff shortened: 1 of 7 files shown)


Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -607,14 +609,20 @@ struct KlaviyoReducer: ReducerProtocol {
...
                 // and sync completely after initialize() (MAGE-1141).
-                let payload = CreateProfilePayload(data: ProfilePayload(
-                    profile,
-                    email: state.email,
-                    phoneNumber: state.phoneNumber,
-                    externalId: state.externalId,
-                    anonymousId: anonymousId
-                ))
-                RequestEnqueuer.enqueueProfile(payload: payload)
+                let profilePayload = state.profilePayload(from: profile, anonymousId: anonymousId)
```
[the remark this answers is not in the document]
belleklaviyo: Fixed in dbb1a5e0. Pre-init `setPushToken` now persists the token to `IdentityStore` as canonical device state, so a subsequent pre-init identity change reads it and buffers a combined token+profile register instead of a bare profile. The buffer coalesces `.pushToken` entries to the latest, so the standalone registration is superseded — exactly one register drains for the new identity. Regression test `testPreInitTokenSetBeforeProfileSwitchRebindsToNewIdentity` covers set-token → switch-profile → drain (asserts one register, embedded profile carries user B, anon != old).

-Claude

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -607,14 +618,20 @@ struct KlaviyoReducer: ReducerProtocol {
...
+                if let tokenData {
+                    // A push token is registered: rebind it to the new identity in one combined
+                    // token+profile request, mirroring the initialized branch (MAGE-1165).
+                    let payload = RequestFactory.tokenPayload(
+                        pushToken: tokenData.pushToken,
+                        enablement: tokenData.pushEnablement,
+                        background: tokenData.pushBackground,
+                        profile: profilePayload
+                    )
+                    RequestEnqueuer.enqueuePushToken(payload: payload)
```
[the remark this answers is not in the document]
belleklaviyo: Fixed in 83db6ec6. Switched the pre-init rebind to two separate calls: a bare `.profile` (structured attributes, in its own buffer slot immune to push-token coalescing) and an identity-only `.pushToken` registration that coalesces to the current identity. A later token set — including an async automatic APNs callback — no longer clobbers the profile's name/location/custom properties. Also removed the now-unused `enqueuePushToken(payload:)` route that the combined path added. Regression test `testPreInitProfileAttributesSurviveLaterTokenRegistration` asserts a pre-init `firstName` survives a subsequent `setPushToken`.

-Claude

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -811,6 +811,155 @@ class StateManagementTests: StateManagementTestCase {
...
+        store.exhaustivity = .off
+        return store
+    }
+
+    /// Shape B rebind: the profile lives in its own `.profile` buffer entry and the token in a
+    /// separate identity-only `.pushToken` entry, both on the new (post-reset) identity. Asserts that
+    /// pair, then drains at init and asserts exactly one createProfile and one registerPushToken.
+    @MainActor
+    private func assertPreInitRebindProducesSeparateProfileAndToken(
+        _ store: PreInitProfileSwitchStore, email: String, token: String, oldAnon: String
```
glenn-klaviyo, line 849: nit: i personally prefer these params stacked as opposed to single line.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -301,6 +301,15 @@ struct KlaviyoReducer: ReducerProtocol {
                   let apiKey = state.apiKey,
                   let anonymousId = state.anonymousId
             else {
+                // Persist the token as canonical device state so a later pre-init identity change
```
glenn-klaviyo, line 304: This comment describes the earlier design, not the one that shipped here. It says "the profile switch's combined token+profile then coalesces this standalone registration away" — but after 83db6ec6 the pre-init `enqueueProfile` rebind buffers a *separate* identity-only `.pushToken` entry, not a combined token+profile payload.

The coalescing claim itself is still accurate (`UnattributedBuffer.addEntry` drops prior `.pushToken` entries before appending, so this standalone registration is superseded either way) — it's just the "combined token+profile" characterization that's now wrong, and it'll send the next reader looking for a combined payload that no longer exists.

Suggest something like:

```swift
// Persist the token as canonical device state so a later pre-init identity change
// reads it and rebinds; that rebind buffers a fresh `.pushToken` registration, which
// coalesces this standalone one away.
```

glenn-klaviyo (approved):

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
belleklaviyo: ah thanks for catching!

Tests/KlaviyoSwiftTests/StateManagementTests.swift
belleklaviyo: I think I'll do a sweep of swift lint/styling fixes in a separate PR when this whole branch is done alongside doc comment cleanup

glenn-klaviyo: Sounds great to me!

PR #716, opened 2026-09-03
rel/5.4.1

belleklaviyo:
Patch fix for forms bug for expired session observers

## Summary by CodeRabbit

* **Release**
  * Updated the Klaviyo Swift SDK and example apps to version 5.4.1.

* **Bug Fixes**
  * Improved in-app form lifecycle handling across foreground and background transitions.
  * Improved cleanup when sessions expire, while preserving ongoing app and company event monitoring.
  * Added clearer logging when background lifecycle processing encounters errors.

### DIFF
--- a/Examples/KlaviyoSwiftExamples/CocoapodsExample/CocoapodsExample.xcodeproj/project.pbxproj
+++ b/Examples/KlaviyoSwiftExamples/CocoapodsExample/CocoapodsExample.xcodeproj/project.pbxproj
@@ -478,7 +478,7 @@
 					"@executable_path/Frameworks",
 					"@executable_path/../../Frameworks",
 				);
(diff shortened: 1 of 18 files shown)


glenn-klaviyo (approved):

glenn-klaviyo:
just noticed this - we should delete the 15.2 debug / release required branch checks on master since we no longer run on that XC version due flakiness and also due to it becoming deprecated soon. don't think i have the permission to.

Screenshot 2026-09-04 at 10 30 16 AM

Update: looks like i do have access. I removed these from the master branch rule set.
cc: @ajaysubra just so you are aware

new set:
Screenshot 2026-09-04 at 10 40 20 AM

belleklaviyo:
thanks for doing that! @glenn-klaviyo

PR #719, opened 2026-09-04
refactor(KlaviyoCore): consolidate QueueStore to a single klaviyo-queue.json (MAGE-1192)

belleklaviyo:
# Description

Consolidate `QueueStore` from per-company `klaviyo-{apiKey}-queue.json` files to a single `klaviyo-queue.json` (MAGE-1192, sub-issue of MAGE-953).

## Why

Every `KlaviyoRequest` already self-tags its apiKey in its endpoint, so a single queue delivers each request to the correct company on flush — the apiKey in the filename is not load-bearing for delivery. The per-apiKey scoping (MAGE-950, "queue stays apiKey-scoped") was a deliberate but lightly-reasoned choice with no hard technical requirement, and it is the sole source of the company-switch complexity: on a switch the old company's queue lands in a file the new session's flush never touches, stranding undelivered requests (notably the push-unregister). A single durable queue restores pre-modularization single-queue delivery (each request flushed FIFO to its own company) **with** the modern `QueueStore`'s explicit/synchronous persistence, and eliminates the need for a best-effort old-queue flush and any orphaned-queue sweep.

This unblocks **MAGE-953** (company switch), which collapses on top of this to: enqueue unregister as `.high` (triggers immediate flush) → identity reset preserving token → enqueue re-register.

## Due Diligence
- [x] I have tested this on a simulator or a physical device.
- [x] I have added sufficient unit/integration tests of my changes.
- [ ] I have adjusted or added new test cases to team test docs, if applicable.
- [x] I am confident these changes are compatible with all iOS and XCode versions the SDK currently supports.

## Release/Versioning Considerations
- [x] `Patch` Contains internal changes or backwards-compatible bug fixes.
- [ ] `Minor` Contains changes to the public API.
- [ ] `Major` Contains **breaking** changes.
- [x] This is planned work for an upcoming release. (iOS modularization line; targets `feat/modularization-2`.)

## Changelog / Code Overview

- `QueueStore`: registry `[String: QueueStore]` → a single lazily-created shared instance guarded by one lock; `DiskIO.production` → `klaviyo-queue.json` (no apiKey).
[...]

PR #720, opened 2026-09-04
fix(KlaviyoSwift): rebind push token on company switch (MAGE-953)

belleklaviyo:
# Description

On a company (apiKey) switch, promptly move the device push token off the old company and re-register it under the new one — on the single consolidated queue (MAGE-1192). Fixes the cold-start bug where the token was cleared and never re-registered (MAGE-953).

**Rebased onto `feat/modularization-2`** now that MAGE-1192 (#719) has merged, so the diff is just the MAGE-953 changes.

## Why / approach

Identity and apiKey keep flowing through the `KlaviyoState` projection + the write-through `defer`; the only genuinely-new behavior is on the queue. The full "move the switch off `state`" is deferred to MAGE-904 (which deletes the reducer).

Each request self-tags its apiKey, so the single-queue flush delivers unregister → old company and re-register → new company.

## Ordering (Cursor Bugbot fix)

The switch unregister is **appended at standard priority, not front-inserted**. The flush is strictly sequential FIFO, so front-inserting (`.high`) would send the unregister first and any old-company register still queued second — re-binding the token to the old company right after we removed it. Appending places the unregister after all old-company requests, so the final order is `register(old) → unregister(old) → register(new)` and the token ends up only on the new company. Promptness comes from `return .task { .flushQueue }`, not from priority — the two are independent. This restores master's ordering (master also appended) while adding the prompt flush.

## Changelog / Code Overview

- **Runtime switch** (`.initialize` while `.initialized`): old-company unregister is appended (standard priority); `state.reset()` re-registers the token to the new key; the block returns `.task { .flushQueue }` for a prompt flush.
- **Cold-start switch** (`.initialize` while `.uninitialized`, stored apiKey differs): the token is now **preserved** (removed the `updatePushToken(nil)` clear); the unregister is appended + persisted synchronously; a re-register of the preserved token under the new company (new anon, identity-only) is enqueued after the identity reset.
- `RequestFactory.unregisterRequest` gains an additive `priority: RequestPriority = .standard` parameter (backwards-compatible).

## Due Diligence

- [x] This is planned work for an upcoming release. (iOS modularization line.)

## Test Plan

`StateManagementEdgeCaseTests`:
- `testInitializeAfterInitialized` — runtime switch: unregister + register land in the queue in order; `.flushQueue` received; unregister is appended at `.standard` priority.
- `testRuntimeCompanySwitchAppendsUnregisterAfterQueuedOldCompanyRequests` — regression guard: with an old-company register pre-seeded, the final order is `[register(old), unregister(old), register(new)]`.
- `testColdStartCompanySwitchPreservesAndReregistersToken` — cold-start: token preserved (regression), unregister + re-register enqueued.
- `testRuntimeCompanySwitchWithNoTokenEnqueuesNoTokenRequests` / `testColdStartCompanySwitchWithNoTokenEnqueuesNoTokenRequests` — no token → no unregister/register; config/identity still switch.
- `RequestFactoryTests.testUnregisterRequestThreadsPriority`.

Full suite: KlaviyoSwiftTests + KlaviyoCoreTests green.

## Related Issues/Tickets

- [MAGE-953](https://linear.app/klaviyo/issue/MAGE-953)
- Builds on [MAGE-1192](https://linear.app/klaviyo/issue/MAGE-1192) (#719, merged).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved company switching by flushing queued requests immediately when appropriate.
  - Preserved push tokens during cold-start company switches and re-registered them for the new company.
  - Ensured previous-company unregister requests are queued reliably without losing existing token data.
  - Applied standard priority to unregister requests by default while preserving explicitly assigned priorities.
  - Improved request ordering during company transitions for more reliable registration and unregistration processing.

### DIFF
--- a/Sources/KlaviyoCore/Networking/RequestFactory.swift
+++ b/Sources/KlaviyoCore/Networking/RequestFactory.swift
@@ -153,7 +153,8 @@ public enum RequestFactory {
 
     public static func unregisterRequest(
         identity: RequestIdentity,
-        pushToken: String
+        pushToken: String,
+        priority: RequestPriority = .standard
     ) -> KlaviyoRequest {
         let payload = UnregisterPushTokenPayload(
             pushToken: pushToken,
@@ -162,7 +163,7 @@ public enum RequestFactory {
             externalId: identity.externalId,
             anonymousId: identity.anonymousId
         )
-        return KlaviyoRequest(endpoint: .unregisterPushToken(identity.apiKey, payload))
+        return KlaviyoRequest(endpoint: .unregisterPushToken(identity.apiKey, payload), priority: priority)
     }
 
     /// Builds a `PushTokenPayload` with a caller-supplied `ProfilePayload`.
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -206,11 +206,11 @@ struct KlaviyoReducer: ReducerProtocol {
                 }
                 // NOTE: do NOT clear the push token here — the switch must preserve it so the
                 // token can be re-registered under the new company immediately below.
-                if previous.email != nil || previous.phoneNumber != nil || previous.externalId != nil {
-                    // Identified profile: mint a fresh anon and drop PII so `.completeInitialization`
-                    // hydrates a clean identity for the new company.
-                    IdentityStore.shared.update(ProfileData(anonymousId: IdentityStore.shared.mintNewAnonymousId()))
-                }
+                // Give the new company a clean identity: mint a fresh anon and drop any PII so
+                // `.completeInitialization` hydrates it. Unconditional (matches the runtime switch
+                // path's `state.reset()`) so an anonymous-only switch does not carry the old
+                // company's anon into the new one.
+                IdentityStore.shared.update(ProfileData(anonymousId: IdentityStore.shared.mintNewAnonymousId()))
                 // Re-register the preserved token under the new company (identity-only, fresh anon).
                 if let tokenData = IdentityStore.shared.pushToken,
                    let newAnon = IdentityStore.shared.current.anonymousId {
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -166,7 +166,8 @@ struct KlaviyoReducer: ReducerProtocol {
                 guard apiKey != state.apiKey else {
                     return .none
                 }
-                // Since we are moving the token to a new company lets remove the token from the old company first.
+                // Moving the token to a new company: unregister from the old one. Appended (not
+                // front-inserted) so it sends after any already-queued old-company requests.
                 if let apiKey = state.apiKey,
                    let anonymousId = state.anonymousId,
                    let tokenData = state.pushTokenData {
@@ -178,6 +179,7 @@ struct KlaviyoReducer: ReducerProtocol {
                 }
                 state.apiKey = apiKey
                 state.reset()
+                return .task { .flushQueue }
             } else if case .uninitialized = state.initalizationState,
                       let previousApiKey = SDKConfigStore.shared.current.apiKey,
                       previousApiKey != apiKey {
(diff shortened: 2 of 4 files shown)


Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -198,15 +200,30 @@ struct KlaviyoReducer: ReducerProtocol {
...
-                    // Identified profile: mint a fresh anon and drop PII so `.completeInitialization`
-                    // hydrates a clean identity for the new company.
-                    IdentityStore.shared.update(ProfileData(anonymousId: IdentityStore.shared.mintNewAnonymousId()))
+                // NOTE: do NOT clear the push token here — the switch must preserve it so the
+                // token can be re-registered under the new company immediately below.
+                // Give the new company a clean identity: mint a fresh anon and drop any PII so
+                // `.completeInitialization` hydrates it. Unconditional (matches the runtime switch
+                // path's `state.reset()`) so an anonymous-only switch does not carry the old
+                // company's anon into the new one.
+                IdentityStore.shared.update(ProfileData(anonymousId: IdentityStore.shared.mintNewAnonymousId()))
```
belleklaviyo, line 213: TODO: check that this path isn't going to be increasing the number of anon id's created unexpectedly (beyond existing paths)

belleklaviyo: this is moreso a reminder for later after the last ticket goes in

belleklaviyo: did a preliminary check and tldr:

_The architecture centralizes minting rather than proliferating it; PR 720 mints in equal-or-fewer scenarios than master. The only new consideration is the lazy auto-mint's timing relative to legacy migration on the cold-start switch —_

_The lazy auto-mint in hydrateIfNeeded() is new: any first read of IdentityStore.shared.current with no persisted anon will mint one. That widens the trigger surface — reads now exist in RequestEnqueuer:21 and several StateManagement sites — but it's idempotent: guarded by the hydrated flag + persistence, so it mints **at most once per install**. It does not increase distinct-anon churn; it just moves the "when" from explicit lifecycle events to first-touch._

Sources/KlaviyoCore/Networking/RequestFactory.swift
```diff
@@ -153,7 +153,8 @@ public enum RequestFactory {
 
     public static func unregisterRequest(
         identity: RequestIdentity,
-        pushToken: String
+        pushToken: String,
+        priority: RequestPriority = .standard
```
ndurell, line 157: should it be high priority?

ab1470 (approved):

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -198,15 +200,30 @@ struct KlaviyoReducer: ReducerProtocol {
...
+                // Re-register the preserved token under the new company (identity-only, fresh anon).
+                if let tokenData = IdentityStore.shared.pushToken,
+                   let newAnon = IdentityStore.shared.current.anonymousId {
+                    let request = RequestFactory.tokenRequest(
+                        apiKey: apiKey,
+                        pushToken: tokenData.pushToken,
+                        enablement: tokenData.pushEnablement,
+                        background: tokenData.pushBackground.rawValue,
+                        profile: ProfilePayload(
+                            email: nil, phoneNumber: nil, externalId: nil, anonymousId: newAnon
```
ndurell, line 223: we don't want to preserve any identifiers when we switch companies right?

Sources/KlaviyoCore/Networking/RequestFactory.swift
belleklaviyo: no it needs to just be standard. if it's high priority it would always be inserted front of queue and opens a path for it to be in front of a register request and might end up rebinding the token

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
belleklaviyo: yup no identifiers, just the push token

PR #722, opened 2026-09-09
refactor(KlaviyoSwift): route the post-init public API through the ungated Core path (MAGE-1196)

belleklaviyo:
# Description

Route the post-init public API of the KlaviyoSwift TCA reducer through the ungated KlaviyoCore path — writing identity to the canonical `IdentityStore` and enqueuing via `RequestEnqueuer` (→ `QueueStore`) — instead of mutating the `KlaviyoState` projection and relying on the write-through `defer`. This strips request-building/identity ownership out of `StateManagement.swift`, a prerequisite for retiring the reducer + vendored TCA in **MAGE-904**.

**Rebased onto `feat/modularization-2`** now that MAGE-953 (#720) has merged — the diff is just the MAGE-1196 changes. Sub-issue of MAGE-904.

## Why / approach

The pre-init branches already delegate to `RequestEnqueuer` + `IdentityStore`; the post-init paths still used the legacy `state`-projection + `defer` write-through. This collapses each action's two branches into one init-state-agnostic path, folding in the post-init-only nuances (prompt-flush, `enrichAndPublishEvent`, dedup, skip optimization, token re-association, identifier-change reset). Parity-preserving, gated by the existing reducer test suite.

## Changelog / Code Overview

- **Identity setters** (`setEmail`/`setPhoneNumber`/`setExternalId`) → new `applyIdentifierChange` helper (writes `IdentityStore`; `isNotEmptyOrSame` empty/unchanged short-circuit + developer warning preserved; token re-association gated on `state.apiKey` so pre-init/warm-start falls through to the buffered/queued path instead of dropping).
- **Push-token setters** (`setPushToken`/`setAutomaticPushToken`/`setPushEnablement`) → dedup against `IdentityStore.pushToken`; `setPushToken` **write-throughs `state.pushTokenData`** (persisted to `IdentityStore` by the reducer `defer`) so the projection stays in sync for all state readers; `setPushEnablement` reads the **canonical `IdentityStore` token**, not the stale projection. Token path gated on `state.apiKey` (matches `enqueueRequest`) so a warm-start persisted apiKey can't drop the register.
- **Events** (`enqueueEvent`/`enqueueAggregateEvent`) → `RequestEnqueuer`; `enrichAndPublishEvent` (`.fireAndForget`) + the high-priority opened-push/geofence prompt-flush kept **post-init only**.
- **`enqueueProfile`** → Core path. **Intentional behavior change:** with an existing token it now sends **two** requests — `createProfile` (attributes) + an identity-only `registerPushToken` — instead of one token request with embedded attributes. Net server state is equivalent (attributes persisted + token bound to the same identity); the identity-only registration, ordered after `createProfile`, carries no attributes so it can't overwrite them. Skip optimization + identifier-change reset (fresh anon + PII drop) preserved.
- **`enqueueSubscription`** + **`trackingLinkResolutionFailed`** → `RequestEnqueuer` (subscription channel/identifier validation preserved — invalid combos still enqueue nothing).
- **`resetProfile`** → Core path (fresh anon + PII drop; canonical push token retained and re-registered under the new identity).
- **Flush safety:** `enqueueProfileOrTokenRequest` (the `pendingProfile` drain in `flushQueue`) no longer nils `state.pushTokenData`, so the write-through `defer` can't persist nil into `IdentityStore` and wipe the canonical/persisted token mid-flush (a crash in that window previously lost the token on disk).
- Removed now-dead `KlaviyoState.updateEmail/updatePhoneNumber/updateExternalId`.
- **Out of scope (MAGE-1197/904):** `setProfileProperty`/`pendingProfile` staging, the flush/retry engine, `initialize`/`completeInitialization`, dispatch-site removal, and `KlaviyoState`/vendored-TCA deletion.

## Review feedback addressed (commits `fd5e09bc`, `4eaec236`)

- **Stale-token clobber:** `setPushEnablement` read the stale `state.pushTokenData` and could re-register/clobber the canonical token → now reads `IdentityStore`.
- **Warm-start drop (Critical):** token path gated on `SDKConfigStore.apiKey` but enqueued via the `state.apiKey`-gated `enqueueRequest` → dropped when a persisted apiKey preceded `.initialize`. Both sites now gate on `state.apiKey`.
- **Flush token-wipe:** see "Flush safety" above.
- **CodeRabbit quick wins:** `setPushToken` write-through (CR#1), `@MainActor` fix, DRY test helpers (`makeInitializedIdentifierStore`, `expectedIdentityOnlyTokenRequest`, `seedCanonicalStores`).

## Due Diligence

- [x] This is planned work for an upcoming release. (iOS modularization line.)

## Test Plan

Full `KlaviyoSwiftTests` + `KlaviyoCoreTests` green (0 failures). Notable additions:
- **Payload-parity snapshots** proving the Core-built request equals the legacy builder output: `createEvent` (incl. `_openedPush` push-token branch) and `createProfile`.
- **Regression gates:** `setPushEnablement` reads the canonical token; warm-start `setEmail` (persisted apiKey, nil `state.apiKey`) enqueues rather than drops; `flushQueue` with a staged `pendingProfile` keeps the canonical push token; empty/whitespace/unchanged setters enqueue nothing; token retained after `resetProfile`.
- Existing setter/event/profile/subscription/reset tests re-pointed to the canonical-store contract (assert `IdentityStore` + spy `QueueStore`) across `StateManagementTests`, `StateManagementEdgeCaseTests`, `StateManagementEnqueueEdgeCaseTests`, `ResolveTrackingLinkTests`.

## Related Issues/Tickets

- [MAGE-1196](https://linear.app/klaviyo/issue/MAGE-1196) (sub-issue of [MAGE-904](https://linear.app/klaviyo/issue/MAGE-904))
- Builds on [MAGE-953](https://linear.app/klaviyo/issue/MAGE-953) (#720, merged).

## Summary by CodeRabbit

* **Improvements**
  * Improved handling of identity, profile, event, subscription, push-token, reset, and tracking-link requests.
  * Added reliable buffering for operations performed before SDK initialization.
  * Improved persistence, synchronization, and deduplication of identifiers and push tokens.
  * Avoided unnecessary profile updates when identifiers remain unchanged.
  * Preserved push-token registration across profile resets.
* **Tests**
  * Expanded coverage for initialization, identity persistence, request routing, token handling, and profile updates.
* **Documentation**
  * Clarified test documentation and removed obsolete issue references.

### DIFF
--- a/Sources/KlaviyoSwift/StateManagement/KlaviyoState+RequestBuilding.swift
+++ b/Sources/KlaviyoSwift/StateManagement/KlaviyoState+RequestBuilding.swift
@@ -65,7 +65,12 @@ extension KlaviyoState {
         // if we have push data and we are switching emails
         // we want to associate the token with the new email.
         if let pushTokenData = pushTokenData {
-            self.pushTokenData = nil
+            // Do NOT nil `self.pushTokenData` here: the request below is built from the captured
+            // value, and the reducer's write-through `defer` would otherwise persist nil into
+            // IdentityStore — wiping the canonical/persisted token until the register completes (a
+            // crash in that window loses the token on disk, and concurrent reads see nil). This is
+            // a token re-registration, not an unregister, so the token stays; deQueueCompletedResults
+            // reconfirms it from the response. (MAGE-1196)
             let request = resolvedTokenRequest(
                 apiKey: apiKey,
                 anonymousId: anonymousId,
--- a/Sources/KlaviyoSwift/StateManagement/KlaviyoState+RequestBuilding.swift
+++ b/Sources/KlaviyoSwift/StateManagement/KlaviyoState+RequestBuilding.swift
@@ -65,12 +65,6 @@ extension KlaviyoState {
         // if we have push data and we are switching emails
         // we want to associate the token with the new email.
         if let pushTokenData = pushTokenData {
-            // Do NOT nil `self.pushTokenData` here: the request below is built from the captured
-            // value, and the reducer's write-through `defer` would otherwise persist nil into
-            // IdentityStore — wiping the canonical/persisted token until the register completes (a
-            // crash in that window loses the token on disk, and concurrent reads see nil). This is
-            // a token re-registration, not an unregister, so the token stays; deQueueCompletedResults
-            // reconfirms it from the response. (MAGE-1196)
             let request = resolvedTokenRequest(
                 apiKey: apiKey,
                 anonymousId: anonymousId,
--- a/Sources/KlaviyoSwift/StateManagement/KlaviyoState+RequestBuilding.swift
+++ b/Sources/KlaviyoSwift/StateManagement/KlaviyoState+RequestBuilding.swift
@@ -57,14 +57,14 @@ extension KlaviyoState {
     }
 
     mutating func enqueueProfileOrTokenRequest() {
-        guard let apiKey = apiKey,
-              let anonymousId = anonymousId else {
+        guard let apiKey,
+              let anonymousId else {
             environment.emitDeveloperWarning("SDK internal error")
             return
         }
         // if we have push data and we are switching emails
         // we want to associate the token with the new email.
-        if let pushTokenData = pushTokenData {
+        if let pushTokenData {
             let request = resolvedTokenRequest(
                 apiKey: apiKey,
                 anonymousId: anonymousId,
@@ -86,7 +86,7 @@ extension KlaviyoState {
     }
 
     mutating func updateRequestAndStateWithPendingProfile(profile: CreateProfilePayload) -> CreateProfilePayload {
-        guard let pendingProfile = pendingProfile else {
+        guard let pendingProfile else {
             return profile
         }
         let updatedProfile = Profile.updateProfileWithProperties(dict: pendingProfile)
--- a/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
+++ b/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
@@ -142,8 +142,8 @@ struct KlaviyoState: Equatable {
         pushTokenData = nil
         if preserveTokenData {
             pushTokenData = previousPushTokenData
-            if let apiKey = apiKey,
-               let anonymousId = anonymousId,
+            if let apiKey,
+               let anonymousId,
                let tokenData = previousPushTokenData {
                 let profile = ProfilePayload(Profile(), anonymousId: anonymousId)
                 let request = RequestFactory.tokenRequest(
--- a/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
+++ b/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
@@ -157,21 +157,6 @@ struct KlaviyoState: Equatable {
             }
         }
     }
-
-    func shouldSendTokenUpdate(newToken: String, enablement: PushEnablement) -> Bool {
-        guard let pushTokenData = pushTokenData else {
-            return true
-        }
-        let currentDeviceMetadata = DeviceMetadata(context: environment.appContextInfo())
-        let newPushTokenData = PushTokenData(
-            pushToken: newToken,
-            pushEnablement: enablement,
-            pushBackground: environment.getBackgroundSetting(),
-            deviceData: currentDeviceMetadata
-        )
-
-        return pushTokenData != newPushTokenData
-    }
 }
 
 // MARK: Klaviyo state persistence
--- a/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
+++ b/Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
@@ -97,27 +97,6 @@ struct KlaviyoState: Equatable {
         QueueStore.shared.enqueue(request)
     }
 
-    mutating func updateEmail(email: String) {
-        if email.isNotEmptyOrSame(as: self.email, identifier: "email") {
-            self.email = email.trimWhiteSpaceOrReturnNilIfEmpty()
-            enqueueProfileOrTokenRequest()
-        }
-    }
-
-    mutating func updateExternalId(externalId: String) {
-        if externalId.isNotEmptyOrSame(as: self.externalId, identifier: "external Id") {
-            self.externalId = externalId.trimWhiteSpaceOrReturnNilIfEmpty()
-            enqueueProfileOrTokenRequest()
-        }
-    }
-
-    mutating func updatePhoneNumber(phoneNumber: String) {
-        if phoneNumber.isNotEmptyOrSame(as: self.phoneNumber, identifier: "phone number") {
-            self.phoneNumber = phoneNumber.trimWhiteSpaceOrReturnNilIfEmpty()
-            enqueueProfileOrTokenRequest()
-        }
-    }
-
     func requestIdentity(apiKey: String, anonymousId: String) -> RequestIdentity {
         RequestIdentity(
             apiKey: apiKey,
--- a/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
+++ b/Sources/KlaviyoSwift/StateManagement/StateManagement.swift
@@ -279,29 +279,28 @@ struct KlaviyoReducer: ReducerProtocol {
             .merge(with: environment.lifecycleEventsWithReachability().map(\.transformToKlaviyoAction).eraseToEffect())
 
         case let .setEmail(email):
-            guard case .initialized = state.initalizationState else {
-                setPreInitIdentifier(&state) { $0.email = email.trimWhiteSpaceOrReturnNilIfEmpty() }
+            guard email.isNotEmptyOrSame(as: IdentityStore.shared.current.email, identifier: "email") else {
                 return .none
             }
-            state.updateEmail(email: email)
+            applyIdentifierChange(&state) { $0.email = email.trimWhiteSpaceOrReturnNilIfEmpty() }
             return .none
 
         case let .setPhoneNumber(phoneNumber):
-            guard case .initialized = state.initalizationState else {
-                setPreInitIdentifier(&state) {
-                    $0.phoneNumber = phoneNumber.trimWhiteSpaceOrReturnNilIfEmpty()
-                }
+            guard phoneNumber.isNotEmptyOrSame(
+                as: IdentityStore.shared.current.phoneNumber, identifier: "phone number"
+            ) else {
                 return .none
             }
-            state.updatePhoneNumber(phoneNumber: phoneNumber)
+            applyIdentifierChange(&state) { $0.phoneNumber = phoneNumber.trimWhiteSpaceOrReturnNilIfEmpty() }
             return .none
 
         case let .setExternalId(externalId):
-            guard case .initialized = state.initalizationState else {
-                setPreInitIdentifier(&state) { $0.externalId = externalId.trimWhiteSpaceOrReturnNilIfEmpty() }
+            guard externalId.isNotEmptyOrSame(
+                as: IdentityStore.shared.current.externalId, identifier: "external id"
+            ) else {
                 return .none
             }
-            state.updateExternalId(externalId: externalId)
+            applyIdentifierChange(&state) { $0.externalId = externalId.trimWhiteSpaceOrReturnNilIfEmpty() }
             return .none
 
         case let .setAutomaticPushToken(pushToken, enablement):
@@ -314,36 +313,32 @@ struct KlaviyoReducer: ReducerProtocol {
             }
 
         case let .setPushToken(pushToken, enablement):
-            guard case .initialized = state.initalizationState,
-                  let apiKey = state.apiKey,
-                  let anonymousId = state.anonymousId
-            else {
-                // Persist the token as canonical device state so a later pre-init identity change
-                // reads it and rebinds; that rebind buffers a fresh `.pushToken` registration, which
-                // coalesces this standalone one away.
-                IdentityStore.shared.updatePushToken(PushTokenData(
-                    pushToken: pushToken,
-                    pushEnablement: enablement,
-                    pushBackground: environment.getBackgroundSetting(),
-                    deviceData: DeviceMetadata(context: environment.appContextInfo())
-                ))
-                RequestEnqueuer.enqueuePushToken(pushToken, enablement: enablement)
-                return .none
-            }
-            if !state.shouldSendTokenUpdate(newToken: pushToken, enablement: enablement) {
-                return .none
-            }
-
-            let request = state.resolvedTokenRequest(
-                apiKey: apiKey,
-                anonymousId: anonymousId,
-                pushToken: pushToken,
-                enablement: enablement
+            let newTokenData = PushTokenData(
+                pushToken: pushToken, pushEnablement: enablement,
+                pushBackground: environment.getBackgroundSetting(),
+                deviceData: DeviceMetadata(context: environment.appContextInfo())
             )
-            state.enqueueRequest(request: request)
+            // Dedup against the canonical token (parity with shouldSendTokenUpdate).
+            guard IdentityStore.shared.pushToken != newTokenData else { return .none }
+            IdentityStore.shared.updatePushToken(newTokenData)
+            guard let anonymousId = IdentityStore.shared.current.anonymousId else {
+                environment.emitDeveloperWarning("SDK internal error: missing anonymousId")
+                return .none
+            }
+            if let apiKey = SDKConfigStore.shared.current.apiKey {
+                // Post-init: fold + consume any pending profile into the registration.
+                state.identity = IdentityStore.shared.current
+                let request = state.resolvedTokenRequest(
+                    apiKey: apiKey, anonymousId: anonymousId, pushToken: pushToken, enablement: enablement
+                )
+                state.enqueueRequest(request: request)
+            } else {
+                RequestEnqueuer.enqueuePushToken(pushToken, enablement: enablement) // pre-init buffer
+            }
             return .none
 
         case let .setPushEnablement(enablement):
+            // TODO(MAGE-1197): read the token from IdentityStore, not state.pushTokenData
             guard let pushToken = state.pushTokenData?.pushToken else {
                 return .none
             }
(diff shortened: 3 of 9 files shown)


Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -97,27 +97,6 @@ struct KlaviyoState: Equatable {
         QueueStore.shared.enqueue(request)
     }
 
-    mutating func updateEmail(email: String) {
```
belleklaviyo, line 100: all removed funcs in this file are no longer used

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -314,37 +311,39 @@ struct KlaviyoReducer: ReducerProtocol {
...
-                ))
-                RequestEnqueuer.enqueuePushToken(pushToken, enablement: enablement)
+            let newTokenData = PushTokenData(
+                pushToken: pushToken, pushEnablement: enablement,
+                pushBackground: environment.getBackgroundSetting(),
+                deviceData: DeviceMetadata(context: environment.appContextInfo())
+            )
+            // Dedup against the canonical token: skip when token + enablement + background +
+            // device metadata all match.
+            guard IdentityStore.shared.pushToken != newTokenData else { return .none }
```
belleklaviyo, line 321: effectively the check that was `shouldSendTokenUpdate`

Sources/KlaviyoSwift/StateManagement/KlaviyoState+RequestBuilding.swift
```diff
@@ -57,15 +57,14 @@ extension KlaviyoState {
...
-              let anonymousId = anonymousId else {
+        guard let apiKey,
+              let anonymousId else {
             environment.emitDeveloperWarning("SDK internal error")
             return
         }
         // if we have push data and we are switching emails
         // we want to associate the token with the new email.
-        if let pushTokenData = pushTokenData {
-            self.pushTokenData = nil
```
belleklaviyo, line 68: this was an intermediary but we shouldn't be nil'ing the push token data, we need to preserve it to transfer it to the next request

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -556,203 +555,81 @@ struct KlaviyoReducer: ReducerProtocol {
...
+            // Identifier change on an already-identified profile → mint a fresh anonymousId and
+            // drop prior PII, so a set(profile:) with different identifiers does not reuse the
+            // previous user's anonymousId (which would merge two people onto one profile).
             // resetProfile() remains available for explicitly clobbering all state.
             if state.isIdentified, identifiersChanged {
                 state.reset(preserveTokenData: false)
             }
             state.updateStateWithProfile(profile: profile)
-
+            IdentityStore.shared.update(state.identity)
```
belleklaviyo, line 604: this duplicative double write to both state and IdentityStore (or SDKConfigStore in other places) is mostly transitional, but will be going away once we remove TCA

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -653,66 +828,167 @@ class StateManagementTests: StateManagementTestCase {
...
 
     // MARK: - Test enqueue event
 
+    /// Payload-parity: the Core path (RequestEnqueuer → RequestFactory.eventPayload) must produce
+    /// a structurally identical payload to the legacy reducer path
+    /// (updateEventWithIdentifiers + eventRequest) for an identified profile with a push token.
+    /// Must pass before the cutover proceeds.
+    @MainActor
+    func testEnqueueEventPayloadParityWithLegacyBuilder() throws {
+        let state = INITIALIZED_TEST_STATE()
```
belleklaviyo, line 904: this test is transitional but will go away with the big kahuna removal

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -653,66 +828,167 @@ class StateManagementTests: StateManagementTestCase {
...
+            legacyPayload,
+            corePayload,
+            "Core event payload must match the legacy builder"
+        )
+    }
+
+    /// Verifies payload parity for `._openedPush` — the event type whose `updateEventWithIdentifiers`
+    /// push_token branch matters most (it injects the token into event properties).
+    @MainActor
+    func testOpenedPushEventPayloadParityWithLegacyBuilder() throws {
```
belleklaviyo, line 939: same with this one

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -320,7 +331,42 @@ class StateManagementTests: StateManagementTestCase {
...
-        XCTAssertEqual(readQueue(), [pushTokenRequest])
+        XCTAssertEqual(readQueue().map(\.endpoint), [pushTokenRequest].map(\.endpoint))
+    }
+
+    /// Regression: after a token rotation, `setPushToken` writes the new token to
+    /// `IdentityStore` but leaves `state.pushTokenData` stale until the register drains. If
+    /// `setPushEnablement` read the stale `state` token it would forward it and clobber the canonical
+    /// store back to the old token. It must read the canonical `IdentityStore` token instead.
+    @MainActor
+    func testSetPushEnablementReadsCanonicalTokenNotStaleState() async throws {
```
belleklaviyo, line 342: transitional test too

Tests/KlaviyoSwiftTests/StateManagementTests.swift
```diff
@@ -622,26 +668,155 @@ class StateManagementTests: StateManagementTestCase {
         }
     }
 
+    /// Regression: `flushQueue` with a staged `pendingProfile` calls
+    /// `enqueueProfileOrTokenRequest`, which must not leave `state.pushTokenData` nil — otherwise the
+    /// write-through `defer` persists nil into `IdentityStore`, wiping the canonical/persisted token
+    /// until the in-flight register completes (a crash in that window loses the token on disk).
+    @MainActor
+    func testFlushWithPendingProfileKeepsCanonicalToken() async throws {
```
belleklaviyo, line 676: transitional

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -808,61 +685,64 @@ struct KlaviyoReducer: ReducerProtocol {
...
+            let request = state.resolvedTokenRequest(
+                apiKey: apiKey,
+                anonymousId: anonymousId,
+                pushToken: tokenData.pushToken,
+                enablement: tokenData.pushEnablement
+            )
+            state.enqueueRequest(request: request) // already targets QueueStore.shared
+        } else {
+            // Pre-init (no apiKey → buffer) or post-init with no token: enqueue a profile via the
+            // ungated RequestEnqueuer, folding any pending profile. Matches the legacy
```
Atharva8168: The `Matches the legacy setPreInitIdentifier.` claim here is not accurate — worth correcting so the next reader does not assume parity.

Legacy `setPreInitIdentifier` built a bare payload and never touched `pendingProfile`:

```swift
let payload = CreateProfilePayload(data: ProfilePayload(
    email: state.email, phoneNumber: state.phoneNumber,
    externalId: state.externalId, anonymousId: anonymousId
))
RequestEnqueuer.enqueueProfile(payload: payload)
```

The new branch routes through `state.updateRequestAndStateWithPendingProfile(profile:)`, which both folds the staged properties in **and consumes them** (`self.pendingProfile = nil`). So a pre-init `setProfileProperty(...)` followed by `setEmail(...)` now ships those properties in the buffered profile request instead of leaving them staged for a later flush.

That looks like a deliberate improvement and I think it is the right behavior — the objection is only to the comment asserting parity where there is a real behavior change. Suggest rewording to call out the fold+consume explicitly (and, if the change is intentional, a test covering pre-init `setProfileProperty` → `setEmail` would pin it down).

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -314,37 +311,39 @@ struct KlaviyoReducer: ReducerProtocol {
...
+            // then would be dropped. Pre-init falls through to the durable buffer instead.
+            if let apiKey = state.apiKey {
+                // Post-init: fold + consume any pending profile into the registration.
+                state.identity = IdentityStore.shared.current
+                let request = state.resolvedTokenRequest(
+                    apiKey: apiKey, anonymousId: anonymousId, pushToken: pushToken, enablement: enablement
+                )
+                state.enqueueRequest(request: request)
+            } else {
+                RequestEnqueuer.enqueuePushToken(pushToken, enablement: enablement) // pre-init buffer
```
Atharva8168: Nit on the comment, not the code — the code is right, but `// pre-init buffer` and the block comment above ("Pre-init falls through to the durable buffer instead") describe a path this branch does not always take.

The branch gates on `state.apiKey`, but `RequestEnqueuer.route` gates on `SDKConfigStore.shared.current.apiKey`:

```swift
if let apiKey = SDKConfigStore.shared.current.apiKey {
    QueueStore.shared.enqueue(build(apiKey))
} else {
    UnattributedBuffer.shared.append(buffered)
}
```

On the exact warm start this fix targets — persisted apiKey in `SDKConfigStore`, `state.apiKey` still nil because `.initialize` has not run through the reducer — we take the else branch and `route` sends the registration **straight to `QueueStore`**, never to `UnattributedBuffer`. Which is the desired outcome; that divergence in gates is what stops the token from being dropped.

Suggest something like `// pre-init or warm start: RequestEnqueuer re-gates on SDKConfigStore` so a reader does not conclude the token is always buffered here. Same nit as the one I left on `applyIdentifierChange`.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -556,203 +555,81 @@ struct KlaviyoReducer: ReducerProtocol {
...
-            // Anonymous ID is the lowest-order identifier, so there's no reason
-            // to regenerate it when higher-order identifiers haven't changed.
-            // Resetting with the same identifiers causes unnecessary anonymous ID
-            // churn, which triggers spurious push-token API requests.
+            // Identifier change on an already-identified profile → mint a fresh anonymousId and
+            // drop prior PII, so a set(profile:) with different identifiers does not reuse the
+            // previous user's anonymousId (which would merge two people onto one profile).
             // resetProfile() remains available for explicitly clobbering all state.
             if state.isIdentified, identifiersChanged {
                 state.reset(preserveTokenData: false)
```
Atharva8168, line 600: **Blocking — this wipes the canonical push token.**

`reset(preserveTokenData: false)` clears the projection:

```swift
pushTokenData = nil
if preserveTokenData { pushTokenData = previousPushTokenData; ... }   // not taken
```

`enqueueProfile` never restores it. On return the write-through `defer` at the top of `reduce` fires:

```swift
if state.pushTokenData != previousPushTokenData {
    IdentityStore.shared.updatePushToken(state.pushTokenData)   // nil
}
```

so `IdentityStore.shared.pushToken` is erased, and persisted to disk as nil.

**Repro:** an identified user with a registered push token calls `set(profile:)` with a different email. `identifiersChanged` is true, `isIdentified` is true, reset fires.

The `registerPushToken` request itself still goes out correctly — `tokenData` is captured from `IdentityStore` before the reset — so the server-side binding is fine. What is lost is local canonical state. Downstream:

- `setPushEnablement` guards on `IdentityStore.shared.pushToken?.pushToken` and silently returns `.none` from then on; enablement changes stop being reported.
- `RequestEnqueuer.enqueueEvent` reads `IdentityStore.shared.pushToken?.pushToken` for event payloads, so events lose push-token attribution — including `_openedPush`.
- A later `resetProfile` reads `tokenData = nil` and skips the re-registration under the new anon.

It does not self-heal on relaunch, since the nil was persisted. Recovery needs a fresh `setPushToken` from an APNs callback.

**This is the same defect class the PR fixes in two other places** — `enqueueProfileOrTokenRequest` (the "flush safety" change) and `resetProfile`, which does exactly the right thing:

```swift
state.reset(preserveTokenData: false)
state.pushTokenData = tokenData        // <-- restores the projection so the defer is a no-op
IdentityStore.shared.update(state.identity)
```

`enqueueProfile` looks like the missed third site. Suggested fix, mirroring `resetProfile`:

```swift
if state.isIdentified, identifiersChanged {
    state.reset(preserveTokenData: false)
    state.pushTokenData = tokenData
}
```

**Test gap:** none of the updated `enqueueProfile` tests would catch this. They assert `readQueue()` contents only, and since `store.exhaustivity = .off` the removed `$0.pushTokenData = nil` expectations are not checked in either direction. Worth adding a regression test that asserts `IdentityStore.shared.pushToken` is still non-nil after an identifier-changing `set(profile:)` — it would sit naturally alongside the `flushQueue` token-retention test already added for the flush-safety fix.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -556,203 +555,81 @@ struct KlaviyoReducer: ReducerProtocol {
...
-            ])
+        case let .enqueueEvent(event):
+            RequestEnqueuer.enqueueEvent(event)
+            // Post-init only, matching today: publish to the EventBus (drives event-triggered in-app
+            // forms via KlaviyoForms' ProfileEventObserver) and prompt-flush high-priority events.
+            // Pre-init there is no forms observer and the flush engine no-ops, so gate on init state.
+            guard case .initialized = state.initalizationState else { return .none }
+            // `.fireAndForget` keeps publish async and reentrancy-safe, matching the pre-cutover
+            // semantics: an EventBus subscriber cannot dispatch back into the store synchronously.
+            let publish = EffectTask<KlaviyoAction>.fireAndForget { enrichAndPublishEvent(event) }
```
Atharva8168: Minor / low severity, but flagging it since the PR is billed as parity-preserving.

The case moved from `case var .enqueueEvent(event)` to `case let`, and with it the reassignment that used to happen before the publish:

```swift
event = event.updateEventWithIdentifiers(
    email: state.email, phoneNumber: state.phoneNumber,
    externalId: state.externalId, pushToken: state.pushTokenData?.pushToken
)
...
.fireAndForget { enrichAndPublishEvent(event) }   // <- stamped event
```

The **outbound request keeps parity** — `RequestEnqueuer.enqueueEvent` goes through `RequestFactory.eventPayload`, which calls `updateEventWithIdentifiers` itself. No concern there.

The difference is on the EventBus publish path, which now receives the raw event. `updateEventWithIdentifiers` does more than populate `.identifiers`:

```swift
if metric.name == EventName._openedPush, let pushToken {
    properties["push_token"] = pushToken
}
```

Tracing the subscriber: `ProfileEventObserver` -> `IAFPresentationManager.handleProfileEventCreated` reads `metric` / `uniqueId` / `time` / `value` / `properties` and serializes them into `dispatchProfileEvent(...)` for KlaviyoJS. It never touches `.identifiers`, so that half genuinely does not matter — but it does serialize `properties`. So `_openedPush` events now reach KlaviyoJS without the `push_token` property they previously carried.

I have no evidence `push_token` is used as a form-targeting criterion, so I would not block on this. But it is an observable change on the forms path, and nothing in the suite covers it. If it is intentional, a one-line note next to the `.fireAndForget` would save the next reader the same trace.

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
glenn-klaviyo: nice

Tests/KlaviyoSwiftTests/StateManagementEnqueueEdgeCaseTests.swift
```diff
@@ -116,26 +116,40 @@ class StateManagementEnqueueEdgeCaseTests: StateManagementTestCase {
...
+            anonymousId: newAnon
         )
-        XCTAssertEqual(readQueue(), [request])
+        let profileRequest = KlaviyoRequest(
+            endpoint: .createProfile(TEST_API_KEY, CreateProfilePayload(data: profilePayload))
+        )
+        let tokenRequest = expectedIdentityOnlyTokenRequest(
+            anonymousId: newAnon, tokenData: initialState.pushTokenData!
+        )
+        XCTAssertEqual(readQueue().map(\.endpoint), [profileRequest, tokenRequest].map(\.endpoint))
```
Atharva8168, line 152: Six assertions in this file moved from comparing whole requests to comparing endpoints only:

```swift
XCTAssertEqual(readQueue(), [request])
// ->
XCTAssertEqual(readQueue().map(\.endpoint), [profileRequest, tokenRequest].map(\.endpoint))
```

I assume the motive was `id` / `enqueuedAt` noise once two requests are enqueued rather than one, which is reasonable. But `KlaviyoRequest` is `Equatable` over `id`, `endpoint`, `enqueuedAt`, **and `priority`** — so projecting to `endpoint` alone drops priority coverage as collateral.

That is a slightly awkward loss in this particular PR, because it also changes how priority is decided:

```swift
// before
let shouldPrioritize = request.priority == .high
// after
return event.priority == .high ? .merge([.task { .flushQueue }, publish]) : publish
```

The prompt-flush decision moved from the built request to the incoming event, and after this change nothing in these tests pins priority on an enqueued request. Comparing both fields would keep it:

```swift
XCTAssertEqual(
    readQueue().map { ($0.endpoint, $0.priority) },
    [(profileRequest.endpoint, profileRequest.priority), (tokenRequest.endpoint, tokenRequest.priority)],
    // or add a small `assertQueue(_:matches:)` helper if the tuple comparison is noisy
)
```

Separately, line 259 (`testSetProfileSameIdentifiersDifferentAttributesStillUpdates`) is a single-request case that was not affected by the one-to-two split, so `.map(\.endpoint)` on both sides is not buying anything there and could stay a plain request comparison.

Atharva8168: Correction to my own comment above — I overstated the impact.

I had not yet read `StateManagementTests.swift` (its diff came back truncated from the API, so I pulled the branch). Priority **is** pinned there, directly and well:

- `testOpenedPushEventProducesHighPriorityRequestAtQueueFront` asserts `front.priority == .high`
- `testGeofenceEventProducesHighPriorityRequestAtQueueFront`
- `testStandardEventProducesStandardPriorityRequestAppendedToQueue`

So the `event.priority` / `request.priority` switch is covered, and my "nothing in these tests pins priority anywhere" line was wrong — it is not pinned *in this file*, which is a much weaker point.

What survives is just the smaller suggestion: line 259 is a single-request case where `.map(\.endpoint)` on both sides is not buying anything and could stay a plain request comparison. Not worth restructuring the other five over.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
Atharva8168: Follow-up now that I have read `StateManagementTests.swift` (its diff was truncated in the PR view, so I pulled the branch) — this makes the gap sharper rather than softer.

You wrote a canonical-token regression gate for **both** of the sites you fixed:

- `testResetProfilePreservesCanonicalPushToken` — `XCTAssertNotNil(IdentityStore.shared.pushToken, "resetProfile must not clear the canonical push token in IdentityStore")`
- `testFlushWithPendingProfileKeepsCanonicalToken` — `"flush with a pending profile must not wipe the canonical push token"`

There is no equivalent for `enqueueProfile`, which is the third site with the same shape and the one that still nils the projection without restoring it.

I also checked whether the new `testEnqueueProfilePayloadParityWithLegacyBuilder` happens to cover it incidentally — it does not. It seeds from `INITIALIZED_TEST_STATE()`, which carries no identifiers, so `state.isIdentified` is false and the `reset(preserveTokenData: false)` branch is never entered. The bug needs `isIdentified == true` **and** `identifiersChanged == true`, which is exactly the setup `identifiedState(...)` produces in `StateManagementEnqueueEdgeCaseTests` — and those tests assert `readQueue()` contents only.

So the test would slot straight into the existing pattern, e.g.:

```swift
func testSetProfileWithChangedIdentifiersPreservesCanonicalPushToken() async throws {
    let initialState = identifiedState(email: "old@email.com")
    SDKConfigStore.shared.update(KlaviyoConfig(apiKey: TEST_API_KEY))
    IdentityStore.shared.update(initialState.identity)
    IdentityStore.shared.updatePushToken(initialState.pushTokenData!)
    store.exhaustivity = .off

    _ = await store.send(.enqueueProfile(Profile(email: "new@email.com")))

    XCTAssertNotNil(
        IdentityStore.shared.pushToken,
        "set(profile:) with changed identifiers must not clear the canonical push token"
    )
}
```

That fails on the current branch and passes with the one-line `state.pushTokenData = tokenData` restore.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
belleklaviyo: reworded comment

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
belleklaviyo: added fix and regression test

glenn-klaviyo (approved):

PR #723, opened 2026-09-10
feat(KlaviyoCore): lift the flush/retry engine onto an actor RequestQueue — unwired (MAGE-1197)

belleklaviyo:
Note: Diff is kinda terrifying but ~1200 lines are purely new tests/test support 😅 

# Description
Lift the flush / retry / connectivity engine off the vendored-TCA reducer onto an `actor RequestQueue` in **KlaviyoCore** (MAGE-1197, PR 1 of 2 — the *engine*). Built alongside the reducer and unit-tested for parity, but **UNWIRED**: nothing in production calls it yet, so the reducer flush engine remains fully authoritative and there is no runtime behavior change. The cutover (wiring + reducer deletion) is a follow-up PR.

## Due Diligence

- [ ] `Patch`
- [x] `Minor` Adds public API to KlaviyoCore (`RequestQueue`, `SleepClock`, `classifyFailure`) — but all unwired, so no behavior change yet.
- [ ] `Major`
- [x] This is planned work for an upcoming release (part of `feat/modularization-2`; final release categorization happens when that branch merges).

## Changelog / Code Overview
Stacks on `feat/modularization-2` (after #722 / MAGE-1196). New KlaviyoCore code, all additive/unwired:

- **`SleepClock`** — injectable closure-based timing seam (à la `QueueStore.PersistScheduler`) for interval/backoff sleeps; production wraps `Task.sleep(nanoseconds:)`, tests inject an immediate/recording variant. Not the stdlib `Clock` (iOS 16+); SDK floor is iOS 13. Clamps non-finite/overflow durations.
- **Type move to Core** — `RetryState`, flush-timing constants, `KlaviyoEndpoint.maxRetries`, `InvalidField`/`parseError`, and the error-response models moved from KlaviyoSwift → KlaviyoCore (reducer keeps compiling via import). `StateManagementConstants` kept as a thin alias.
- **`classifyFailure`** — Core failure→decision function; parity twin of the reducer's `handleRequestError` (minus action wrapping).
- **`actor RequestQueue`** — run-loop heartbeat; `flush()` = `willDrain` seam → drain `QueueStore` → sequential FIFO send → success dequeue + push-token write-back to `IdentityStore`; retry / direct-sleep backoff / maxRetries drop / lease restore; 4xx invalid-field clear on the canonical store; connectivity-driven interval + loop cancel/restart. `isFlushing` reentrancy guard + post-`send` join-point empty-lease guard.

**Intended behavioral divergence from the reducer (one):** direct-sleep backoff instead of the timer-era per-tick countdown — observable retry timing (exponential + jitter, 300s cap) is equivalent.

**Deferred to the cutover PR (documented in the spec + cutover plan):** make `IdentityStore` writer-vs-writer safe + add an atomic field-clear (it's only single-writer safe today; harmless while the actor is unwired); wire the actor into the environment/lifecycle; relocate pending-profile staging; delete the 9 reducer flush actions.

## Test Plan
- `xcodebuild test -scheme klaviyo-swift-sdk-Package -destination "platform=iOS Simulator,name=iPhone 17 Pro"` → **354 KlaviyoSwift + 327 KlaviyoCore tests, 0 failures.**
- New: `RequestQueueTests` (27) + `FlushDecisionTests` (16). Parity + concurrency covered: willDrain-before-drain, FIFO, token write-back, transient retry (next tick), rate-limit/server backoff (sleep + in-place resend), maxRetries drop (both retry + backoff branches), 4xx email/phone clear (sibling fields survive), connectivity interval switching + coalescing, offline lease restore, reentrancy clobber, and stop-during-send / stop-during-backoff-sleep races.
- Deterministic — injected `SleepClock`, no wall-clock waits.

## Related Issues/Tickets
- Linear: MAGE-1197 (parent MAGE-904). Follows MAGE-1196 (#722).

## New Features
- Added automatic FIFO request processing with scheduled and immediate flushing.
- Added connectivity-aware delivery cadence and lifecycle controls.
- Added reliable retry handling for network, rate-limit, and server failures, including backoff and retry limits.
- Added persistence for successful push-token registrations.

## Bug Fixes
- Invalid email or phone fields are cleared when rejected by the API while preserving other identity data.
- Improved protection against duplicate or concurrent queue processing.
- Requests are restored safely after cancellation or retryable failures.


Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -0,0 +1,245 @@
...
+    /// Transport seam: sends one request with its per-attempt retry metadata and reports the result.
+    public typealias Send = @Sendable (KlaviyoRequest, RequestAttemptInfo)
+        async -> Result<Data, KlaviyoAPIError>
+
+    // MARK: - Injected dependencies
+
+    private let clock: SleepClock
+    private let send: Send
+    /// Optional hook invoked before each drain; wired by a later task.
+    private let willDrain: (@Sendable () async -> Void)?
```
belleklaviyo, line 27: this is here to replace the effective pendingProfile batching we previously had. previously not all profile attribute setters sent a request per property, they'd be batched into one profile request before flush. this maintains that parity by allowing a seam before flush that folds in the pending profile attributes before draining. tried to keep name profile-agnostic as that lives in KlaviyoSwift while this is in KlaviyoCore

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -0,0 +1,245 @@
...
+                }
+                requestsInFlight.removeFirst()
+                retryState = .retry(FlushConstants.initialAttempt)
+
+            case let .failure(error):
+                // Classify the failure and act on it. Mirrors `handleRequestError` +
+                // `requestFailed`/`deQueueCompletedResults` in the legacy reducer:
+                // non-retryable errors dequeue the head and CONTINUE the flush; retryable
+                // errors set retryState, drop the head if it exceeded `maxRetries`, then STOP
+                // (network retries) or SLEEP+retry-in-place (backoff).
```
belleklaviyo: all of these comments btw will be cleaned up when we remove the legacy reducer, but inline docs referencing the legacy reducer just helps review and development for now

Sources/KlaviyoCore/RequestQueueSupport.swift
```diff
@@ -0,0 +1,200 @@
+//
+//  RequestQueueSupport.swift
```
belleklaviyo, line 2: most of these types were just moved from StateManagement, APIRequestErrorHandling, and Error (all in KlaviyoSwift) to here in KlaviyoCore

Sources/KlaviyoCore/RequestQueueSupport.swift
```diff
@@ -0,0 +1,200 @@
...
+    let title: String
+    let detail: String
+    let source: ErrorSource
+}
+
+struct ErrorSource: Codable {
+    let pointer: String
+}
+
+// MARK: - FlushDecision
```
belleklaviyo, line 105: FlushDecision and classifyFailure are the only real net new support types in this file

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -0,0 +1,260 @@
...
+    /// (`classifyFailure`) and handled: non-retryable → dequeue + continue; transient → stop + lease
+    /// restore (retries next tick); rate-limit/server → direct-sleep backoff then retry in place.
+    private func flush() async {
+        // 1. Gate: pre-init or offline. Do not flush.
+        guard SDKConfigStore.shared.current.apiKey != nil, flushInterval.isFinite else { return }
+        guard !isFlushing else { return }
+        isFlushing = true
+        defer { isFlushing = false }
+
+        // 2. Let the owner enqueue any last-minute requests before we take the snapshot.
```
belleklaviyo, line 125: slightly misleading comment, will clean it up when its actually hooked up (next PR) as this is only going to really be used for the pendingProfile request

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -0,0 +1,260 @@
...
+            } catch {
+                environment.emitDeveloperWarning("Invalid RequestAttemptInfo parameters: \(error)")
+                restoreLease()
+                return
+            }
+
+            let outcome = await send(head, attemptInfo)
+            guard !requestsInFlight.isEmpty else { return }
+            switch outcome {
+            case .success:
```
glenn-klaviyo, line 155: I see we're not doing anything with the success (Data) here. is that intentional?

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -0,0 +1,260 @@
...
+/// through an injected transport. Owns its state and lifecycle (`start`/`stop`), runs the timed run
+/// loop, and implements the drain/send happy path: `flush()` leases the pending batch and sends it
+/// head-first in FIFO order, writing a registered push token back to `IdentityStore` on success.
+/// On failure it classifies the error and either dequeues + continues (non-retryable), stops and
+/// restores the lease so retries resume next tick (transient), or sleeps the backoff and retries
+/// the same head in place (rate-limit / server error).
+public actor RequestQueue {
+    /// Transport seam: sends one request with its per-attempt retry metadata and reports the result.
+    public typealias Send = @Sendable (KlaviyoRequest, RequestAttemptInfo)
+        async -> Result<Data, KlaviyoAPIError>
```
glenn-klaviyo, line 20: thought: wondering if maybe this would be better suited as `async throws(KlaviyoAPIError) -> Data` and then below we don't have to switch on success or failure and instead just use a `try { } catch { }`. Feels like the more "swift concurrency" way to do this 🤔

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -0,0 +1,260 @@
...
+    private var flushInterval: TimeInterval = FlushConstants.wifiFlushInterval
+    /// Retry bookkeeping for the request currently being sent.
+    private var retryState: RetryState = .retry(FlushConstants.initialAttempt)
+    /// The active run loop, or `nil` when stopped.
+    private var runLoop: Task<Void, Never>?
+    private var isFlushing = false
+
+    public init(clock: SleepClock,
+                send: @escaping Send,
+                willDrain: (@Sendable () async -> Void)? = nil) {
```
glenn-klaviyo: small formatting nit: personal preference
```suggestion
    public init(
        clock: SleepClock,
        send: @escaping Send,
        willDrain: (@Sendable () async -> Void)? = nil
    ) {
```
feel free to make this change or ignore since i see this pattern in other places.

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -0,0 +1,260 @@
...
+                requestsInFlight.removeFirst()
+                retryState = .retry(FlushConstants.initialAttempt)
+
+            case let .failure(error):
+                // Classify the failure and act on it. Mirrors `handleRequestError` +
+                // `requestFailed`/`deQueueCompletedResults` in the legacy reducer:
+                // non-retryable errors dequeue the head and CONTINUE the flush; retryable
+                // errors set retryState, drop the head if it exceeded `maxRetries`, then STOP
+                // (network retries) or SLEEP+retry-in-place (backoff).
+                switch classifyFailure(error: error, retryState: retryState) {
```
glenn-klaviyo: could we move this switch into it's own function? this `flush` function is getting quite large.

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -0,0 +1,260 @@
...
+        let attributes = payload.data.attributes
+        let enablement = PushEnablement(rawValue: attributes.enablementStatus) ?? .authorized
+        let background = PushBackground(rawValue: attributes.backgroundStatus) ?? .available
+        return PushTokenData(
+            pushToken: attributes.token,
+            pushEnablement: enablement,
+            pushBackground: background,
+            deviceData: attributes.deviceMetadata
+        )
+    }
```
glenn-klaviyo, line 262: to avoid these "mapping functions" we could instead make use of the "failable" initializer by making this an extension off of PushTokenData. For example:

```
extension PushTokenData {
    /// Reconstructs a ``PushTokenData`` from a registered-push-token request payload so a successful
    /// registration can be written back to `IdentityStore`.
    init?(_ payload: PushTokenPayload) {
        let attributes = payload.data.attributes
        let enablement = PushEnablement(rawValue: attributes.enablementStatus) ?? .authorized
        let background = PushBackground(rawValue: attributes.backgroundStatus) ?? .available
        self = PushTokenData(
            pushToken: attributes.token,
            pushEnablement: enablement,
            pushBackground: background,
            deviceData: attributes.deviceMetadata
        )
    }
}
```
This extension could live in `PushTokenData.swift`

Then line 155 becomes:
```
if case let .registerPushToken(_, payload) = head.endpoint {
    IdentityStore.shared.updatePushToken(PushTokenData(payload))
}
```
instead of 
```
if case let .registerPushToken(_, payload) = head.endpoint {
    IdentityStore.shared.updatePushToken(pushTokenData(from: payload))
}
```

small, but i think it improves the code.

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -0,0 +1,260 @@
...
+            requestsInFlight = []
+        }
+    }
+
+    /// Immediate flush outside the timed cadence (high-priority path).
+    public func flushNow() async {
+        await flush()
+    }
+
+    public func networkConnectivityChanged(_ status: Reachability.NetworkStatus) {
```
glenn-klaviyo, line 96: is this only used for tests?

glenn-klaviyo (commented):
just finished a first round of reviewing. left a few comments

Sources/KlaviyoCore/RequestQueue.swift
belleklaviyo: will be used when it's hooked up!

Sources/KlaviyoCore/RequestQueue.swift
belleklaviyo: yeah this was just to maintain as much parity while moving it out from the existing `sendRequest`'s body but it isn't actually being used. it's intentional to maintain it though so it works with the existing `klaviyoApi.send` signature, can maybe be cleaned up later?

Sources/KlaviyoCore/RequestQueue.swift
belleklaviyo: hmm I see what you mean but I think the typed throws are not available in Swift lang 5.9? maybe we can revisit this pattern when we do support Swift 6 :')

Sources/KlaviyoCore/RequestQueue.swift
belleklaviyo: hahah longtime disagreer of this formatting buut fair as we have the precedent in other spots, will align

glenn-klaviyo: hahah i'm also fine with keeping the way it is. saves us a couple lines in the file

Sources/KlaviyoCore/RequestQueue.swift
glenn-klaviyo: ah i see that explains it. i'm fine with the way it is. we can look to clean it up if it causes problems or confusion in the future.

Sources/KlaviyoCore/RequestQueue.swift
glenn-klaviyo: ah yes, i forget that it's in 6 and not 5.9. you can ignore my msg but yeah would be good to support this pattern at a later stage.

Sources/KlaviyoCore/RequestQueue.swift
belleklaviyo: I actually think it should be non-failable  as both token and deviceMetadata on the attributes are non-optional, but I can still make the init to clean up the code a bit!

glenn-klaviyo: thanks for looking into this. that sounds good to me 👍

glenn-klaviyo (approved):
LGTM. thanks for addressing comments. Nice work. There's one coderabbit comment that seemed like nonblocking on this PR

belleklaviyo:
@glenn-klaviyo yeah actually gonna address that separately in a different PR as an improvement already on my radar 👍 thanks so much for the review 🫡

PR #724, opened 2026-09-11
refactor(KlaviyoCore): durable countdown backoff for RequestQueue (MAGE-1197)

belleklaviyo:
# Description
Replace the `RequestQueue` actor's direct-sleep backoff with a **durable countdown** (pre-modularization parity). In #732 we introduced a direct-sleep model where in if a request is rate limited, it is only parked in-memory while waiting for the backoff time to pass, in which exactly on expiry it retries. That differs from the pre-existing behavior where we stored a rate limited request on disk during wait and flushed it on the next scheduled flush if its backoff has passed/expired. tldr;

| Model | Durable (survives crash)? | Precise timing? |
|---|---|---|
| Existing (countdown on ticks) | ✅ on disk during wait | ❌ retries on the next tick (up to +1 interval) |
| Introduced in #732 | ❌ in-memory during wait | ✅ retries at ~exactly the backoff |
| New "durable countdown" | ✅ | ❌ (same as before) |

Still **unwired** — no app behavior change.

## Due Diligence

- [x] `Patch` — internal engine change, unwired (no public API or behavior change yet).
- [x] This is planned work for an upcoming release (part of `feat/modularization-2`).

## Changelog / Code Overview
The engine PR (#723) shipped a **direct-sleep** backoff: on a rate-limit/5xx it held the request in the in-memory lease and `await clock.sleep(backoff)`s in place. Review (CodeRabbit + Cursor) surfaced that this is crash-lossy (lease in memory for up to ~300s) and cancellation-fragile (a `start()` mid-backoff bypasses the wait). This PR swaps it for the reducer's proven **countdown** algorithm — ported to the actor, no TCA:

- **Failure branch:** on rate-limit/5xx, record the backoff on `retryState` and **restore the request to the durable `QueueStore`**, then stop the flush (no `clock.sleep` in the backoff path).
- **Countdown gate** at the top of `flush()`: each run-loop tick decrements the backoff by `flushInterval`; when it reaches zero it promotes to `.retry(requestCount)` and drains + sends.
- The request stays **on disk during the entire wait** (crash-safe); a `start()`/`stop()` mid-backoff just resumes the countdown on the next tick (`retryState` is actor state) — no immediate-retry bypass.
- Exceeded-`maxRetries` on backoff resets to `.retry(initialAttempt)` (NOT `requestCount: 0`) — a deliberate divergence from the reducer, whose reset would promote to `.retry(0)` and stall `RequestAttemptInfo`.
- Removed the now-dead `FlushDecision.retryWithBackoff` `seconds:` value and the `RecordingSleepClock` test double (no backoff sleep to record). `SleepClock` is now used only for the run-loop interval.

Trade-off: retry timing is tick-coarse (fires on the next tick after the backoff expires, up to one interval late) — accepted, matches the reducer.

## Test Plan
- `xcodebuild test -scheme klaviyo-swift-sdk-Package` → **KlaviyoCore suite green** (329 tests, 0 failures).
- Backoff tests rewritten to the countdown model (flushNow-as-tick): a rate-limited request is durable in `QueueStore` during the wait, skipped for N ticks, then resent with an advanced attempt number; exceeded-on-backoff drops the request and a successor still sends (no stall).

## Related Issues/Tickets
- Linear: MAGE-1197 (parent MAGE-904). Follow-up to merged #723.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of rate limits and server errors by preserving queued requests during backoff periods.
  * Requests now remain durable across processing cycles and shutdowns, then retry automatically after the backoff countdown expires.
  * Retry tracking now resets cleanly after the maximum retry limit is exceeded.

### DIFF
--- a/Sources/KlaviyoCore/RequestQueue.swift
+++ b/Sources/KlaviyoCore/RequestQueue.swift
@@ -124,20 +124,10 @@ public actor RequestQueue {
         isFlushing = true
         defer { isFlushing = false }
 
-        // Durable countdown backoff: if a prior failure set a backoff, decrement it by one tick and skip
-        // the flush until it elapses. The failing request is already restored to QueueStore (below), so it
-        // stays durable during the wait. On expiry, promote to `.retry(requestCount)` and fall through to
-        // drain + send. Ports the reducer's flushQueue countdown (no TCA).
-        if case let .retryWithBackoff(requestCount, totalCount, backoff) = retryState {
-            let remaining = max(backoff - Int(flushInterval), 0)
-            if remaining > 0 {
-                retryState = .retryWithBackoff(requestCount: requestCount,
-                                               totalRetryCount: totalCount,
-                                               currentBackoff: remaining)
-                return
-            }
-            retryState = .retry(requestCount)
-        }
+        // Durable countdown backoff gate. Advances the countdown once per flush; if a backoff is still
+        // outstanding, skip this flush (the failing request stays durable in `QueueStore` during the
+        // wait). See `advanceBackoffGate`.
+        if case .wait = advanceBackoffGate() { return }
 
         // 2. Let the owner enqueue any last-minute requests before we take the snapshot.
         await willDrain?()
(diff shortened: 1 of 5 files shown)


Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -114,14 +115,20 @@ public actor RequestQueue {
...
         isFlushing = true
         defer { isFlushing = false }
 
+        // Durable countdown backoff gate. Advances the countdown once per flush; if a backoff is still
+        // outstanding, skip this flush (the failing request stays durable in `QueueStore` during the
+        // wait). See `advanceBackoffGate`.
+        if case .wait = advanceBackoffGate() { return }
+
         // 2. Let the owner enqueue any last-minute requests before we take the snapshot.
         await willDrain?()
```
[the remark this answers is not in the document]
belleklaviyo: this is already a known quirk and separately going to be looked at in https://linear.app/klaviyo/issue/MAGE-1021/ios-align-retry-delay-scheduling-with-android-tick-based-countdown as part of the queueing improvements project. this behavior currently maintains parity with master

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -179,6 +200,36 @@ public actor RequestQueue {
...
+    }
+
+    /// Advances the durable countdown backoff by one flush interval and reports whether `flush()`
+    /// should skip this pass. Called once per flush — from a scheduled run-loop tick OR an immediate
+    /// `flushNow()` — which mirrors the reducer, whose `flushQueue` decremented the backoff on every
+    /// dispatch (timer and high-priority paths alike). Timing is therefore approximate in both
+    /// directions: a backoff fires at most one interval late on the tick path, and a burst of
+    /// immediate flushes can expire it early. That imprecision is the accepted cost of reducer parity.
+    /// The failing request is already restored to the durable `QueueStore`, so it survives the wait.
+    private func advanceBackoffGate() -> BackoffGate {
```
belleklaviyo, line 217: tldr; on each flush(), we are deciding whether or not to actually flush or block it based on if there is a backoff/wait period currently running and if that has expired by the time of checking this gate. expiry is checked based on decrementing the value in this gate check, not comparing against objective wall clock timing.

small issue is flush() can be called directly by high priority events, so when they call flush(), this check and decrement still run, so backoff could be unfaithfully underwaited. this is already ticketed to be addressed in https://linear.app/klaviyo/issue/MAGE-1021/ios-align-retry-delay-scheduling-with-android-tick-based-countdown as this is existing behavior

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -232,32 +283,22 @@ public actor RequestQueue {
...
-            // Decision 2: sleep the backoff directly, then retry the SAME head in place (rather than
-            // restoring + waiting for the reducer's per-tick countdown). The `isFlushing` guard
-            // stays true across the sleep, so no concurrent flush runs.
-            try? await clock.sleep(Double(seconds))
-            // Promote back to `.retry(requestCount)` so the in-place retry advances `attemptNumber`
-            // (parity with the reducer's backoff-expiry: `state.retryState = .retry(requestCount)`).
-            // Without this the retried send would keep sourcing `numAttempts` as the initial attempt.
-            if case let .retryWithBackoff(requestCount, _, _) = retryState {
-                retryState = .retry(requestCount)
+                retryState = .retry(FlushConstants.initialAttempt)
```
belleklaviyo, line 298: so this is effectively .retry(1) and setting it correctly to the next attempt route

glenn-klaviyo (approved):
LGTM

PR #725, opened 2026-09-14
refactor: cut over to the RequestQueue actor as the sole flush engine (MAGE-1197 PR2)

belleklaviyo:
# Description
MAGE-1197 PR2 (of 2) — the **cutover**. Wires the Core `RequestQueue` actor (shipped unwired in #723/#724) into the SDK as the **sole flush engine** and deletes the reducer's flush/retry/connectivity engine. This is the behavior-affecting half; the engine + durable-countdown backoff already landed on `feat/modularization-2`.

## Due Diligence

- [x] `Minor` — internal engine cutover; no public API change, behavior parity with `master`.
- [x] Planned work for an upcoming release (part of `feat/modularization-2`).

## Changelog / Code Overview
The actor now owns draining/sending; the reducer no longer flushes. Landed as 7 focused commits:

1. **`ProfilePropertyBuffer`** — KlaviyoSwift-side staging for `setProfileProperty`, folded into a request at the actor's `willDrain` (mirrors the old `pendingProfile` branch, sourced from `IdentityStore`).
2. **Bootstrap** — a single shared `RequestQueue` on `klaviyoSwiftEnvironment` wired with `.production` clock / `klaviyoAPI.send` / `willDrain` → `ProfilePropertyBuffer.flushIntoQueue`.
3. **Repoint all triggers** — `completeInitialization` drives `requestQueue.start/stop/networkConnectivityChanged` from the lifecycle stream (keeping push-enablement + badge side effects, byte-for-byte); company-switch + high-priority events → `flushNow()`; `setProfileProperty` → `ProfilePropertyBuffer.stage`. Adds a `RequestQueueProtocol` seam + `SpyRequestQueue` so the wiring is unit-testable.
4. **Delete the dead engine** — removes the 9 reducer flush actions + arms, `RequestId`/`FlushTimer`, `handleRequestError`; relocates/prunes their tests (behavior now covered by Core `RequestQueueTests`/`FlushDecisionTests`).
5. **Drop dead `KlaviyoState` fields** — `requestsInFlight`/`flushing`/`flushInterval`/`retryState`/`pendingProfile` and the now-dead pending-profile fold plumbing.
6. **`IdentityStore` writer-safety** (prereq) — a serializing `writeLock` (concurrent writers can't diverge disk from last-emit) + atomic `mutate` (closes the 4xx read-modify-write TOCTOU); the actor becomes a 2nd concurrent identity writer, so this is required before wiring.
7. **Cleanup** — removes the orphaned `environment.timer` (fed only the deleted reducer flush timer).

**Scope fence:** `KlaviyoState`, the write-through `defer`, and vendored TCA remain (their removal is MAGE-904).

**Intended behavior change (documented):** staged profile properties now ship via the buffer's `willDrain` (a separate request on the same flush) rather than being folded into an identifier-change/token request. Properties still ship; parity otherwise holds.

## Test Plan
- `xcodebuild test -scheme klaviyo-swift-sdk-Package` → **full suite green** (KlaviyoSwift 328 + KlaviyoCore 331 + Forms/Location/others, 0 failures), deterministic across runs.
- New: `RequestQueueLifecycleTests`, `ProfilePropertyBufferTests`, `RequestQueueBootstrapTests`, `SpyRequestQueue`.
- The `.test()` clock makes the actor run loop a no-op so no background draining leaks across tests; `flushNow()` drives real drains explicitly.

## Related Issues/Tickets
- Linear: MAGE-1197 (parent MAGE-904). Follow-up to merged #723 (engine) / #724 (durable countdown).
- Out of scope (separate tickets): retry-timing precision / wall-clock deadline + backoff floor — **MAGE-1021** / **MAGE-1209**.

## Summary by CodeRabbit

- **New Features**
  - Profile properties are safely buffered and included when identity or push-token data becomes available.
  - Request delivery responds to app lifecycle and network connectivity changes, with immediate flushing for priority events.
  - Lifecycle events are delivered through an asynchronous event stream.
  - Request queue operations support coordinated starting, stopping, flushing, and connectivity updates.
  - Identity updates are serialized for consistent persisted and in-memory data.

- **Bug Fixes**
  - Improved consistency for concurrent identity updates.
  - Profile and push-token requests use the latest canonical identity information.
  - Improved request queue flushing and delivery reliability.


Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -254,18 +256,14 @@ public actor RequestQueue {
...
             // canonical store so the next request to the API won't carry a stale bad value.
-            // NOTE: read-modify-write is a TOCTOU vs any other IdentityStore writer. Safe here
-            // only because the actor is unwired in this PR. The request-queue cutover must make
-            // IdentityStore concurrent-writer-safe and give it an atomic field-clear; see
-            // IdentityStore's SINGLE WRITER note.
-            var identity = IdentityStore.shared.current
-            for field in fields {
-                switch field {
-                case .email: identity.email = nil
-                case .phone: identity.phoneNumber = nil
```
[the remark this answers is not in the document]
belleklaviyo: the `defer` write will be going away with the final removal (after this PR) so this double write won't actually be an issue, easier to just leave for now and it will go away with the reducer removal

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -254,18 +256,14 @@ public actor RequestQueue {
...
+            IdentityStore.shared.mutate { identity in
+                for field in fields {
+                    switch field {
+                    case .email: identity.email = nil
+                    case .phone: identity.phoneNumber = nil
+                    }
```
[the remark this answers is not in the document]
belleklaviyo: same as above, this is just because we are still maintaing the defer write block in our reducer, but that will be 🪓 'd once and for all in the next PR

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -110,35 +108,25 @@ public actor RequestQueue {
...
-        // Durable countdown backoff gate. Advances the countdown once per flush; if a backoff is still
-        // outstanding, skip this flush (the failing request stays durable in `QueueStore` during the
-        // wait). See `advanceBackoffGate`.
+        // Advance the durable countdown backoff once per flush; skip while one is outstanding (the
+        // failing request stays durable in `QueueStore` during the wait). See `advanceBackoffGate`.
         if case .wait = advanceBackoffGate() { return }
 
-        // 2. Let the owner enqueue any last-minute requests before we take the snapshot.
+        // Let the owner enqueue last-minute requests, then lease everything (incl. those).
         await willDrain?()
```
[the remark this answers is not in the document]
belleklaviyo: Fixed in 6ef13637 — re-check `Task.isCancelled` after `await willDrain?()` before `drainAll()`, so a `stop()` during that suspension no longer drains + sends.

Sources/KlaviyoSwift/ProfilePropertyBuffer.swift
```diff
@@ -0,0 +1,137 @@
...
+            from: pendingProfile, into: attributes.location ?? .init()
+        )
+        basePayload = .init(data: .init(attributes: attributes))
+        RequestEnqueuer.enqueueProfile(payload: basePayload)
+    }
+
+    /// Drops all staged properties. Called from `KlaviyoState.reset()` (profile reset / company
+    /// switch / profile-clobber) so staged props never leak onto a new identity — parity with the
+    /// old reducer clearing `pendingProfile`. Also used for test isolation.
+    func reset() {
```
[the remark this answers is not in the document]
belleklaviyo: Fixed in 6ef13637 — added a generation counter: captured with the snapshot, bumped in `reset()`, and checked before enqueue/restore, so a `reset()` racing an in-flight flush drops the stale snapshot.

Sources/KlaviyoCore/RequestQueue.swift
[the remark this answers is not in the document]
belleklaviyo: Known coexistence limitation — the reducer `state.identity` projection is deleted by MAGE-904 (in progress), which removes this race at its root. Documented on the write-through `defer`; not patched here since any fix lives in code 904 deletes.

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -113,23 +84,25 @@ enum KlaviyoAction: Equatable {
...
         let previousApiKey = state.apiKey
         let previousIdentity = state.identity
         let previousPushTokenData = state.pushTokenData
+        // KNOWN COEXISTENCE LIMITATION (resolved by MAGE-904): this defer persists the reducer's
+        // `state.identity` PROJECTION wholesale. The `RequestQueue` actor can clear a rejected
+        // identifier (4xx field-clear) directly on `IdentityStore` out of band; since the projection
+        // is not re-read here, a later identity-mutating action can persist the stale projection and
+        // resurrect the cleared value. Exists only while both writers (reducer projection + actor)
+        // coexist. MAGE-904 deletes this reducer / projection, making `IdentityStore` the sole source
+        // of truth — do not patch it here (any fix lives in code 904 removes).
```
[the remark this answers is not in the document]
belleklaviyo: Same root as the write-through `defer` coexistence limitation — resolved by MAGE-904 (deletes the `state.identity` projection). Documented in-code; not patched in this PR.

Sources/KlaviyoCore/IdentityStore.swift
```diff
@@ -127,21 +154,27 @@ public final class IdentityStore: IdentityReading, IdentityWriting {
...
-        lock.withLock {
-            hydrated = false
-            pushTokenValue = nil
+        writeLock.withLock {
+            lock.withLock {
+                hydrated = false
+                pushTokenValue = nil
+                // Delete under `lock` so the file removal can't interleave with a concurrent
+                // hydrate/persist. Emit stays outside `lock` (INVARIANT), inside `writeLock`.
+                removePersisted(fileName: StoreFile.identity)
```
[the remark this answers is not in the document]
belleklaviyo: Acknowledged — `reset()` is test-isolation only (cross-test bleed, not a production path). Leaving as-is for now.

Tests/KlaviyoSwiftTests/GeofenceEventDispatchTests.swift
```diff
@@ -94,29 +93,48 @@ final class GeofenceEventDispatchTests: XCTestCase {
...
+        IdentityStore.shared.update(state.identity)
+        IdentityStore.shared.updatePushToken(state.pushTokenData)
+        _ = makeTestStore(initialState: state)
+
+        // Capture what the actor sends through the transport.
+        let sentRequests = ThreadSafeBox<[KlaviyoRequest]>([])
+        environment.klaviyoAPI.send = { request, _ in
+            sentRequests.mutate { $0.append(request) }
+            return .success(Data())
         }
```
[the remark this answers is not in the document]
belleklaviyo: Leaving as-is — minor test refactor on the geofence tests being set aside; not blocking.

Sources/KlaviyoCore/IdentityStore.swift
```diff
@@ -29,24 +37,22 @@ public protocol IdentityWriting {
...
-    // `pushToken`) during delivery would deadlock. Always mutate under the lock, then emit outside it.
+    // TWO LOCKS (mirrors `QueueStore`'s `persistLock`/`queueLock` split):
+    //
+    // `writeLock` serializes a whole write (persist THEN emit) so two writers can't persist in one
+    // order but emit in another, diverging disk from the last-emitted value. Only writers take it.
+    //
+    // `lock` (non-recursive `UnfairLock`) guards `hydrated`, `pushTokenValue`, and disk I/O for short
+    // critical sections. INVARIANT: never hold `lock` across `subject.send` — Combine delivers
+    // synchronously, so a subscriber reading a `lock`-guarded accessor (e.g. `pushToken`) during
+    // delivery would deadlock. Mutate/persist under `lock`, release, then emit.
```
belleklaviyo, line 48: since we are making the setters effectively write to IdentityStore directly now, there can be multiple writers, so we needed to make it writer-vs-writer safe. 

it does seem risky at first look but there's not really a real concern for a dead lock situation as
1) readers won't deadlock ever
2) the write paths from a consumer pov won't hit this and
3) the only potential ways to deadlock are if a synchronous subscriber writes back during delivery or a mutate transform calls the store, both of which are documented scenarios for this to avoid

Sources/KlaviyoCore/KlaviyoEnvironment.swift
```diff
@@ -41,7 +41,6 @@ public struct KlaviyoEnvironment {
         timeZone: @escaping () -> String,
         appContextInfo: @escaping () -> AppContextInfo,
         klaviyoAPI: KlaviyoAPI,
-        timer: @escaping (Double) -> AnyPublisher<Date, Never>,
```
belleklaviyo, line 44: no longer using this timer in the reducer

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -7,14 +7,12 @@
...
-/// On failure it classifies the error and either dequeues + continues (non-retryable), stops and
-/// restores the lease so retries resume next tick (transient), or records a durable countdown
-/// backoff and restores the request to the queue (rate-limit / server error) — the request stays
-/// on disk while the countdown gate at the top of `flush()` waits it out over ticks, then resends.
+/// Core-owned flush engine: drains `QueueStore.shared` on a timed cadence and sends through an
+/// injected transport. `flush()` leases the pending batch and sends it head-first (FIFO), writing a
+/// registered push token back to `IdentityStore` on success. On failure it classifies the error and
+/// either dequeues + continues (non-retryable), restores the lease and stops so retries resume next
+/// tick (transient), or records a durable countdown backoff and restores the request so the
+/// countdown gate in `flush()` waits it out over ticks before resending (rate-limit / server error).
```
belleklaviyo, line 15: just making it more succinct

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -110,35 +108,26 @@ public actor RequestQueue {
...
+        // failing request stays durable in `QueueStore` during the wait). See `advanceBackoffGate`.
         if case .wait = advanceBackoffGate() { return }
 
-        // 2. Let the owner enqueue any last-minute requests before we take the snapshot.
+        // Let the owner enqueue last-minute requests, then lease everything (incl. those).
         await willDrain?()
-
-        // 3. Lease the whole pending batch. Anything enqueued by `willDrain` above is already in the
-        //    store, so it is included in this drain.
+        guard !Task.isCancelled else { return }
```
belleklaviyo, line 127: prevents interference if stop() is called at any point mid flush

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -284,13 +264,11 @@ public actor RequestQueue {
...
-            // over ticks, then resends. If it already exceeded `maxRetries`, drop the head and reset
-            // to a fresh `.retry(initialAttempt)`. This DELIBERATELY diverges from the reducer's
-            // `.retryWithBackoff(requestCount: 0)` reset — under the countdown gate that would promote
-            // to `.retry(0)`, which `RequestAttemptInfo` rejects → a permanent stall. Do NOT restore
-            // that parity.
+            // Rate-limit / server error. Record the backoff; the countdown gate waits it out then
+            // resends. If past `maxRetries`, drop the head and reset to `.retry(initialAttempt)`.
+            // This DELIBERATELY diverges from the reducer's `.retryWithBackoff(requestCount: 0)`
+            // reset — under the countdown gate that promotes to `.retry(0)`, which
+            // `RequestAttemptInfo` rejects → a permanent stall. Do NOT restore that parity.
```
belleklaviyo, line 271: overall just comment cleanup

Sources/KlaviyoCore/RequestQueueProtocol.swift
```diff
@@ -0,0 +1,23 @@
...
+
+import Foundation
+
+/// Seam over the concrete `RequestQueue` actor so lifecycle wiring can be exercised against a test
+/// double that records `start`/`stop`/`flushNow`/`networkConnectivityChanged` invocations.
+///
+/// `RequestQueue`'s synchronous actor methods satisfy the `async` requirements here: cross-actor
+/// calls are already `await`ed at the call site, so the protocol can declare them `async` while the
+/// actor implements `start()`/`stop()`/`networkConnectivityChanged(_:)` synchronously.
+public protocol RequestQueueProtocol: Sendable {
```
belleklaviyo, line 16: this is just so lifecycle wiring can be unit-tested against a spy (the concrete actor can't be subclassed)

Sources/KlaviyoSwift/StateManagement/StateManagement.swift
```diff
@@ -253,10 +227,44 @@ struct KlaviyoReducer: ReducerProtocol {
...
+                    case .foregrounded:
+                        await handleForeground()
+                    case .backgrounded, .terminated:
+                        await handleBackground()
+                    case let .reachabilityChanged(status):
+                        await klaviyoSwiftEnvironment.requestQueue.networkConnectivityChanged(status)
+                    }
+                }
             }
-            .merge(with: environment.lifecycleEventsWithReachability().map(\.transformToKlaviyoAction).eraseToEffect())
```
belleklaviyo, line 259: we no longer need to transform it into a reducer action, we can just listen to the lifecycle event stream directly and dispatch whatever effects to the actor

Sources/KlaviyoSwift/Models/LifecycleEventsExtension.swift
```diff
@@ -5,20 +5,22 @@
...
+    /// `completeInitialization` effect. `AnyPublisher.values` requires iOS 15; this `sink`-based
+    /// bridge works back to the package's iOS 13 floor and forwards completion so a finite publisher
+    /// (e.g. the test lifecycle stream) terminates the `for await` loop.
+    func lifecycleEventStream() -> AsyncStream<LifeCycleEvents> {
+        AsyncStream(bufferingPolicy: .unbounded) { continuation in
+            let cancellable = sink(
+                receiveCompletion: { _ in continuation.finish() },
+                receiveValue: { continuation.yield($0) }
+            )
+            continuation.onTermination = { _ in cancellable.cancel() }
```
belleklaviyo, line 23: so this func replaces the transform and instead is just the direct stream that also works with tests

Sources/KlaviyoSwift/StateManagement/APIRequestErrorHandling.swift
```diff
@@ -1,94 +0,0 @@
-//
-//  APIRequestErrorHandling.swift
```
belleklaviyo, line 2: this was wholly replaced by the new [RequestQueueSupport.swift](https://github.com/klaviyo/klaviyo-swift-sdk/blob/feat/modularization-2/Sources/KlaviyoCore/RequestQueueSupport.swift) introduced in a prev PR

the only difference was `handleRequestError` (old) returned a KlaviyoAction (reducer-shaped, e.g. .deQueueCompletedResults/.requestFailed); `classifyFailure` (new) returns a plain FlushDecision enum (.dequeue / .retry / .retryWithBackoff / .clearInvalidFieldsAndDequeue) that the actor acts on directly — no TCA action wrapping.

Sources/KlaviyoSwift/ProfilePropertyBuffer.swift
```diff
@@ -0,0 +1,145 @@
+//
+//  ProfilePropertyBuffer.swift
```
belleklaviyo, line 2: some background on this -- setting individual (non identifier) profile props (setProfileProperty) used to stash them in an in-memory dict on KlaviyoState — pendingProfile. The reducer then folded those props into the next outbound request (at flush, identifier change, or token register) and cleared the dict

this is effectively that but relocated off the reducer

Sources/KlaviyoSwift/ProfilePropertyBuffer.swift
```diff
@@ -0,0 +1,145 @@
...
+    static let shared = ProfilePropertyBuffer()
+
+    private let lock = NSLock()
+    private var staged: [Profile.ProfileKey: AnyEncodable] = [:]
+    private var generation = 0
+
+    // MARK: - API
+
+    /// Stages a single property key/value. Safe to call from any thread.
+    func stage(_ profileKey: Profile.ProfileKey, _ value: AnyEncodable) {
```
belleklaviyo, line 30: this is what the setters call, just no longer on KlaviyoState

Sources/KlaviyoSwift/ProfilePropertyBuffer.swift
```diff
@@ -0,0 +1,145 @@
...
+        lock.withLock { staged[profileKey] = value }
+    }
+
+    /// Snapshot-and-clears the staged dict, folds the properties into the appropriate request, and
+    /// enqueues it. No-op when the buffer is empty or when `apiKey` is not yet configured.
+    func flushIntoQueue() async {
```
belleklaviyo, line 36: this is called by the actor's `willDrain`

Sources/KlaviyoSwift/ProfilePropertyBuffer.swift
```diff
@@ -0,0 +1,145 @@
...
+        let pushTokenData = IdentityStore.shared.pushToken
+        guard let anonymousId = identity.anonymousId else {
+            restore(snapshot, generation: capturedGeneration)
+            environment.emitDeveloperWarning(
+                "ProfilePropertyBuffer.flushIntoQueue: missing anonymousId; staged properties retained"
+            )
+            return
+        }
+
+        guard lock.withLock({ generation == capturedGeneration }) else { return }
```
belleklaviyo, line 65: this is for safety so a reset racing a flush won't ship stale props

Sources/KlaviyoSwift/StateManagement/KlaviyoState+RequestBuilding.swift
```diff
@@ -52,112 +44,8 @@ extension KlaviyoState {
             pushToken: pushToken,
             enablement: enablement,
             background: environment.getBackgroundSetting().rawValue,
-            profile: ProfilePayload(resolveAndConsumePendingProfile(), anonymousId: anonymousId)
-        )
-    }
-
-    mutating func enqueueProfileOrTokenRequest() {
```
belleklaviyo, line 59: these were called originally from the reducer, effect moved to ProfilePropertyBuffer

Sources/KlaviyoSwift/StateManagement/KlaviyoState+RequestBuilding.swift
```diff
@@ -52,112 +44,8 @@ extension KlaviyoState {
...
-        mergePendingAttributes(from: updatedProfile, into: &attributes)
-        attributes.location = mergedLocation(from: updatedProfile, into: attributes.location ?? .init())
-        self.pendingProfile = nil
-
-        return .init(data: .init(attributes: attributes))
-    }
-
-    /// Fills in profile attributes (name, title, organization, image, properties) from a pending
-    /// profile without overwriting values already present on the request.
-    private func mergePendingAttributes(
```
belleklaviyo, line 103: this effect is preserved and moved to PendingProfileFold for the new buffer to use

Sources/KlaviyoSwift/StateManagement/PendingProfileFold.swift
```diff
@@ -0,0 +1,77 @@
...
+//  Created by Isobelle Lim on 9/11/26.
+//
+
+import AnyCodable
+import KlaviyoCore
+
+/// State-free helpers for folding a staged profile (from `setProfileProperty`) into an existing
+/// `CreateProfilePayload`. Used by the `ProfilePropertyBuffer` drain path to merge staged properties
+/// into the outbound request with no `self` dependency.
+enum PendingProfileFold {
```
belleklaviyo, line 14: same code here just relocated as mentioned

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
```diff
@@ -63,12 +56,7 @@ struct KlaviyoState: Equatable {
         phoneNumber: String? = nil,
         externalId: String? = nil,
         pushTokenData: PushTokenData? = nil,
-        requestsInFlight: [KlaviyoRequest] = [],
-        initalizationState: InitializationState = .uninitialized,
-        flushing: Bool = false,
-        flushInterval: Double = StateManagementConstants.wifiFlushInterval,
-        retryState: RetryState = .retry(StateManagementConstants.initialAttempt),
-        pendingProfile: [Profile.ProfileKey: AnyEncodable]? = nil
```
belleklaviyo, line 71: all of this moved to the `RequestQueue` in KlaviyoCore

Tests/KlaviyoSwiftTests/GeofenceEventDispatchTests.swift
```diff
@@ -94,29 +93,48 @@ final class GeofenceEventDispatchTests: XCTestCase {
...
+        // When: dispatch a geofence event with matching API key. The prioritized event forces an
+        // immediate flush on the Core RequestQueue actor, which drains the durable QueueStore and
+        // sends through the test transport.
         GeofenceEventDispatch.dispatch(event: makeGeofenceEvent(), apiKey: apiKey)
-        await fulfillment(of: [flushed], timeout: 1.0)
 
-        // Then: the durable queue is drained (items leased into the in-flight set on flush).
+        // Then: the durable queue drains (the actor flush is async/off-store, so poll it).
+        try await waitUntilQueueEmpty(readQueue)
+        assertGeofenceEventSent(sentRequests.value)
```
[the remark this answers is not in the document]
belleklaviyo: can refactor later, low priority

Sources/KlaviyoCore/RequestQueue.swift
glenn-klaviyo: will this be cleaned up some more after we are dev complete? Do we need to be explicit about why this divereges or is that just a comment for later ai agents?

belleklaviyo: yeah I can reword it so it's less focused on what looks like divergence against "the parity" -- functionally this is not actually different, it's just that the engine driving the attempt number is different now and stores and uses the first attempt number directly as opposed to the reducer started at 0 and incremented before using it

Sources/KlaviyoCore/RequestQueue.swift
glenn-klaviyo: makes sense

Sources/KlaviyoCore/RequestQueueProtocol.swift
glenn-klaviyo: smart!

Sources/KlaviyoSwift/StateManagement/KlaviyoState.swift
glenn-klaviyo: love the simplification here

Sources/KlaviyoSwift/StateManagement/PendingProfileFold.swift
glenn-klaviyo: wondering if a better name here would be `ProfileAttributeMerge` or something 🤔 also okay with keeping this if it makes sense to you :)

Sources/KlaviyoCore/RequestQueue.swift
```diff
@@ -110,35 +108,26 @@ public actor RequestQueue {
...
-    /// restore (retries next tick); rate-limit/server → record a durable countdown backoff + restore
-    /// the request, which the countdown gate below waits out over ticks before resending.
+    /// Runs `willDrain`, leases the whole `QueueStore` batch, and sends head-first (FIFO). See the
+    /// type doc for the success/failure handling.
     private func flush() async {
-        // 1. Gate: pre-init or offline. Do not flush.
+        // Gate: pre-init or offline.
         guard SDKConfigStore.shared.current.apiKey != nil, flushInterval.isFinite else { return }
+        // Reentrancy guard: a `flushNow()` must not interleave with an in-progress flush.
         guard !isFlushing else { return }
```
amber-klaviyo, line 116: Nothing exercises this reentrancy guard directly — a test that fires two near-simultaneous flush triggers (e.g. two rapid `flushNow()` calls, or a company switch racing a high-priority event) and asserts the transport only sees one drain cycle would give real confidence this holds, rather than resting on the comment alone. This is one of the actor's core concurrency guarantees, so it seems worth a dedicated regression test alongside the other `RequestQueueLifecycleTests` wiring coverage.

Tests/KlaviyoSwiftTests/GeofenceEventDispatchTests.swift
```diff
@@ -94,29 +93,48 @@ final class GeofenceEventDispatchTests: XCTestCase {
...
+    }
+
+    /// Polls the durable QueueStore until it drains, since the Core `RequestQueue` actor's flush runs
+    /// asynchronously off the reducer store (no reducer state change to observe).
+    private func waitUntilQueueEmpty(
```
amber-klaviyo, line 122: This polls `readQueue().isEmpty` every 20ms against a 2s timeout instead of a completion signal. Under a loaded CI runner this either flakes on a real timeout or just burns up to 2s of wasted wall time per run. Both call sites are a few lines away from a transport stub (`environment.klaviyoAPI.send`) — an `XCTestExpectation` fulfilled from that stub (same pattern already used for `sentRequests` right above) would be deterministic instead of wall-clock-based.

Sources/KlaviyoSwift/ProfilePropertyBuffer.swift
```diff
@@ -0,0 +1,145 @@
...
+        let (snapshot, capturedGeneration) = lock.withLock {
+            defer { staged = [:] }
+            return (staged, generation)
+        }
+
+        guard !snapshot.isEmpty else { return }
+
+        guard let apiKey = SDKConfigStore.shared.current.apiKey else {
+            restore(snapshot, generation: capturedGeneration)
+            environment.emitDeveloperWarning(
```
amber-klaviyo, line 47: Minor: both `emitDeveloperWarning` calls here (missing apiKey, missing anonymousId) fire on every periodic drain while the condition persists — roughly every `flushInterval` tick (~30s on wifi) for as long as an app is stuck pre-init with staged properties. Worth a "warn once until resolved" guard so a real integration issue doesn't get buried under repeated identical log spam. Not blocking, just a quick win.

amber-klaviyo (approved):
I left a couple comments with suggestions for small improvements (mainly better tests), but it's up to you if you think they're worth implementing.

glenn-klaviyo (approved):
Approving — nice cutover. Doing the `IdentityStore` locking work first as its own prereq was the right call, and the new `mutate` genuinely earns its place: the old read-then-write meant two writers could stomp each other (read the profile, someone else writes, you save your stale copy over theirs), and `mutate` closes that. Same for the `RequestQueueProtocol` seam — it's there because the spy tests actually need it, not just in case.

Three things looked like regressions at first glance, so I dug in, and all three are fine:
- Staged profile props moving to an in-memory buffer looks like it'd lose them on a cold start, but `pendingProfile` was never in `KlaviyoState.CodingKeys` on master — it was already memory-only. Nothing lost.
- `classifyFailure` really is just `handleRequestError` with the TCA wrapping taken off. Same classifications.
- Deleting `AttemptNumberTests` doesn't lose coverage — `RequestQueueTests` asserts `sentAttempts == [1, 2]`, which pins the first attempt at 1 *and* the advance, so it covers strictly more.

I also poked at the coexistence limitation you documented. Every identity-mutating reducer path I traced reseeds `state.identity` from `IdentityStore` before mutating, and the `defer` only writes when the projection actually changed — so the window where a cleared email could come back looks narrower than the comment implies. Didn't check exhaustively, and it's MAGE-904's problem regardless, but it's less scary than it reads.

One thing worth a look whenever you're next in here: `.terminated` now goes through the same `handleBackground()` as `.backgrounded`, which means `await requestQueue.stop()` is the only thing standing between app teardown and the queue hitting disk. The `.synchronous` persist keeps the lease safe *if* that await actually lands, but on `willTerminate` the process might not stick around long enough to hop into the actor. Pretty sure this predates you — the old `.stop` arm awaited the same way — but it used to be one of a few things protecting the queue and now it's the only one, so worth a deliberate "yeah, still fine" instead of just inheriting it.

Couple of nits, take or leave: `StateManagementConstants` is fully dead now (no call sites anywhere in Sources/Tests/Examples, even though its comment says it's kept for internal callers). And `PendingProfileFold` keeps the "pending" name this PR is deleting — something like `extension ProfilePayload.Attributes { mutating func backfill(from:) }` would also say *which side wins* in the name, which is the part you currently have to read the bodies to figure out.

-- via Claude (review with Glenn)

belleklaviyo:
want to get this in a dev-complete state and am planning on doing a clean up PR to address any testing gaps, old tests, outdated comments, etc. but noted all of these nits for later! thanks both for the reviews!!

PR #726, opened 2026-09-15
refactor: delete vendored TCA; route KlaviyoSwift entry points at Core (MAGE-904)

belleklaviyo:
# Description
Terminal chip of the iOS modularization: delete the vendored The-Composable-Architecture and dissolve the `KlaviyoReducer`/`KlaviyoAction`/`KlaviyoState` projection into direct `KlaviyoOrchestration` calls against the canonical KlaviyoCore stores. Behavior-preserving.

> **Stacked PR.** Base is `isobellelim/mage-1197-flush-cutover` (#725). To be rebased onto `feat/modularization-2` once #725 merges. Review the [#725..904 diff](https://github.com/klaviyo/klaviyo-swift-sdk/compare/isobellelim/mage-1197-flush-cutover...isobellelim/mage-904-replace-the-queue-engine-with-an-actor-requestqueue-delete) to see only this PR's changes.

## Due Diligence

Internal refactor only — the public `Klaviyo` facade and the `@_spi(KlaviyoPrivate)` `StateChangePublisher` contract are unchanged. Also drops the now-unused `swift-case-paths` and `combine-schedulers` SPM dependencies (backwards-compatible dependency-graph change for SPM/CocoaPods consumers).

## Changelog / Code Overview
By the time this runs, the flush/retry engine is already an `actor RequestQueue` in KlaviyoCore (MAGE-1197). This PR removes the last TCA usage:

- **Deleted:** `Sources/KlaviyoSwift/Vendor/ComposableArchitecture/` (8 files, ~2844 LOC), `KlaviyoReducer` + `KlaviyoAction` + `Store.production` (`StateManagement.swift`), and the `KlaviyoState` projection + its write-through `defer`.
- **Added:** `LifecycleState` — an un-persisted, session-scoped init-state holder (the only surviving mutable SDK state); a `KlaviyoOrchestration` layer reproducing each former reducer case as a direct call into Core (`IdentityStore`/`SDKConfigStore`/`QueueStore`/`RequestEnqueuer`/`RequestQueue`), using atomic `IdentityStore.mutate` for identity read-modify-writes; pure request-builder free functions; and `IdentityStore.tokenPublisher` in Core.
- **Rewired:** `Klaviyo.swift`, `GeofenceEventDispatch`, `KlaviyoEventDispatcher`, and `KlaviyoSwiftEnvironment` (drops `send`/`state`/`statePublisher`); `StateChangePublisher` reassembled from `CombineLatest3(IdentityStore.publisher, IdentityStore.tokenPublisher, LifecycleState.publisher)` so push-token changes still reach the `@_spi` observer.
- **Deps:** removed `swift-case-paths` + `combine-schedulers` from `Package.swift`.

Serialization is unchanged (the existing `dispatchOnMainThread` main-queue funnel). This also resolves the "known coexistence limitation" #725 documented — a 4xx field-clear can no longer be resurrected by a stale projection, since `IdentityStore` is now the sole writer (regression test included).

Commit history is ordered so the Core prep lands first; recommend **squash merge**.

## Test Plan
- Full package test suite green on iOS Simulator (iPhone 16 Pro): KlaviyoCore 333, KlaviyoSwift 309, KlaviyoForms 65, KlaviyoLocation 37, KlaviyoSwiftExtension 19.
- Behavior parity is enforced by the orchestration test suites (`OrchestrationIdentityTests`, `OrchestrationProfileTokenTests`, `OrchestrationEventTrackingTests`, `OrchestrationInitializeTests`), which reproduce the deleted reducer tests' coverage, plus a regression test for the 4xx-clear-then-set case.
- Repro: `xcodebuild test -scheme klaviyo-swift-sdk-Package -destination "platform=iOS Simulator,name=iPhone 16 Pro"`.

## Related Issues/Tickets
Related to MAGE-904
Stacked on #725 (MAGE-1197 PR2)

## Summary by CodeRabbit

* **New Features**
  * Added lifecycle-aware SDK initialization and state handling.
  * Improved support for API key changes while preserving identity and push-token data.
  * Added reliable buffering and ordered processing for events, profiles, subscriptions, and tracking activity.
  * Added lifecycle-aware request queueing, flushing, reachability, and foreground/background handling.

* **Bug Fixes**
  * Prevented stale automatic push-token updates from overwriting newer values.
  * Improved geofence event handling during initialization and API key changes.


belleklaviyo:
bugbot run

Sources/KlaviyoSwift/StateManagement/KlaviyoOrchestration.swift
```diff
@@ -0,0 +1,403 @@
...
+//  - The post-init gate checks `LifecycleState.shared.current != .uninitialized` (session-fresh,
+//    equivalent to the old `state.apiKey != nil`); the apiKey VALUE is read from `SDKConfigStore`.
+
+import AnyCodable
+import Foundation
+import KlaviyoCore
+import OSLog
+
+/// Namespace for direct orchestration functions that mirror the identity-setter reducer cases.
+enum KlaviyoOrchestration {
```
belleklaviyo, line 23: as discussed, since this is now the top level surface layer, we probably will want to promote this outside StateManagement folder and maybe rename it something else. Any rename/organizational moves will be in a separate PR

Sources/KlaviyoCore/IdentityStore.swift
```diff
@@ -13,9 +13,14 @@ public protocol IdentityReading {
     var current: ProfileData { get }
     var pushToken: PushTokenData? { get }
     var publisher: AnyPublisher<ProfileData, Never> { get }
+    var tokenPublisher: AnyPublisher<PushTokenData?, Never> { get }
```
belleklaviyo, line 16: this is to support the `internalStatePublisher` that includes `pushToken` in its private state, since push token data is not in ProfileData, it needs its own publisher

Sources/KlaviyoSwift/StateManagement/LifecycleState.swift
```diff
@@ -0,0 +1,97 @@
...
+
+/// The SDK's session-scoped initialization state machine.
+///
+/// `uninitialized → initializing → initialized` is the only legal forward path.
+/// `reset()` returns the holder to `.uninitialized` for test isolation.
+///
+/// - Important: This state is **never persisted**. Each cold-start begins `.uninitialized`
+///   so that the migration pass, UnattributedBuffer drain, and lifecycle `start()` always
+///   run exactly once per process lifetime.
+enum InitializationState: Equatable {
```
belleklaviyo, line 19: organization-wise, I think this makes sense to keep in KlaviyoSwift as its more per-session based rather than persisted info

Sources/KlaviyoSwift/StateManagement/StateChangePublisher.swift
```diff
@@ -7,19 +7,36 @@
...
 public enum StateChangePublisher {
-    private static func createStatePublisher() -> AnyPublisher<KlaviyoState, Never> {
-        klaviyoSwiftEnvironment.statePublisher()
-            .filter { state in state.initalizationState == .initialized }
-            .removeDuplicates()
-            .eraseToAnyPublisher()
+    /// Assembles the private state feed from the canonical Core publishers (identity + token) gated
+    /// on the SDK having reached `.initialized`. Replaces the old TCA `statePublisher()` seam.
+    private static func createStatePublisher() -> AnyPublisher<PrivateState, Never> {
+        Publishers.CombineLatest3(
```
belleklaviyo, line 18: neat https://developer.apple.com/documentation/combine/publishers/combinelatest3

Sources/KlaviyoSwift/Klaviyo.swift
```diff
@@ -14,9 +14,11 @@ import KlaviyoCore
 import OSLog
 import UIKit
 
-func dispatchOnMainThread(action: KlaviyoAction) {
+/// Funnels SDK entry-point work onto the main queue. Matches the prior main-async semantics of the
+/// TCA `send` dispatch (always hops to main; no inline-on-main fast path) so ordering is unchanged.
+func dispatchOnMainThread(_ work: @escaping () -> Void) {
```
belleklaviyo, line 19: feels like a flimsier wrapper now, will look into if this can be reduced and/or if we could just utilize the `@MainActor` tags in the functions that actually need it

Sources/KlaviyoSwift/KlaviyoSwiftEnvironment.swift
```diff
@@ -14,9 +14,6 @@ import UserNotifications
 var klaviyoSwiftEnvironment = KlaviyoSwiftEnvironment.production
 
 struct KlaviyoSwiftEnvironment {
```
belleklaviyo, line 16: I originally thought this was also a relic of TCA because of the name but it is still being used functionally. as dependency injection site for global side effects, so we still need to maintain it, but at least it is slimmed down more without state or the publisher

Sources/KlaviyoSwift/StateManagement/KlaviyoOrchestration+Initialize.swift
```diff
@@ -0,0 +1,204 @@
...
+                // Appended so it sends after any queued old-company requests; persisted
+                // synchronously so it survives a crash before the first flush.
+                QueueStore.shared.enqueue(request, persist: .synchronous)
+            }
+            // NOTE: do NOT clear the push token here — the switch must preserve it so the
+            // token can be re-registered under the new company immediately below.
+            // Give the new company a clean identity: mint a fresh anon and drop any PII so
+            // `completeInitialization` hydrates it. Unconditional (matches the runtime switch
+            // path's reset()) so an anonymous-only switch does not carry the old company's anon.
+            IdentityStore.shared.update(ProfileData(anonymousId: IdentityStore.shared.mintNewAnonymousId()))
```
glenn-klaviyo, line 123: Branch 2 (cold-start company switch, just above) clears identity PII the same way Branch 1 (runtime company switch) does, but doesn't call `ProfilePropertyBuffer.shared.reset()` afterward the way Branch 1 does (see line 670 above). Is that intentional — e.g. the buffer is guaranteed empty pre-init so there's nothing to clear — or a parity gap versus Branch 1?

belleklaviyo: The buffer only holds properties staged in the current app launch, and on a cold start those belong to the new company you're about to init — init flushes them a few lines later. Calling reset() here would throw away the new company's own properties. Branch 1 resets because there the SDK was already running, so the buffer could be holding old-company data and needs that explicit reset()

Sources/KlaviyoSwift/StateManagement/KlaviyoOrchestration+Initialize.swift
```diff
@@ -0,0 +1,204 @@
...
+            // Reset identity with preserveTokenData:true (port of KlaviyoState.reset(true)):
+            //  - If identified → mint a fresh anonymousId.
+            //  - Clear PII. Token untouched inside mutate (lives in IdentityStore separately).
+            //  - Clear staged profile properties.
+            //  - Re-enqueue the token under the NEW apiKey.
+            let previousPushTokenData = IdentityStore.shared.pushToken
+            IdentityStore.shared.mutate { profile in
+                if profile.email != nil || profile.phoneNumber != nil || profile.externalId != nil {
+                    profile.anonymousId = IdentityStore.shared.mintNewAnonymousId()
+                }
```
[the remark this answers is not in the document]
belleklaviyo: this was discussed as a known parity difference with Android, addressed in https://github.com/klaviyo/klaviyo-swift-sdk/pull/729

ajaysubra (approved):
Belle's run though of the changes made sense at a high level and approving based on that.

Sources/KlaviyoSwift/StateManagement/KlaviyoOrchestration+Initialize.swift
```diff
@@ -0,0 +1,204 @@
...
+    /// Ports `KlaviyoReducer.reduce(.completeInitialization)` effect (StateManagement.swift:237-267)
+    /// verbatim. Replaces `send(.setPushEnablement(settings))` with a direct `setPushEnablement`
+    /// call (no TCA dispatch).
+    private static func runLifecycle() async {
+        @Sendable
+        func handleForeground() async {
+            await klaviyoSwiftEnvironment.requestQueue.start()
+            let settings = await environment.getNotificationSettings()
+            // Direct call instead of `send(.setPushEnablement(settings))`.
+            setPushEnablement(settings)
```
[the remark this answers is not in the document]
belleklaviyo: Fixed: `runLifecycle()` is now `@MainActor`, so `setPushEnablement` (and any identity/token writes it triggers) rejoin the main funnel and can no longer interleave with a host `setPushToken`, `resetProfile`, or company switch.

PR #728, opened 2026-09-17
chore: bump Klaviyo API revision to 2026-07-15 (MAGE-1239)

Atharva8168:
# Description
Bumps the Klaviyo API revision from `2026-01-15` to `2026-07-15`, and deletes the geofence-specific `.pre` override that is no longer needed.

## Due Diligence
- [x] I have adjusted or added new test cases to team test docs, if applicable.

Planned version: **5.5.0**. Minor rather than patch because this changes the wire behavior of every request an integrator sends. Every previous revision bump in this repo shipped in a minor (5.1.0, 5.2.0).

## Changelog / Code Overview
- `KlaviyoEndpoint.swift` — `revision` collapses from a two-branch computed property to `"2026-07-15"`. Kept deliberately plain per review; the reason the `.pre` pin existed is recorded below rather than in a code comment.
- `KlaviyoEndpointTests.swift` — the 5 revision assertions now read a single `expectedRevision` constant, so the next bump touches two places instead of six; `testRevisionHeaderForGeofenceEndpoint` renamed to `testRevisionHeaderForFetchGeofencesEndpoint`, and the "standard revision" comments dropped since there is now only one revision.

**Why delete rather than re-date:** the override existed only because `GET /client/geofences` was beta-only at `2026-01-15`. The endpoint went GA at `2026-04-15`. Re-dating to `2026-07-15.pre` would pin production to a pre-release contract, and a second literal would go stale independently of the default.

The renamed geofence test is kept rather than folded into its sibling: it is the only test that drives the real `fetchGeofences` endpoint through `urlRequest`, so it still guards against a per-endpoint override returning.

iOS has no integrator-facing revision override, so there is no configuration tradeoff on this side.

## Test Plan
Automated (run, green):
- `make IOS_RUNTIME="iOS 26" test-library` → `** TEST SUCCEEDED **` (debug; CI also covers release).
- Note: plain `swift test` cannot build this package — it fails on `UIKit/UIKit.h`. Use the Makefile/xcodebuild path.
- Pre-commit SwiftLint and SwiftFormat both passed on the changed files.

Live API evidence for the revision itself (`GET /client/geofences`, bogus `company_id`, no account data read):

| revision sent | result |
|---|---|
| `2026-01-15` | 404 |
| `2026-01-15.pre` | 200 |
| `2026-04-15` | 200 |
| `2026-07-15` | 200 |

Manual, still to do before merge:
1. Fetch geofences from the test app against an account **that has geofences**, and assert the parsed count and ids — not HTTP 200.
2. Send one event and one push-token registration; expect 202 each.
3. Identify a profile (`POST /client/profiles`); expect 202.

Note for whoever tests manually: `klaviyo-ios-test-app` has its own endpoint package pinned to `2025-10-15`, so its traffic does not exercise this change.

## Related Issues/Tickets
Part of MAGE-1130

- MAGE-1239 — https://linear.app/klaviyo/issue/MAGE-1239/ios-bump-api-revision-to-2026-07-15
- Parent MAGE-1130 — https://linear.app/klaviyo/issue/MAGE-1130/bump-sdk-api-revision-from-2026-01-15-to-latest

## Summary by CodeRabbit

* **Updates**
  * Standardized the API revision used by all endpoints to `2026-07-15`.
  * Updated revision-header coverage across geofence, profile, event, geofence-event, and subscription requests.


Sources/KlaviyoCore/Networking/KlaviyoEndpoint.swift
```diff
@@ -152,12 +152,10 @@ public enum KlaviyoEndpoint: Equatable, Codable {
         }
     }
 
+    // One revision for every endpoint. Geofences pinned a `.pre` revision until
+    // GET /client/geofences went GA on 2026-04-15 (MAGE-1130).
```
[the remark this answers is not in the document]
Atharva8168: Ahh!! Good find.

Sources/KlaviyoCore/Networking/KlaviyoEndpoint.swift
```diff
@@ -152,12 +152,10 @@ public enum KlaviyoEndpoint: Equatable, Codable {
         }
     }
 
+    // One revision for every endpoint. Geofences pinned a `.pre` revision until
```
belleklaviyo: not sure we really need this comment just to keep it slim, this comment can maybe just be noted in this PR if we ever need to trace it up but I'd be fine just keeping it plain

belleklaviyo (approved):

Sources/KlaviyoCore/Networking/KlaviyoEndpoint.swift
Atharva8168: Sounds good to me. Moving it to PR from the comments.

